Issue #6, opened 2021-04-27
[common] Improve query deserialization


irakliyk:
`into_batch()` method of [Queries](https://github.com/novifinancial/winterfell/blob/c3e201228b321f48dcb3cd27cf09b7677b2d0802/common/src/proof.rs#L130) struct should return a vector of field elements as a second tuple value.

PR #15, opened 2021-04-30
Queries serialization improvements

irakliyk:
This PR addresses #6 and also improves queries serialization/deserialization overall. Specifically:

- [x] Implement deserialization of query bytes directly into field elements
- [x] Implement flat serialization of query values and Merkle paths (for trace and constraint queries only)

The overall impact of this PR is ~1% reduction in proof sizes, but it also lays groundwork for switching over to new Merkle tree implementation.

### DIFF
--- a/common/src/proof.rs
+++ b/common/src/proof.rs
@@ -51,8 +51,8 @@ pub struct Commitments {
 
 #[derive(Clone, Serialize, Deserialize)]
 pub struct Queries {
-    pub paths: Vec<Vec<[u8; 32]>>,
-    pub values: Vec<Vec<u8>>,
+    pub paths: Vec<u8>,
+    pub values: Vec<u8>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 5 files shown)


common/src/proof.rs
```diff
@@ -113,34 +117,77 @@ impl StarkProof {
...
 
 impl Queries {
-    /// Returns a set of queries constructed from a batch Merkle proof and corresponding values.
-    pub fn new<E: FieldElement>(merkle_proof: BatchMerkleProof, values: Vec<Vec<E>>) -> Self {
+    /// Returns a set of queries constructed from a batch Merkle proof and corresponding elements.
+    pub fn new<E: FieldElement>(merkle_proof: BatchMerkleProof, query_values: Vec<Vec<E>>) -> Self {
+        assert!(!query_values.is_empty(), "query values cannot be empty");
+        let elements_per_query = query_values[0].len();
+        assert!(
+            elements_per_query != 0,
```
kchalkias, line 125: prefer `is_empty`

common/src/proof.rs
```diff
@@ -113,34 +117,77 @@ impl StarkProof {
...
+        let paths = merkle_proof.serialize_nodes();
+
+        Queries { paths, values }
     }
 
-    /// Convert a set of queries into a batch Merkle proof and corresponding values.
-    /// TODO: return values as a vector of field elements
-    pub fn into_batch<H: Hasher>(self, num_leaves: usize) -> (BatchMerkleProof, Vec<Vec<u8>>) {
+    /// Convert a set of queries into a batch Merkle proof and corresponding query values.
+    pub fn deserialize<H: Hasher, E: FieldElement>(
```
kchalkias, line 151: Can any zero value or length cause issues or panic (ie division by zero?)

kchalkias (approved):
minor comments

common/src/proof.rs
irakliyk: Good point! I've added some assertions here as this code gets called internally from within Winterfell. So, if it is called with wrong parameters, it's ok to panic.

common/src/proof.rs
irakliyk: I usually do too - but in this case we need to cache the `elements_per_query` value - so, I think it's OK to have it like this.

Issue #5, opened 2021-04-27
[FRI] Use Queries struct in FriProof


irakliyk:
We should refactor [FriProof](https://github.com/novifinancial/winterfell/blob/c3e201228b321f48dcb3cd27cf09b7677b2d0802/fri/src/proof.rs#L16) struct to rely on [Queries](https://github.com/novifinancial/winterfell/blob/c3e201228b321f48dcb3cd27cf09b7677b2d0802/common/src/proof.rs#L49) struct for internal representation of queries. This will unify serialization/deserialization approaches. But, it will also require moving `Queries` struct into `utils` crate, since `fri` crate does not depend on `common` crate.

PR #16, opened 2021-05-04
Fri proof serialization improvements

irakliyk:
This PR refactors FRI proof serialization/deserialization to be more in-line with main proof serialization. This is similar to #5, however, we cannot use `Queries` struct directly in FRI because the structures are slightly different.

The impact of this PR is further reduction of proof size by ~3% (in addition to the reduction from #15). It also lays groundwork in the FRI module for migration to the new Merkle tree implementation.

### DIFF
--- a/fri/src/proof.rs
+++ b/fri/src/proof.rs
@@ -3,18 +3,180 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use crate::ProofSerializationError;
+use crypto::{BatchMerkleProof, Hasher};
+use math::{
+    field::FieldElement,
+    utils::{log2, read_elements_into_vec},
+};
+
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct FriProofLayer {
-    pub values: Vec<Vec<u8>>,
-    pub paths: Vec<Vec<[u8; 32]>>,
-    pub depth: u8,
-}
+// FRI PROOF
+// ================================================================================================
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct FriProof {
-    pub layers: Vec<FriProofLayer>,
-    pub rem_values: Vec<u8>,
-    pub partitioned: bool,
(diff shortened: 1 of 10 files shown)


fri/src/proof.rs
```diff
@@ -3,18 +3,170 @@
...
+        domain_size: usize,
+        folding_factor: usize,
+    ) -> Result<(Vec<E>, BatchMerkleProof), SerializationError> {
+        let hash_fn = H::hash_fn();
+
+        // make sure we have enough bytes to read the expected number of queries
+        let num_query_bytes = E::ELEMENT_BYTES * folding_factor;
+        let num_queries = self.values.len() / num_query_bytes;
+        if self.values.len() % num_query_bytes != 0 {
+            let expected_bytes = (num_queries + 1) * num_query_bytes;
```
kchalkias, line 145: is this the same with `self.values.len() + num_query_bytes`? Can you crosscheck whether this `if` statement is correct?

fri/src/proof.rs
```diff
@@ -3,18 +3,170 @@
...
+    pub fn new<E: FieldElement, const N: usize>(
+        query_values: Vec<[E; N]>,
+        merkle_proof: BatchMerkleProof,
+    ) -> Self {
+        assert!(!query_values.is_empty(), "query values cannot be empty");
+
+        // TODO: add debug check that values actually hash into the leaf nodes of the batch proof
+
+        // concatenate all query values together into a single vector of bytes
+        let num_queries = query_values.len();
```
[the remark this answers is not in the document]
irakliyk: fixed!

fri/src/proof.rs
irakliyk: Not quite: the division on line 143 is a truncating division - so, if the bytes in `self.values` can't be evenly divided into query elements it will be one less than whole number of query values. For example:

* Let's say `num_query_bytes` = 64 and `self.values.len()` = 129
* Then, `num_queries` = 2, and `self.values.len() % num_query_bytes` = 1
* This means there is one extra byte, and thus `expected_bytes` will be computed as 3 * 64 = 192

kchalkias (approved):

Issue #8, opened 2021-04-28
Migrate to new MerkleTree implementation


irakliyk:
A new [MerkleTree](https://github.com/novifinancial/winterfell/blob/c3e201228b321f48dcb3cd27cf09b7677b2d0802/crypto/src/merkle2/mod.rs#L20) has already been implemented but it is not yet integrated into the rest of the library. Migrating to this new implementation will allow us to relax the condition that hash function output must be 32 bytes.

PR #17, opened 2021-05-06
New Merkle tree implementation

irakliyk:
This PR migrates the codebase to the new `MerkleTree` implementation as described in #8.

- [x] Implement batch Merkle proof serialization for new Merkle tree
- [x] Update random element generator to use new `Hasher` trait
- [x] Migrate the codebase to the new Merkle tree implementation
- [x] Update Merkle proof verification example
- [x] Update comments and docs


crypto/src/hash/mod.rs
```diff
@@ -35,13 +52,31 @@ impl Hasher for Blake3_256 {
...
         let bytes = E::elements_as_bytes(elements);
         blake3::hash(&bytes).into()
     }
 
-    fn hash_fn() -> HashFunction {
-        blake3
+    fn merge_with_int(seed: Self::Digest, value: u64) -> Self::Digest {
+        let mut data = [0; 64];
+        data[..32].copy_from_slice(&seed);
+        data[56..].copy_from_slice(&value.to_le_bytes());
```
kchalkias, line 71: Ooc, why did we choose to append the u64 value at the end and not use data[40]? Same for sha3.

irakliyk: no good reason. fixed!

common/src/proof/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        // conjectured security requires half the queries as compared to proven security
+        let num_queries = if optimistic {
+            options.num_queries()
+        } else {
+            options.num_queries() / 2
+        };
+
+        let one_over_rho =
+            (options.blowup_factor() / self.context.ce_blowup_factor as usize) as u32;
+        let security_per_query = 31 - one_over_rho.leading_zeros(); // same as log2(one_over_rho)
```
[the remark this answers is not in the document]
irakliyk: The extra bit was accounted for in the number of queries computation above. But this was super confusing, and proven security estimate wasn't accurate (the conjectured security was fine). So, I refactored proof security computations.

crypto/src/hash/mod.rs
```diff
@@ -68,7 +68,7 @@ impl Hasher for Blake3_256 {
     fn merge_with_int(seed: Self::Digest, value: u64) -> Self::Digest {
         let mut data = [0; 64];
```
kchalkias, line 69: Prefer `let mut data = [0; 40]`, then `data[32..].copy_from_slice` for both hash schemes.

kchalkias (approved):
LGTM, just a micro-optimization comment

crypto/src/hash/mod.rs
irakliyk: fixed!

PR #18, opened 2021-05-10
Lamport+ signature aggregation

irakliyk:
This PR add an example of STARK proofs for Lamport+ signature aggregation. This is also the first example to illustrate how execution trace table can be constructed concurrently (in multiple threads).


kchalkias:
Please fix the benchmarks deployment issue.

PR #19, opened 2021-05-12
[prover] Point-wise division

irakliyk:
This PR implements point-wise application of constraint divisors during constraint polynomial composition. The advantages of point-wise division are:

- It is fully parallelizable (and trivially distributable).
- It is actually 2x - 3x faster that previous implementation which used synthetic division. Though, the impact on the overall proof generation is negligible since constraint composition was very fast to begin with.
- It lays groundwork for improved constraint evaluation methodology which should result in significant speedup, once implemented.

This PR affects only prover code as verifier code use a different methodology for applying constraint divisors.

### DIFF
--- a/prover/src/monolith/constraints/evaluation_table.rs
+++ b/prover/src/monolith/constraints/evaluation_table.rs
@@ -8,7 +8,7 @@ use common::{errors::ProverError, ConstraintDivisor};
 use math::{
     fft,
     field::{FieldElement, StarkField},
-    utils::{batch_inversion, get_power_series_with_offset},
+    utils::batch_inversion,
 };
 use utils::uninit_vector;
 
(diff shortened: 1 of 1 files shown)


prover/src/monolith/constraints/evaluation_table.rs
```diff
@@ -288,38 +275,162 @@ fn apply_divisor<B: StarkField, E: FieldElement + From<B>>(
...
-        polynom::syn_div_in_place(&mut poly, degree, a);
+        // the column represents merged evaluations of boundary constraints, and divisor has the
+        // form of (x^a - b); thus to divide the column by the divisor, we compute: value * z,
+        // where z = 1 / (x^a - 1) and has already been computed above.
+
+        if cfg!(feature = "concurrent") && result.len() >= MIN_CONCURRENT_SIZE {
+            #[cfg(feature = "concurrent")]
+            {
+                #[rustfmt::skip]
+                result.par_iter_mut().zip(column).enumerate().for_each(|(i, (result, value))| {
```
kchalkias, line 297: Reusing `result` for `result_i` was a bit confusing for the reader, maybe rename to `result_i` or `col_i` etc.

prover/src/monolith/constraints/evaluation_table.rs
```diff
@@ -288,38 +275,162 @@ fn apply_divisor<B: StarkField, E: FieldElement + From<B>>(
...
+                    // determine which value of z corresponds to the current domain point
+                    let z = E::from(z[i % z.len()]);
+                    // compute value * z and add it to the result
+                    *result += value * z;
+                });
+            }
+        } else {
+            for (i, (result, value)) in result.iter_mut().zip(column).enumerate() {
```
kchalkias, line 305: ditto

prover/src/monolith/constraints/evaluation_table.rs
```diff
@@ -288,38 +275,162 @@ fn apply_divisor<B: StarkField, E: FieldElement + From<B>>(
...
+
+        if cfg!(feature = "concurrent") && result.len() >= MIN_CONCURRENT_SIZE {
+            #[cfg(feature = "concurrent")]
+            {
+                let batch_size = result.len() / rayon::current_num_threads().next_power_of_two();
+                #[rustfmt::skip]
+                result.par_chunks_mut(batch_size).enumerate().for_each(|(i, batch)| {
+                    let batch_offset = i * batch_size;
+                    let mut x = domain_offset * g.exp((batch_offset as u64).into());
+                    for (i, result) in batch.iter_mut().enumerate() {
```
kchalkias, line 329: ditto

kchalkias (approved):
Ooc, how do we ensure that the new implementation is consistent with the previous approach?

prover/src/monolith/constraints/evaluation_table.rs
irakliyk: fixed!

prover/src/monolith/constraints/evaluation_table.rs
irakliyk: fixed!

prover/src/monolith/constraints/evaluation_table.rs
irakliyk: fixed!

irakliyk:
> Ooc, how do we ensure that the new implementation is consistent with the previous approach?

In this specific case, I've manually verified that new results are identical to the old results before removing the old approach from the codebase.

PR #21, opened 2021-05-15
New constraint evaluation methodology

irakliyk:
This PR will implement new constraint evaluation methodology where the composition polynomial is split into multiple columns each of length equal to the trace length.

- [x] Implement the new evaluation methodology
- [x] Cleanup, refactoring, and commenting
- [x] Update examples and documentation 

The impact of this PR should be ~3x speed up in prover time and ~20% decrees in proof size.


kevinlewi (approved):
Excellent!

PR #22, opened 2021-05-18
Threshold lamport example

irakliyk:
This PR will implement threshold version of LamportPlus signature aggregation.


kchalkias (approved):
LGTM, obviously in the future we have to provide a description on how LamportPlus works.
There exists a few duplicated code blocks i.e., `build_key_schedule` and `init_hash_state` are the same as in aggregated Lamport. Maybe we should reuse some of them in the future, but it's fine for this PR.

PR #23, opened 2021-05-23
Public coin improvements

irakliyk:
This PR improves public coin handling in the prover and verifier so that it follows more closely the interactive version of the protocol.

- [x] Refactor generation of coefficients for random linear combinations
- [x] Replace public coin in common crate with a more generic version in crypto crate
- [x] Integrate new public coin into prover and verifier channels (including FRI verifier channel)
- [x] Incorporate public inputs into public coin seed initialization
- [x] Updated documentation

---

The new approach relies on [PublicCoin](https://github.com/novifinancial/winterfell/blob/public-coin-improvements/crypto/src/random/mod.rs#L16) struct which can be reseeded as the protocol progresses. The reseeding is done roughly as `seed = hash(seed || new_seed)`. At the high-level, the public coin simulates prover-verifier interaction as follows:

1. The public coin is initialized with the hash of `public inputs` || `proof context`, where proof context contains basic proof parameters such as number of queries, blowup factor etc.
2. After the prover builds the execution trace Merkle tree, the coin is reseeded with the root of this tree. This commits the prover to the extended execution trace.
   a. After this step, the prover draws a set of random coefficients from the coin. These coefficients are used in building constraint composition polynomial.
3. After the prover builds the constraint evaluation Merkle tree, the coin is reseeded with the root of this tree. This commits the prover to the evaluations of composition polynomial.
  a. After this step, the prover draws a random out-of-domain point from the public coin. This point is used to build an out-of-domain frame, which contains evaluations of trace and constraint composition polynomials at the out-of-domain point(s).
4. After the prover builds the OOD frame, the frame is serialized, and the public coin is reseeded with the resulting bytes. This imitates the prover sending the OOD frame to the verifier.
  a. After this step, the prover draws a set of random coefficients from the coin. These coefficients are used in building DEEP composition polynomial.
5. After the prover build the DEEP composition polynomial, the prover runs the commit phase of FRI protocol which consists of building FRI "layers" (via repeated application of degree respecting projection) from the evaluations of the DEEP composition polynomial. For each FRI layer, the prover build a Merkle tree, and the public coin is reseeded with the root of this tree. This commits the prover to each FRI layer.
  a. After each layer is committed, the prover draws a random element `alpha` from the pubic coin, and uses this element in construction of the next FRI layer.
6. After commit phase of FRI protocol is completed, the prover performs proof-of-work to find a nonce such that `hash(coin_seed || nonce)` satisfies the requested grinding factor. Once the nonce is found, the public coin is reseeded with its value.
  a. After this, the prover draws a set of random query positions from the public coin needed for the remainder of the protocol.


kevinlewi (approved):
LGTM and thanks for the detailed explanation!

Issue #10, opened 2021-04-28
[FRI] Implement custom folding factor


irakliyk:
Currently, our implementation of degree-preserving projection folds evaluation domains by a factor of 4. Would be good to have the ability to dynamically change the folding factor between values of 4, 8, and 16 to see what impact this makes on prover performance, proof size etc.

PR #28, opened 2021-06-01
Customizable FRI folding

irakliyk:
This PR builds on #27 to implement customizable FRI folding factor as described in #10 . Specifically:

- Implemented generic DRP procedure which can work with any folding factor. This function uses FFT-based interpolation - which is much faster than what we had before. The net effect is that FRI running time has been reduced by about 2x - though, this has fairly negligible impact on the overall proof generation time.
- Integrate generic DRP procedure into FRI prover and related modules. Even though DRP supports any folding factor, folding factors for FRI prover/verifier are limited to 4, 8, and 16.
- Removed quartic modules (generic code is much faster anyway), and also got rid of duplicated code in constraint evaluation (in the prover).


fri/src/prover/monolith/mod.rs
```diff
@@ -152,16 +156,17 @@ where
...
 
+            // sort of a static dispatch for folding_factor parameter
             let proof_layer = match folding_factor {
-                4 => build_proof_layer::<B, E, H, 4>(&self.layers[i], &positions),
-                8 => build_proof_layer::<B, E, H, 8>(&self.layers[i], &positions),
-                16 => build_proof_layer::<B, E, H, 16>(&self.layers[i], &positions),
+                4 => query_layer::<B, E, H, 4>(&self.layers[i], &positions),
+                8 => query_layer::<B, E, H, 8>(&self.layers[i], &positions),
+                16 => query_layer::<B, E, H, 16>(&self.layers[i], &positions),
                 _ => unimplemented!("folding factor {} is not supported", folding_factor),
```
Jasleen1, line 170: Is there a reason this is not generic? Or even, say, generic up to a power of 2? Is it because the hash function we use here only supports enough bits to fit up to 16 elements per hash? Could we replace it with a different function, since you use the type param H here anyway?

irakliyk: This is my way of getting around the need to translate the value of the folding factor which comes from the `options` struct (and is known only at runtime) and a value of the folding factor used by `query_layer` function (which must be known at compile time). 

The only other more generic way of doing it (that I can think of) is to add folding factor as a generic type parameter to the `options` struct - but I'm trying to avoid doing this for a number of reason.

PR #44, opened 2021-07-30
Make all crates (except examples) no-std compliant

Nashtare:
I attempted to make all crates `no-std` for my personal use, but assumed it could be useful in the main repository, following what has been said in https://github.com/novifinancial/winterfell/issues/37. The crates now rely on the `std` feature by default, which can be replaced by the `alloc` feature.

The `rand` dependency actually mentioned in the linked issue is not moved to dev-dependency, so that the `examples` crate can still provide the `merkle` and `lamport` examples without changes. I don't know if that's problematic / something that you really wanted to get rid of.
Note that the `examples` crate is not `no-std`, but it can take the prover and verifier dependencies with the `alloc` feature enabled only (at the cost of losing the two examples mentioned previously).

The CI includes now a check for all-crates with target `wasm32-unknown-unknown`.

If the pull-request is of any interest and requires change, please let me know what should be fixed!


irakliyk:
Thank you so much! This is really cool! Will review in the next day or so.

irakliyk:
@huitseeker - would love it if you could take a look as well.

crypto/Cargo.toml
```diff
@@ -23,13 +23,17 @@ harness = false
...
 [dependencies]
-utils = { path = "../utils", package = "winter-utils" }
-math = { path = "../math", package = "winter-math" }
-blake3 = "1.0"
-sha3 = "0.9"
+utils = { path = "../utils", package = "winter-utils", default-features = false }
+math = { path = "../math", package = "winter-math", default-features = false }
+blake3 = { version = "1.0", default-features = false }
+sha3 = { version = "0.9", default-features = false }
+hashbrown = { version = "0.9.1", optional = true, default-features = false, features = ["ahash", "inline-more"] }
```
irakliyk, line 36: I wonder if we should move `hashbrown` dependency to the `utils` create. Then, in the `utils` crate we could create `collections` module which could look something like this:

```Rust
#[cfg(feature = "alloc")]
pub use alloc::collections::{BTreeMap, BTreeSet};

#[cfg(feature = "alloc")]
pub use alloc::vec::Vec;

#[cfg(feature = "alloc")]
pub use hashbrown::HashMap;

#[cfg(feature = "std")]
pub use std::collections::{BTreeSet, HashMap};

#[cfg(feature = "std")]
pub use std::vec::Vec;
```

Then, in many places when we need to import these structs, we could just do `use utils::collections::*;`

crypto/src/merkle/mod.rs
```diff
@@ -6,8 +6,17 @@
 use crate::{errors::MerkleTreeError, hash::Hasher};
 use core::slice;
 use math::log2;
+
+#[cfg(feature = "alloc")]
```
irakliyk, line 10: If we have `utils::collections` module as described in the other comment, we could replace this with `use utils::collections::*;`

crypto/src/merkle/proofs.rs
```diff
@@ -4,9 +4,18 @@
 // LICENSE file in the root directory of this source tree.
 
 use crate::{errors::MerkleTreeError, Hasher};
-use std::collections::{BTreeMap, HashMap};
 use utils::{ByteReader, Deserializable, DeserializationError, Serializable};
 
+#[cfg(feature = "alloc")]
```
irakliyk, line 9: If we have `utils::collections` module as described in the other comment, we could replace this with use `utils::collections::*`;

irakliyk (changes requested):
Thank you again for submitting this! I'm still going through the code (will take me a bit longer since I haven't worked with `no_std` stuff before), but I left a few inline comments. 

Basically, wondering if we can replace a lot of the conditional import declarations with a single declaration to `utils::collections` module which would handle all the conditional imports internally.

crypto/src/merkle/mod.rs
```diff
@@ -6,8 +6,17 @@
...
+use alloc::collections::BTreeSet;
+#[cfg(feature = "alloc")]
+use hashbrown::HashMap;
+#[cfg(feature = "std")]
 use std::collections::{BTreeSet, HashMap};
```
irakliyk, line 15: If we have `utils::collections` module as described in the other comment, we could replace this with use `utils::collections::*`;

crypto/src/merkle/proofs.rs
```diff
@@ -4,9 +4,18 @@
...
+use alloc::collections::BTreeMap;
+#[cfg(feature = "alloc")]
+use hashbrown::HashMap;
+#[cfg(feature = "std")]
+use std::collections::{BTreeMap, HashMap};
```
irakliyk, line 14: If we have `utils::collections` module as described in the other comment, we could replace this with use `utils::collections::*`;

crypto/Cargo.toml
Nashtare: Will be cleaner indeed. Will do!

Nashtare:
The last commits addresses the comment about refactoring collections import into a `collections` module in the `utils` crate. I also put the `String`, `ToString` imports behind a tiny string module in `utils` (didn't want to pollute `utils/src/lib.rs`).

crypto/src/lib.rs
```diff
@@ -16,6 +16,19 @@
...
+#[cfg(not(any(feature = "std", feature = "alloc")))]
+compile_error!("Either feature \"std\" or \"alloc\" must be enabled for this crate.");
+
+#[cfg(feature = "alloc")]
+#[macro_use]
+extern crate alloc;
+
+#[cfg(feature = "std")]
+#[macro_use]
+extern crate std;
```
irakliyk, line 30: I'm curious why do we need this here (and in other crates). It seems to compile fine without these lines.

irakliyk (commented):
Looks great! And thank you for rebasing from the latest `main` (I was planning to do it, but you beat me to it :) )

I have just one inline question - but other than that, everything looks good to me.

air/src/air/tests.rs
```diff
@@ -11,6 +11,10 @@ use crate::{FieldExtension, HashFunction};
 use crypto::{hashers::Blake3_256, RandomCoin};
 use math::{fields::f128::BaseElement, get_power_series, log2, polynom, FieldElement, StarkField};
 use rand::{seq::SliceRandom, thread_rng};
+
+#[cfg(feature = "alloc")]
+use crypto::hashbrown::HashMap;
+#[cfg(feature = "std")]
```
irakliyk, line 17: Should this also be replaced with `use utils::collections::HashMap`

crypto/src/lib.rs
Nashtare: That's an old habit I have 🙃 I think it would have been needed pre Rust 2018 but `extern crate std` is already implicit I believe if `std` is enabled that's right. I'll remove them.

air/src/air/tests.rs
Nashtare: That's actually a miss! Thank you for catching it! (tests have not been made no-std, so technically we could just use `use std::collections::HashMap`). I can try moving tests as well to `no-std` if wanted.

Nashtare: Note that this would imply to make the `rand()` method of the `FieldElement` trait accessible in `no-std` as well (currently hidden behind the `std` feature

crypto/src/lib.rs
Nashtare: Actually (the above answer is only partially correct), we can remove it because the no_std requirement is only enabled conditionally through `#![cfg_attr(not(feature = "std"), no_std)]`

If we had instead `#![no_std]`, we would be forced to include the `extern crate std` conditionally, as otherwise we could not use the standard functions. As here we only conditionally require `no_std`, indeed that can be removed.

Nashtare: I'll keep the import and change the crate roots to #![no_std], I believe it's more common among no-std crates (I'm definitely not a specialist on the question 😅)

Nashtare: Sorry for the several force-push, changing to `#![no_std]` in the crate roots required to add some imports from utils::collections in test modules. Should be all good now

utils/Cargo.toml
```diff
@@ -14,7 +14,11 @@ edition = "2018"
...
 
 [features]
+alloc = ["hashbrown"]
 concurrent = ["rayon"]
+default = ["std"]
+std = []
 
 [dependencies]
+hashbrown = { version = "0.9.1", optional = true, default-features = false, features = ["ahash", "inline-more"] }
 rayon = { version = "1.5", optional = true }
```
irakliyk, line 24: Is there a reason we are not using the latest version of `hashbrown` here? (I believe the latest one is 0.11.2).

Also, nit: could we add a new line at the end of the file? (I know it wasn't there before either).

irakliyk (commented):
Awesome! Thank you so much for all the changes!

I have a couple more questions/comments (sorry for trickling them in!). I left one as an inline comment, and the other one is: should we make `alloc` and `concurrent` feature mutually exclusive? I'm not sure in what context it would make sense to have them both enabled.

utils/Cargo.toml
Nashtare: Hmm I'm not sure neither, I thought I had used the latest one. I may have missed up something when moving things here and there. Will correct it, sorry.
As for `alloc` and `concurrent`, I couldn't see which context would require them both neither, but as they are not incompatible per say, I didn't see the need to make them mutually exclusive. What would you prefer?

irakliyk: Yep - seems like the latest version of `hashbrown` is 0.11.2 rather than 0.9.1: https://crates.io/crates/hashbrown

Regarding `alloc` and `concurrent`, I think making them mutually exclusive could lead to less potential confusion/errors in the future. So, if not too difficult, would be great to add that to this PR.

Nashtare: Alright will do!

irakliyk (approved):
Looks great! Thank you again for the amazing work!

I'll try to get another set of eyes on this and will merge tomorrow.

crypto/src/lib.rs
```diff
@@ -16,6 +16,22 @@
 //!   [RandomCoin] implementation uses a cryptographic hash function to generate pseudo-random
 //!   elements form a seed.
 
+#![no_std]
+
+#[cfg(not(any(feature = "std", feature = "alloc")))]
+compile_error!("Either feature \"std\" or \"alloc\" must be enabled for this crate.");
+
+#[cfg(all(feature = "alloc", feature = "concurrent"))]
+compile_error!("This crate does not support features \"alloc\" and \"concurrent\" simultaneously.");
```
huitseeker, line 25: Since rayon is not `no_std` compatible, wouldn't it be simpler to make the `concurrent` feature imply `std` and then branch only on `std` vs. `no_std + alloc`? The less conditions the better.

fri/src/lib.rs
```diff
@@ -61,6 +61,22 @@
...
 //! * Swastik Kooparty's [talk on DEEP-FRI](https://www.youtube.com/watch?v=txo_kPSn59Y&list=PLcIyXLwiPilWvjvNkhMn283LV370Pk5CT&index=6)
 
+#![no_std]
+
+#[cfg(not(any(feature = "std", feature = "alloc")))]
+compile_error!("Either feature \"std\" or \"alloc\" must be enabled for this crate.");
+
+#[cfg(all(feature = "alloc", feature = "concurrent"))]
+compile_error!("This crate does not support features \"alloc\" and \"concurrent\" simultaneously.");
+
```
huitseeker, line 71: same as above on only having `std` vs `no_std` + `alloc`

huitseeker (commented):
I'd love a bit more detail in the Readme as to how this caters to the `no_std`+ `alloc` use case, not the `no_std` use case, so that embedded developers don't think this is an easy pickup.

@nashtare Have you run cargo no-std-check to get a rough idea of how quickly dependencies pull std?

Nashtare:
@huitseeker Yes, READMEs may need some more explanations about the requirement for the global allocator.

As for cargo-no-std-check it runs fine in all crates except example which hasn't been made no-std (currently complaining about lazy-static which is a dependency of structopt and is not brought with its "spin_no_std" feature)

crypto/src/lib.rs
Nashtare: That's a possibility indeed. One thing that may need clarity, either in READMEs or through a conditional compiler error is if both alloc and std are allowed simultaneously. We currently require at least 1 of them, but having both will break compilation of the winter-utils crate, so we may need to have this second conditional attribute anyway (changed to `#cfg[not(all(feature = "alloc", feature = "std"))]`)

irakliyk:
I can add more detail to readmes/docs in a separate PR. But I think it would be good to add checks for various feature combinations to this PR. I'm thinking of the following:

* We should not allow `alloc` and `std` simultaneously.
* `concurrent` should imply `std`.

Basically, the only possible combinations of features should be `alloc + no_std`, or `std`, or `std + concurrent`.

irakliyk:
Looks great! Thanks again for doing all this work!

Nashtare:
Happy to help!

PR #45, opened 2021-07-31
Malleable field elements

irakliyk:
This PR introduces a concept of malleable field elements - i.e., elements which might have redundant internal representation for the same canonical representation. For example, this is the case for field elements where internally elements are stored in the range `[0, 2M)` (rather than `[0, M)`) which we use for our `f62` field (the `f128` field does not have this issue because we don't use Montgomery arithmetic for it).

This also allows us to hash elements consistently by first normalizing their internal representation, and only then hashing them. We do this for malleable elements only, since non-malleable elements don't need to be normalized, and thus, we can hash them directly.

We also make sure that we always write elements into the proof in canonical representation. This avoids issues when trying to deserialize elements where canonical and internal representations differ.

The above fixes should resolve the problems mentioned in https://github.com/novifinancial/winterfell/issues/13#issuecomment-888018333.

### DIFF
--- a/math/src/field/extensions/quadratic.rs
+++ b/math/src/field/extensions/quadratic.rs
@@ -46,6 +46,7 @@ impl<B: StarkField> FieldElement for QuadExtensionA<B> {
     type BaseField = B;
 
     const ELEMENT_BYTES: usize = B::ELEMENT_BYTES * 2;
+    const IS_MALLEABLE: bool = B::IS_MALLEABLE;
     const ZERO: Self = Self(B::ZERO, B::ZERO);
     const ONE: Self = Self(B::ONE, B::ZERO);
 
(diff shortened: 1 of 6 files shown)


math/src/field/extensions/quadratic.rs
```diff
@@ -120,6 +121,11 @@ impl<B: StarkField> FieldElement for QuadExtensionA<B> {
         let result = B::prng_vector(seed, n * 2);
         Self::base_to_quad_vector(result)
     }
+
+    fn normalize(&mut self) {
```
kchalkias, line 125: Why do we need mutability here and not deserialize/construct the elements directly in their canonical form? So no object can even be non-canonical in the first place.

irakliyk: There are a couple of reasons:

First, when we use Montgomery reduction/multiplication, it is expensive to convert to and from Montgomery form all the time - so, we try to keep elements in Montgomery form as long as possible. This means that the internal representation of elements is in Montgomery form and we convert to/from Montgomery form only when an element is created or serialized.

The above doesn't require redundant representation of elements - every internal representation maps to a unique canonical representation. However, to improve efficiency we can encode elements in redundant form. That is, each canonical representation can have more than one internal representation (in the case of `f62` field, there are two internal representations per canonical representation since the range for internal elements is `[0, 2M)`). For example, this allows us to avoid the last conditional check in the multiplication procedure (so, multiplication becomes fully branchless).

PR #47, opened 2021-08-02
Winterfell crate

irakliyk:
This PR will add `winterfell` create which re-exports prover and verifier crates. The primary purpose of this is to simplify crate imports in external projects.

- [x] Crete `winterfell` create
- [x] Update `examples` create
- [x] Update documentation


examples/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "winterfell"
+name = "examples"
```
kchalkias, line 2: I would prefer a better name here, i.e., applications or something similar (I prefer gadgets, but unfortunately they are not necessarily reusable, so let's pick something else).

winterfell/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+//!    result
+//! }
+//! ```
+//!
+//! This computation starts with an element in a finite field and then, for the specified number
+//! of steps, cubes the element and adds value `42` to it.
+//!
+//! Suppose, we run this computation for a million steps and get some result. Using STARKs we can
+//! prove that we did the work correctly without requiring any verifying party to re-execute the
+//! computation. Here is how to do it.
```
kchalkias, line 81: replace . with :

kchalkias (approved):
Approved, minor comments (mainly to rename "examples" in another PR)

winterfell/src/lib.rs
irakliyk: Thanks! Fixed!

examples/Cargo.toml
irakliyk: I thing `gadgets` makes sense to me. Will rename in another PR.

Issue #37, opened 2021-07-19
[math] Remove rand crate dependency


irakliyk:
Ideally, this should be a dev dependency as we use methods which rely on `rand` crate almost exclusively for testing and benchmarking.

Nashtare:
Out of curiosity (as the rand dependency is part of it by default), is some no-std support in the pipeline of this project?

irakliyk:
Yes, we'd like to do this at least for the verifier. Though, hard to tell when it will actually happen.

PR #55, opened 2021-08-21
Dependency cleanup

irakliyk:
This PR removes unnecessary dependencies from the project (this includes addressing #37 ).

- [x] Move rand functions used for testing into a separate crate and remove `rand` as a dependency from all crates.
- [x] Remove `hashbrown` dependency (by using `BTreeMap` instead of `HashMap` everywhere).
- [x] Update documentation

As a result of this PR, the project has only 4 external dependancies (not counting the ones in the example crate):
* `blake3` and `sha3` used in `crypto` create
* `log` used in the `prover` crate
* `rayon` used only when `concurent` feature is enabled


kevinlewi (approved):
Great!

Issue #54, opened 2021-08-20
Supporting RAPs


pgrinaway:
Hi,

First, this is a really nice project. Thank you! 

Onto my question: the recent [Cairo paper](https://eprint.iacr.org/2021/1063.pdf) describes using Randomized AIRs with Preprocessing (RAPs) for Cairo's memory. While they don't describe in detail how they modify the STARK protocol, it appears to allow the prover to first commit to a set of trace columns, then use that commitment to generate randomness that can be used in an additional set of trace columns. The main use case here is that with such capability, local constraints (as in AIR) can be used to verify global properties (such as that one column is a permutation of another, which is what is done in Cairo's memory).

Is this of interest to anyone here? I'd really appreciate some pointers or advice on how the API for winterfell might be best modified to support this. Of course, if the modifications are of use to anyone here I'd be happy to contribute them.

Thanks and have a nice weekend

irakliyk:
This would be an awesome addition to the library and will enable lots of interesting use cases. I haven't really dug into what machinery would be required to implement this - so, my thoughts on potential interfaces are rather preliminary. But I can think of couple of approaches:

One approach might be to modify the interface of [ExecutionTrace](https://github.com/novifinancial/winterfell/blob/main/prover/src/trace/execution_trace.rs) struct so that the user can build columns in stages. This could be tricky because we use [ProveChannel](https://github.com/novifinancial/winterfell/blob/main/prover/src/channel.rs) struct to draw randomness, and prove channel is constructed after the call to the prover's `prove()` function. So, it might potentially require fairly significant interface changes to the library.

Another approach might be to introduce a concept of "lookup tables". These tables could be used to represent read-only memory (like in Cairo), but also other things (e.g., for checking if a value in a valid range). These tables could then be passed to the prover (together with some other info) and the prover can figure out how to build the execution trace from them.

I kind of like the second approach, but there are still quite a few things to figure. For example:
* I imagine lookup tables could be public or private. In case of public tables, it might make sense to make them a part of [AIR](https://github.com/novifinancial/winterfell/blob/main/air/src/air/mod.rs#L145) definition. In case of private tables, they would need to be passed to prover directly (potentially as one of the arguments to [prove()](https://github.com/novifinancial/winterfell/blob/main/prover/src/lib.rs#L117) function).
* How would the user define the rest of the execution trace in relation to these lookup tables? Right now, an execution trace is just a matrix of values, would it need to become more complex than that?
[...]

PR #58, opened 2021-09-06
Implement execution trace construction via TraceBuilder trait

irakliyk:
This PR will change the way execution traces are constructed. This will be the first step in implementing RAPs (see #54 ).

Right now, the user is responsible for constructing an execution trace and submitting it to the prover. We have an `ExecutionTrace` struct which helps the user with this task. However, trace construction happens outside of prover logic. This works fine when we construct the trace in one go (as is done now), but if we need to construct the trace in stages (and each stages needs to be processed by the prover before the next stage is constructed) we run into a problem.

One of the ways to solve this problem is to move trace construction logic into the prover. With this approach, the user provides a description of how the trace should be built to the prover, and the prover executes the logic. For the description of the logic, we can introduce `TraceBuilder` trait.

There are a few nuances associated with this approach. The main one is that for some computations a full set of public inputs may not be known prior to trace constructions. Thus, `TraceBuilder` interface must also define the logic of how public inputs can be derived from a given execution trace. Also, the prover must return not just the STARK proof, but also the set of public inputs derived from the execution trace.

Overall, this PR consists of the following tasks:

- [x] Add `TraceBuilder` trait.
- [x] Update all examples to use `TraceBuilder` trait.
- [ ] Update prover to return public inputs in addition to the STARK proof.
- [x] Get rid of `ExecutionTrace` struct since it is no longer needed.
- [ ] Update documentation


irakliyk:
Superseeded by #69

PR #59, opened 2021-09-16
Adding small fields

Jasleen1:
Added support for small prime fields to be used for testing under the field crate. See the `field/smallprimefield/tests.rs` for example usage but in general, implements the `FieldElement` trait.


math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+    collections::Vec,
+    string::{String, ToString},
+    AsBytes, ByteReader, ByteWriter, Deserializable, DeserializationError, Randomizable,
+    Serializable,
+};
+
+// Number of bytes needed to represent field element
+const ELEMENT_BYTES: usize = mem::size_of::<u64>();
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
```
irakliyk, line 22: Would be great to add some comments here (nothing fancy) about the purpose of this field etc.

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+pub struct GenericPrimeFieldElt<const M: u64, const G: u64> {
```
irakliyk, line 23: I wonder if we should keep the same naming convention as with other field - i.e. call this `BaseElement`?

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+    pub(crate) value: u64,
+}
+
+impl<const M: u64, const G: u64> GenericPrimeFieldElt<M, G> {
+    // TODO: this may need to be something else, e.g. its own type param
+    const fn get_twoadic_root() -> Self {
+        Self::new(G)
+    }
```
irakliyk, line 31: Is this function needed? I think it can be replaced in favor of using `Self::new(G)` directly.

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    const fn get_modulus_bits() -> u32 {
+        64u32 - M.leading_zeros()
+    }
+
+    /// Creates a new field element from a u128 value. If the value is greater than or equal to
+    /// the field modulus, modular reduction is silently preformed. This function can also be used
+    /// to initialize constants.
```
irakliyk, line 39: I think this comment needs to be updated.

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+    pub const fn new(value: u64) -> Self {
+        let val = value % M; //if  == Ordering::Less { value } else { };
+        Self { value: val }
+    }
+
+    /// Returns field element converted to u64 representation.
+    pub fn as_u64(&self) -> u64 {
+        self.value
+    }
```
irakliyk, line 48: Can we use `as_int()` method of `StarkField` trait instead of this one? If so, we can probably delete this method.

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+    #[allow(unused)]
+    unsafe fn get_power_series(b: Self, n: usize) -> Vec<Self> {
+        let mut result = utils::uninit_vector(n);
+        result[0] = Self::ONE;
+        for i in 1..result.len() {
+            result[i] = result[i - 1] * b;
+        }
+        result
+    }
+}
```
irakliyk, line 59: Can we use `get_power_series()` function from the `utils` module instead of this one? If so, we can probably delete this function.

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+        // We also use u64 to make sure the memory is aligned correctly for our element size.
+        debug_assert_eq!(Self::ELEMENT_BYTES, mem::size_of::<u64>());
+        let result = vec![0u128; n];
+
+        // translate a zero-filled vector of u128s into a vector of base field elements
+        let mut v = core::mem::ManuallyDrop::new(result);
+        let p = v.as_mut_ptr();
+        let len = v.len();
+        let cap = v.capacity();
+        unsafe { Vec::from_raw_parts(p as *mut Self, len, cap) }
```
irakliyk, line 118: You can probably return `vec![0; n]` from here.

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+        r
+    }
+}
+
+impl<const M: u64, const G: u64> StarkField for GenericPrimeFieldElt<M, G> {
+    type QuadExtension = QuadExtensionA<Self>;
+    const MODULUS: Self::PositiveInteger = M;
+    const MODULUS_BITS: u32 = Self::get_modulus_bits();
+
+    const GENERATOR: Self = Self::new(G);
```
irakliyk, line 168: I'm not sure this is correct. The generator is supposed to be the primitive element in the field. If it can't be easily inferred from the modulus, maybe pass it as another generic parameter?

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    const TWO_ADICITY: u32 = 2;
```
irakliyk, line 170: Can we compute this from `G`? Maybe add a const function for this to `GenericPrimeFieldElt` impl?

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    const TWO_ADIC_ROOT_OF_UNITY: Self = Self::get_twoadic_root();
```
irakliyk, line 172: I think `Self::get_twoadic_root()` can be replaced with just `Self::new(G)`

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    fn get_root_of_unity(n: u32) -> Self {
+        super::get_prime_field_root_of_unity(n, Self::MODULUS)
```
irakliyk, line 175: Can we just reply on the generic implementation of this function available in the `StarkField` trait?

math/src/field/smallprimefield/generic_prime_field_elt.rs
```diff
@@ -0,0 +1,426 @@
...
+        Self::MODULUS.to_le_bytes().to_vec()
+    }
+
+    fn as_int(&self) -> Self::PositiveInteger {
+        self.value
+    }
+
+    // fn from_int(value: u64) -> Self {
+    //     SmallFieldElement37::new(value)
+    // }
```
irakliyk, line 188: We can probably delete this.

math/src/field/smallprimefield/mod.rs
```diff
@@ -0,0 +1,26 @@
+use core::convert::TryInto;
+
+use crate::StarkField;
+
+pub mod generic_prime_field_elt;
```
irakliyk, line 5: I'm not sure we need this extra level of module nesting. Could we put the code from `generic_prime_field_elt.rs` into this file directly?

math/src/field/smallprimefield/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+        small_field_size
+    );
+    assert!(
+        small_field_size % n == 0,
+        "Order invalid for field size {}",
+        small_field_size
+    );
+    let power = small_field_size / n;
+    E::exp(E::GENERATOR, power.into())
+}
```
irakliyk, line 26: Can we remove this in favor of `StarkField::get_root_of_unity()` function?

irakliyk (changes requested):
Thank you! Very good solution for handling field modulus via const generics! I left a few inline comments - mostly code cleanup. But other than that, it looks great!

math/src/field/smallprimefield/mod.rs
Jasleen1: See reply to https://github.com/novifinancial/winterfell/pull/59#discussion_r711272855

math/src/field/smallprimefield/mod.rs
Jasleen1: Yes, thanks!

math/src/field/smallprimefield/generic_prime_field_elt.rs
Jasleen1: Done!

math/src/field/smallprimefield/generic_prime_field_elt.rs
Jasleen1: This wasn't actually the right value. Updated now.

math/src/field/smallprimefield/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+        debug_assert_eq!(Self::ELEMENT_BYTES, mem::size_of::<u64>());
+        let result = vec![0u128; n];
+
+        // translate a zero-filled vector of u128s into a vector of base field elements
+        let mut v = core::mem::ManuallyDrop::new(result);
+        let p = v.as_mut_ptr();
+        let len = v.len();
+        let cap = v.capacity();
+        unsafe { Vec::from_raw_parts(p as *mut Self, len, cap) }
+    }
```
irakliyk, line 154: Since we are not concerned about performance here that much, we can probably just delete this specialized implementation.

math/src/field/smallprimefield/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+            return Err(DeserializationError::InvalidValue(format!(
+                "number of bytes ({}) does not divide into whole number of field elements",
+                bytes.len(),
+            )));
+        }
+
+        let p = bytes.as_ptr();
+        let len = bytes.len() / Self::ELEMENT_BYTES;
+
+        if (p as usize) % mem::align_of::<u128>() != 0 {
```
irakliyk, line 132: Since the backing type is `u64`, we can probably relax the alignment a bit and make sure the values align on `u64` boundaries.

math/src/field/smallprimefield/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+// SMALL PRIME FIELD ELEMENT
+// ================================================================================================
+
+/// Represents a base field element with a small modulus.
+/// Primarily for testing, so we can hand calculate expected outputs.
+/// Note that all the operations here are naive field operations without optimizations.
+/// We also don't end up needing to use the twoadic roots anymore,
+/// since we can factor MODULUS - 1 easily. Meaning, any factor root of MODULUS - 1 can be found.
+/// Internal values are stored in their canonical form in the range [0, MODULUS). The backing type is
+/// `u64`.
```
irakliyk, line 36: Looks great! A couple of minor comments:

1. Would be great to explain the meaning of generic parameters here. For example, initially, I thought that `G` was the 2-addic root of unity, but it seems like it is actually the primitive element in the field.
2. We should add a blank line in the comments between the first line and the rest. This way, the comment will be formatted better in the docs (you can run `cargo doc --open` to generate and view documentation).
3. Nit: we should get rid of the blank line between the comment and the struct.

irakliyk (changes requested):

rubyazz (commented):
cool stuff

bobbinth:
Closing as stale.

irakliyk:
Closing as stale.

PR #64, opened 2021-11-15
Fix cubic extension

Nashtare:
There are currently two issues when proving with field extensions:

- the first one, rather minor, is when sampling from the public coin. It may happen for cubic extension that we run out of trials by lack of luck. Increasing the number of mentioned trials before outputting an error solved this.
- the second one is with respect to the cubic extension. Currently, FriProof serialization is expecting a power-of-two remainder length, which is not the case when proving in the cubic extension. The second commit fixes this.

Both errors are visible on my temporary branch https://github.com/Nashtare/winterfell/tree/f62-example where I added an example of Fibonacci with `f62`, followed by the two fixes above (go back in the history when running `./target/release/winterfell -e 3 fib-f62` to see the actual errors prior fix)

### DIFF
--- a/crypto/src/random/mod.rs
+++ b/crypto/src/random/mod.rs
@@ -174,13 +174,13 @@ impl<B: StarkField, H: Hasher> RandomCoin<B, H> {
     /// Returns the next pseudo-random field element.
     ///
     /// # Errors
-    /// Returns an error if a valid field element could not be generated after 100 calls to the
+    /// Returns an error if a valid field element could not be generated after 1000 calls to the
     /// PRNG.
     pub fn draw<E>(&mut self) -> Result<E, RandomCoinError>
     where
         E: FieldElement<BaseField = B>,
     {
-        for _ in 0..200 {
+        for _ in 0..1000 {
             // get the next pseudo-random value and take the first ELEMENT_BYTES from it
             let value = self.next();
             let bytes = &value.as_bytes()[..E::ELEMENT_BYTES as usize];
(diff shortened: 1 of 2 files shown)


crypto/src/random/mod.rs
```diff
@@ -180,7 +180,7 @@ impl<B: StarkField, H: Hasher> RandomCoin<B, H> {
     where
         E: FieldElement<BaseField = B>,
     {
-        for _ in 0..200 {
+        for _ in 0..500 {
```
irakliyk, line 183: Let's bump it up to 1000, and also let's update the documentation comment above.

irakliyk (changes requested):
Thank you for fixing these! Looks great! I left just one small comment.

One note: the ideal solution to `f62` field random element generator running out of tries would be to fix the `from_random_bytes()` method on `f62` field element. But I think this approach works fine for now.

crypto/src/random/mod.rs
Nashtare: Updated!

Nashtare:
Regarding your comment, indeed fixing the `from_random_bytes()` would be cleaner but requires potential changes of the other small field `f64` and potential future ones, while that change fixes it generally (this doesn't happen that often though, so no big slowdown expected in practice)

irakliyk (approved):

irakliyk:
> Regarding your comment, indeed fixing the `from_random_bytes()` would be cleaner but requires potential changes of the other small field `f64` and potential future ones as well, while this change fixes it generally (this doesn't happen that often though, so no big slowdown expected in practice)

Agreed! Thank you again!

PR #65, opened 2021-11-16
Fix some typos

Nashtare:
I didn't attach it to the last PR (and I'm not sure it deserves a PR honestly) but here are some typo fixes, mostly from `winter_math`.
There is a similar typo in the Changelog (`filed` -> `field`) but I didn't touch it, as I wasn't sure if I could change previous entries of the Changelog.


irakliyk (approved):
Thank you for fixing all of these!

I'll fix the typos in the changelog next time I update it.

PR #68, opened 2021-12-26
Cargo updates

irakliyk:
This PR updates all cargo files to switch to Rust 2021 edition and update min `rustc` version to 1.57


Jasleen1 (approved):
Looks good, thank you!

PR #69, opened 2021-12-28
Prover trait

irakliyk:
This PR supersedes #58. Moving trace building into the prover seems like too big of a change and may sacrifice flexibility. Instead, this PR takes the following approach:

We introduce a `Prover` trait instead of a single `prover::prove()` function. We also split `ExecutionTrace` struct into two parts:
1. `Trace` trait, which is required for the prover to work.
2. `TraceTable` struct, which implements `Trace` trait can be used to build trace tables which should suffice for most use cases.

The idea of the above is that users of the library can provide different implementations of `Trace` trait which may better work for their use cases, and in case a generic trace table works, they can use `TraceTable` struct.

In the context of RAPs, we can update either `Prover` or `Trace` traits to expose methods needed for construction of auxiliary columns. Which one is more appropriate, can be decided later on.

Tasks for this PR:

- [x]  Implement `Prover` trait
- [x] Split `ExecutionTrace` struct into `Trace` trait and `TraceTable` struct.
- [x] Expose public input getter  in the `Prover` trait.
- [x] Update examples
- [x] Update documentation

### DIFF
--- a/prover/src/trace/mod.rs
+++ b/prover/src/trace/mod.rs
@@ -7,7 +7,7 @@ use super::StarkDomain;
 
 mod trace_lde;
 use air::{Air, EvaluationFrame, TraceInfo};
-use math::{fft, polynom, StarkField};
+use math::{fft, polynom, FieldElement, StarkField};
 pub use trace_lde::TraceLde;
 
 mod poly_table;
@@ -28,7 +28,9 @@ mod tests;
 // ================================================================================================
 
 // TODO: add docs
-pub trait Trace<B: StarkField>: Sized {
+pub trait Trace: Sized {
+    type BaseField: StarkField;
+
     // REQUIRED METHODS
     // --------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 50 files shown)


Jasleen1 (approved):
Great work and especially grateful for your thorough documentation!

prover/src/trace/mod.rs
```diff
@@ -4,15 +4,200 @@
...
+
+// TRACE TRAIT
+// ================================================================================================
+/// Defines an execution trace of a computation.
+///
+/// Execution trace can be reduced to a two-dimensional matrix in which each row represents the
+/// state of a computation at a single point in time and each column corresponds to an algebraic
+/// register tracked over all steps of the computation.
+///
+/// Building a trace is required for STARK proof generation. An execution tace of a specific
```
dbeaverfb, line 35: trace

dbeaverfb (approved):
A nice, clean, clear refactoring - lgtm!

prover/src/trace/mod.rs
irakliyk: Thank you! Fixed.

PR #71, opened 2022-01-13
Rp64_256 improvements

irakliyk:
This PR makes a few minor updates to Rp64_256 hash function implementation:

* Implemented digest to array conversion for Rp64_256 digest.
* Makes `apply_permutation()` and `apply_round()` functions public.

### DIFF
--- a/crypto/src/hash/rescue/rp64_256/mod.rs
+++ b/crypto/src/hash/rescue/rp64_256/mod.rs
@@ -133,7 +133,7 @@ impl Hasher for Rp64_256 {
             state[i] += BaseElement::new(u64::from_le_bytes(buf));
             i += 1;
             if i % RATE_WIDTH == 0 {
-                apply_permutation(&mut state);
+                Self::apply_permutation(&mut state);
                 i = 0;
             }
         }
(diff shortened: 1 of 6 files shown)


crypto/src/hash/rescue/rp64_256/mod.rs
```diff
@@ -218,102 +218,116 @@ impl ElementHasher for Rp64_256 {
...
 
-    // compute base^100100100100
-    let t4 = exp_acc::<BaseElement, STATE_WIDTH, 6>(t3, t3);
+        // compute base^100100100100
+        let t4 = exp_acc::<BaseElement, STATE_WIDTH, 6>(t3, t3);
 
-    // compute base^100100100100100100100100
-    let t4 = exp_acc::<BaseElement, STATE_WIDTH, 12>(t4, t4);
+        // compute base^100100100100100100100100
+        let t4 = exp_acc::<BaseElement, STATE_WIDTH, 12>(t4, t4);
```
dbeaverfb, line 317: This is pure refactor so it's okay, but it's also unsettling to read and perhaps to maintain: overwriting t4 makes the reviewer (and later coder?) wonder if there's a mistake.

After reading carefully, it looks like the first instance of t4 (100100100100) isn't directly used later, explaining why one might not need it. If there's no performance issue with using unique varnames, the reviewer wouldn't have to be suspicious.

dbeaverfb (approved):
Absolutely wonderful comments - so easy to read and understand!

One very optional pre-existing style remark.

crypto/src/hash/rescue/rp64_256/mod.rs
irakliyk: Good point! Fixed!

PR #72, opened 2022-01-19
Improvements to RP64_256 hash function

irakliyk:
This PR modifies a few things about the RP64_256 hash function. Specifically:

- Adds ability to convert the digest to an array of bytes.
- Changes how the sponge state is interpreted. Specifically, we move the capacity portion to the front of the state.

In the previous instantiation the state was arranged as follows:
- Elements 0 through 7 were reserved for the rate.
- Elements 8 through 11 were reserved for the capacity.
- The output of the hash was read from the first four elements of the rate (elements 0, 1, 2, and 3).

In the new instantiation the state is arranged as follows:
- Elements 4 through 11 are reserved for the rate.
- Elements 0 through 3 are reserved for the capacity.
- The output of the hash is still read from the first four elements of the rate (which are now elements 4, 5, 6, and 7).

This arrangement of the state makes using this hash function a bit easier on stack-based machines.


crypto/README.md
```diff
@@ -11,6 +11,7 @@ This crate contains modules with cryptographic operations needed in STARK proof
 ### Rescue hash function implementation
 Rescue hash function is implemented according to the Rescue Prime [specifications](https://eprint.iacr.org/2020/1143.pdf) with the following exception:
 * We set the number of rounds to 7, which implies a 40% security margin instead of the 50% margin used in the specifications (a 50% margin rounds up to 8 rounds). The primary motivation for this is that having the number of rounds be one less than a power of two simplifies AIR design for computations involving the hash function.
+* For instantiation `RP64_256`, we use the first 4 elements of the state (rather than the last 4 elements of the sate) for capacity and the remaining 8 elements for rate. The output of the hash function comes from the first four elements of the rate portion of the state (elements 4, 5, 6, and 7).
```
dbeaverfb, line 14: 'state' not 'sate'

crypto/src/hash/rescue/rp64_256/mod.rs
```diff
@@ -49,8 +63,11 @@ const INV_ALPHA: u64 = 10540996611094048183;
 ///   margin used in the specifications (a 50% margin rounds up to 8 rounds). The primary
 ///   motivation for this is that having the number of rounds be one less than a power of two
 ///   simplifies AIR design for computations involving the hash function.
+/// * We use the first 4 elements of the state (rather than the last 4 elements of the sate) for
```
dbeaverfb, line 66: 'state'

crypto/README.md
dbeaverfb, line 14: This probably should enjoy a comment on security.

We are effectively applying a fixed bit-permutation P before and after the f_XLIX permutation. The results will not match what happens when P is the identity, but we assert without proof that if f_XLIX is indistinguishable from a random permutation then  P^{-1} f_XLIX P should likewise be indistinguishable.

dbeaverfb:
In my brief reading, this all looks fine modulo three concerns:

1. This isn't fully standard rescue prime (but wasn't, perhaps, anyway, because of the 7-round modifiction). But the atypical sponge embedding should make no difference in security.

2. Why aren't there any changes to the tests?  I would imagine the results will look radically different now.

3. Longer term: while this is fine for any (P,V) pair of the same "generation," how do we handle making hash-changes like this in general? Versioning somewhere?  Earlier, stored proofs should fail with modern verifiers after this.

Approval is optimistic based on (2) being resolved through fixing either my misunderstanding or the test code. :)

dbeaverfb (approved):

crypto/README.md
irakliyk: fixed!

crypto/src/hash/rescue/rp64_256/mod.rs
irakliyk: Fixed!

crypto/README.md
irakliyk: Makes sense! I've included a slightly simplified version of your text in the comments. Thank you!

irakliyk:
Thank you! I've addressed the comments, and to answer the questions:

1. The tests we currently have test only the correctness of the permutation - but not the hash function. It is a good idea to add more tests which test the hash function end-to-end, but we should probably do this in a separate PR.

2. Longer-term changes like this shouldn't happen :) My hope is that "production" version of such hash functions would be built to some external specs which also provide test vectors to test against.

PR #73, opened 2022-01-25
Improve inversion in f64 field

irakliyk:
This PR improves inversion in the `f64` field by replacing the current implementation with the method based on Fermat’s little theorem. Specifically, we fully unroll the multiplication by `M - 2` which allows us to achieve modular in version in 72 multiplications.

The net result is that this implementation is about 2.5x faster than the previous one.

### DIFF
--- a/math/src/field/f64/mod.rs
+++ b/math/src/field/f64/mod.rs
@@ -99,51 +99,33 @@ impl FieldElement for BaseElement {
     #[inline]
     #[allow(clippy::many_single_char_names)]
     fn inv(self) -> Self {
-        let x = self.as_int();
+        // compute base^(M - 2) using 72 multiplications
+        // M - 2 = 0b1111111111111111111111111111111011111111111111111111111111111111
 
-        if x == 0 {
-            return Self::ZERO;
-        };
+        // compute base^11
+        let t2 = self.square() * self;
 
-        let mut a: u128 = 0;
-        let mut u: u128 = if x & 1 == 1 {
-            x as u128
-        } else {
-            (x as u128) + (M as u128)
-        };
-        let mut v: u128 = M as u128;
-        let mut d = (M as u128) - 1;
+        // compute base^111
+        let t3 = t2.square() * self;
 
-        while v != 1 {
-            while v < u {
-                u -= v;
-                d += a;
-                while u & 1 == 0 {
-                    if d & 1 == 1 {
-                        d += M as u128;
-                    }
-                    u >>= 1;
-                    d >>= 1;
-                }
-            }
+        // compute base^111111 (6 ones)
+        let t6 = exp_acc::<3>(t3, t3);
 
-            v -= u;
-            a += d;
+        // compute base^111111111111 (12 ones)
+        let t12 = exp_acc::<6>(t6, t6);
 
-            while v & 1 == 0 {
-                if a & 1 == 1 {
-                    a += M as u128;
-                }
-                v >>= 1;
-                a >>= 1;
-            }
-        }
+        // compute base^111111111111111111111111 (24 ones)
+        let t24 = exp_acc::<12>(t12, t12);
 
-        while a > (M as u128) {
-            a -= M as u128;
-        }
+        // compute base^1111111111111111111111111111111 (31 ones)
+        let t30 = exp_acc::<6>(t24, t6);
+        let t31 = t30.square() * self;
 
-        Self(a as u64)
+        // compute base^111111111111111111111111111111101111111111111111111111111111111
+        let t63 = exp_acc::<32>(t31, t31);
+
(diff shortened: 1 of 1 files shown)


math/src/field/f64/mod.rs
```diff
@@ -99,51 +99,33 @@ impl FieldElement for BaseElement {
...
+        let t12 = exp_acc::<6>(t6, t6);
 
-        while a > (M as u128) {
-            a -= M as u128;
-        }
+        // compute base^111111111111111111111111
+        let t24 = exp_acc::<12>(t12, t12);
 
-        Self(a as u64)
+        // compute base^111111111111111111111111
```
kchalkias, line 120: replace with // compute base^111111111111111111111111111111 (30 ones)

kchalkias (approved):

math/src/field/f64/mod.rs
```diff
@@ -99,51 +99,33 @@ impl FieldElement for BaseElement {
     #[inline]
     #[allow(clippy::many_single_char_names)]
     fn inv(self) -> Self {
-        let x = self.as_int();
+        // compute base^(M - 2) using 72 multiplications
+        // M - 2 = 0b1111111111111111111111111111111011111111111111111111111111111111
```
dbeaverfb, line 104: // t_i represents base^{2^i}

math/src/field/f64/mod.rs
```diff
@@ -569,3 +551,13 @@ fn mod_reduce(x: u128) -> u64 {
     let (result, over) = tmp0.overflowing_add(tmp1);
     result.wrapping_add(E * (over as u64))
 }
+
+/// Squares the base N number of times and multiplies the result by the tail value.
```
dbeaverfb, line 555: So: logarithmic terms:      (logx << N) + logy.

(No need to use this as a comment, just confirming.)

math/src/field/f64/mod.rs
```diff
@@ -569,3 +551,13 @@ fn mod_reduce(x: u128) -> u64 {
...
+#[inline(always)]
+fn exp_acc<const N: usize>(base: BaseElement, tail: BaseElement) -> BaseElement {
+    let mut result = base;
+    for _ in 0..N {
+        result = result.square();
+    }
+    result * tail
+}
```
dbeaverfb, line 563: This is very nicely written and optimized.

I only see unit tests for 0 and 1.  Maybe I am missing something.  To know that this isn't a regression, maybe a unit test with 2 or some other larger number would help.  Even using the values calculated from the pre-existing code would be fine.

dbeaverfb (approved):

math/src/field/f64/mod.rs
bobbinth: fixed!

math/src/field/f64/mod.rs
bobbinth: kind of: up until `t31` that's how it works. beyond that, the number is just for the number of bits of the exponent included in the temp value.

math/src/field/f64/mod.rs
```diff
@@ -569,3 +551,13 @@ fn mod_reduce(x: u128) -> u64 {
     let (result, over) = tmp0.overflowing_add(tmp1);
     result.wrapping_add(E * (over as u64))
 }
+
+/// Squares the base N number of times and multiplies the result by the tail value.
+#[inline(always)]
+fn exp_acc<const N: usize>(base: BaseElement, tail: BaseElement) -> BaseElement {
```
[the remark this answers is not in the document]
bobbinth: I was going for something like "exponent_accumulate" :)

math/src/field/f64/mod.rs
bobbinth: There are randomized tests in `inv_proptest()` which should catch issues with other values.

math/src/field/f64/mod.rs
irakliyk: Fixed!

math/src/field/f64/mod.rs
irakliyk: kind of: up until `t31` that's how it works. beyond that, the number is just for the number of bits of the exponent included in the temp value.

math/src/field/f64/mod.rs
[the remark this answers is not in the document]
irakliyk: I was going for something like "exponent_accumulate" :)

math/src/field/f64/mod.rs
irakliyk: There are randomized tests in `inv_proptest()` which should cover all other values.

math/src/field/f64/mod.rs
irakliyk: I think so.

PR #74, opened 2022-02-14
Add sum and fact examples

cpunkzzz:
DO NOT MERGE

Hi @irakliyk, I've been trying to work through the library so I wrote these two simple examples. They are both failing for the same reason: https://pastebin.com/24FrCrM2.

I'm wondering if you have a suggestion on how to debug this issue?

### DIFF
--- a/examples/src/factorial/air.rs
+++ b/examples/src/factorial/air.rs
@@ -13,12 +13,12 @@ use log::debug;
 // FIBONACCI AIR
 // ================================================================================================
 
-pub struct FibAir {
+pub struct FactAir {
     context: AirContext<BaseElement>,
     result: BaseElement,
 }
 
-impl Air for FibAir {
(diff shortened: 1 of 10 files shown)


Nashtare:
I'm not @irakliyk but I'll try giving some hints here. There are several issues in your program. For instance your factorial function always outputs 0, no matter which number you use as input (the loop index starts at 0). Your trace length is not a power of two and hence causes the prover to fail as well.
All these things can be debugged more easily if you compile in debug mode instead of release. This will perform additional checks on the prover side (which are pruned in release mode, hence you just get a random looking failure on the verifier side).

cpunkzzz:
@Nashtare thank you! That solved it. I actually did run it with debug mode and didn't get a meaningful error, when you run it in debug mode, do you see anything more descriptive?

PR #77, opened 2022-03-17
Trace commitment refactoring

irakliyk:
This PR refactors how we handle trace commitments in preparation for implementing RAPs (#54).

Specifically:
* Breaks out building of trace and constraint commitments into separate functions of the `Prover` trait.
* Introduces a `Matrix` struct which is used a backing type for trace and constraint evaluations.
* Combine trace LDE and trace Merkle tree into `TraceCommitment` struct.


Jasleen1 (approved):
Looks good to me -- thank you for working on this!

PR #79, opened 2022-04-03
ExtensionOf trait

irakliyk:
This PR implements `ExtensionOf` trait in the `math` crate. This trait provides a way to multiply an extension field element by a base field element directly (without going through conversion from base field to extension field first). The net result of this is that multiplications of base and extension field elements become much faster. Specifically, for FFT-based polynomial evaluation, when polynomial coefficients are in the extension field, we get about 15% speedup vs. the previous implementation.

### DIFF
--- a/math/src/field/f64/tests.rs
+++ b/math/src/field/f64/tests.rs
@@ -6,7 +6,7 @@
 use super::{
     AsBytes, BaseElement, DeserializationError, FieldElement, Serializable, StarkField, E, M,
 };
-use crate::field::{CubeExtension, QuadExtension};
+use crate::field::{CubeExtension, ExtensionOf, QuadExtension};
 use core::convert::TryFrom;
 use num_bigint::BigUint;
 use proptest::prelude::*;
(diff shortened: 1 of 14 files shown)


math/src/field/f64/tests.rs
```diff
@@ -271,6 +271,16 @@ fn quad_mul() {
...
 }
 
+#[test]
+fn quad_mul_base() {
+    let a = <QuadExtension<BaseElement>>::new(rand_value(), rand_value());
+    let b0 = rand_value();
+    let b = <QuadExtension<BaseElement>>::new(b0, BaseElement::ZERO);
+
+    let expected = a * b;
+    assert_eq!(expected, a.mul_base(b0));
```
dbeaverfb, line 281: Reasonable test.

This is great.

Someday we should chat about more robust multiplication testing.

(a + random1) * (b + random2) is even better (for the cases where you have not found an expected result by hand).

dbeaverfb (approved):
super!

Issue #80, opened 2022-04-04
Custom evaluation frames


maxgillett:
Some AIR constraints cannot be expressed as functions involving just two consecutive rows of trace registers. An example of such a constraint is found in the Cairo whitepaper, where a multi-set permutation check is used to ensure that memory is continuous and read-only. This constraint checks address-value pairs for four different pointer types (`pc`, `dst`, `op0`, `op1`). When these pointers are located in the same trace row, it is not possible to implement the required cumulative product step constraint for nontrivial programs, as memory accesses for more than one pointer type can occur in a single execution step and are not broken down into individual rows. Thus when we specify additional auxiliary columns for *all* of the various address-value pairs (regardless of pointer type, and sorted by address) there will be a mismatch in trace length. While sorted address-value columns could be constructed for each pointer type, and continuity constraints could be modified to ensure that memory accesses restricted to a specific pointer type are read-only, it does not appear possible to ensure that the same memory accesses across different pointer types also remain read-only.  To solve this issue, the Cairo whitepaper introduces "virtual columns" and "subcolumns" (page 49) to allow for larger evaluation frames that effectively provide non-overlapping windows of the execution trace.

With the above in mind, a simple modification of the current frame implementation could allow for larger window sizes, and otherwise seek to keep the interface the same. We could split the current `EvaluationFrame` struct into an `EvaluationFrame` trait and `ConsecutiveEvaluationFrame` struct implementing this trait (similar to how `ExecutionTrace` was split in a previous pull request). The `ConsecutiveEvaluationFrame` (there is probably a better name) will have the same functionality as the current struct (it has a window size of 1), and remain in use for both the two OOD point evaluations and for AIR transition evaluations that only need the current and next row. Users that need a larger window for their transition function (e.g. the current 4 rows and next 4 rows) can create a custom struct satisfying the trait. It might also make sense to allow for the `current` and `next` functions to have differing windows sizes. I think at a minimum we will need to modify the `evaluate_fragment` function in `ConstraintEvaluator` and the `read_main_trace_frame_into` function in `TraceLde`.

4l0n50:
[...]

PR #81, opened 2022-04-12
Custom evaluation frames

maxgillett:
This is an incomplete and very rough outline of the approach that I had in mind for issue #80. I've highlighted code segments below with some questions.

### DIFF
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -226,7 +226,7 @@ pub trait Air: Send + Sync {
     fn evaluate_aux_transition<F, E>(
         &self,
         main_frame: &Self::Frame,
-        aux_frame: &Self::Frame,
+        aux_frame: &Self::AuxFrame,
         periodic_values: &[F],
         aux_rand_elements: &AuxTraceRandElements<E>,
         result: &mut [E],
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -444,7 +444,7 @@ pub trait Air: Send + Sync {
         self.context().options.domain_offset()
     }
 
-    fn eval_frame_window_size(&self) -> usize {
+    fn eval_frame_size(&self) -> usize {
         // TODO
         2
     }
@@ -517,7 +517,11 @@ pub trait Air: Send + Sync {
     {
         let mut t_coefficients = Vec::new();
         for _ in 0..self.trace_info().width() {
-            t_coefficients.push(public_coin.draw_triple()?);
+            let mut values = Vec::new();
+            for _ in 0..self.eval_frame_size() + 1 {
+                values.push(public_coin.draw()?);
+            }
+            t_coefficients.push(values);
         }
 
         // self.ce_blowup_factor() is the same as number of composition columns
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -156,8 +156,8 @@ pub trait Air: Send + Sync {
     /// This could be any type as long as it can be serialized into a sequence of bytes.
     type PublicInputs: Serializable;
 
-    type Frame: EvaluationFrame<Self::BaseField>;
-    type AuxFrame: EvaluationFrame<Self::BaseField>;
+    type Frame<E: FieldElement>: EvaluationFrame<E>;
+    type AuxFrame<E: FieldElement>: EvaluationFrame<E>;
 
     // REQUIRED METHODS
     // --------------------------------------------------------------------------------------------
@@ -187,7 +187,7 @@ pub trait Air: Send + Sync {
     /// (when extension fields are used).
     fn evaluate_transition<E: FieldElement<BaseField = Self::BaseField>>(
         &self,
-        frame: &Self::Frame,
+        frame: &Self::Frame<E>,
         periodic_values: &[E],
         result: &mut [E],
     );
@@ -225,8 +225,8 @@ pub trait Air: Send + Sync {
     #[allow(unused_variables)]
     fn evaluate_aux_transition<F, E>(
         &self,
-        main_frame: &Self::Frame,
-        aux_frame: &Self::AuxFrame,
+        main_frame: &Self::Frame<F>,
+        aux_frame: &Self::AuxFrame<E>,
         periodic_values: &[F],
         aux_rand_elements: &AuxTraceRandElements<E>,
         result: &mut [E],
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -156,6 +156,8 @@ pub trait Air: Send + Sync {
     /// This could be any type as long as it can be serialized into a sequence of bytes.
     type PublicInputs: Serializable;
 
+    type Frame: EvaluationFrame<Self::BaseField>;
+
     // REQUIRED METHODS
     // --------------------------------------------------------------------------------------------
 
@@ -184,7 +186,7 @@ pub trait Air: Send + Sync {
     /// (when extension fields are used).
     fn evaluate_transition<E: FieldElement<BaseField = Self::BaseField>>(
         &self,
-        frame: &EvaluationFrame<E>,
+        frame: &Self::Frame,
         periodic_values: &[E],
         result: &mut [E],
     );
@@ -222,8 +224,8 @@ pub trait Air: Send + Sync {
     #[allow(unused_variables)]
     fn evaluate_aux_transition<F, E>(
         &self,
-        main_frame: &EvaluationFrame<F>,
-        aux_frame: &EvaluationFrame<E>,
+        main_frame: &Self::Frame,
+        aux_frame: &Self::Frame,
         periodic_values: &[F],
         aux_rand_elements: &AuxTraceRandElements<E>,
         result: &mut [E],
@@ -440,6 +442,11 @@ pub trait Air: Send + Sync {
     /// to the execution trace domain.
     fn domain_offset(&self) -> Self::BaseField {
         self.context().options.domain_offset()
+    }
+
+    fn eval_frame_window_size(&self) -> usize {
+        // TODO
+        2
     }
 
     // TRACE SEGMENT RANDOMNESS
--- a/air/src/air/transition/frame.rs
+++ b/air/src/air/transition/frame.rs
@@ -56,8 +56,18 @@ impl<E: FieldElement> EvaluationFrame<E> for DefaultEvaluationFrame<E> {
     // ROW MUTATORS
     // --------------------------------------------------------------------------------------------
 
-    fn read_from<'a, I: StreamingIterator>(&mut self, _columns: I, _step: usize) {
+    fn read_from<'a, I: StreamingIterator>(&mut self, mut columns: I, step: usize) {
         // TODO
+        //loop {
+        //    match columns.next() {
+        //        Some(col) => {
+        //            for (i, row) in self.data.iter_mut().enumerate() {
+        //                row[j] = col[i];
+        //            }
+        //        }
+        //        None => break,
+        //    }
+        //}
     }
 
     // ROW ACCESSORS
--- a/air/src/air/transition/frame.rs
+++ b/air/src/air/transition/frame.rs
@@ -4,15 +4,12 @@
 // LICENSE file in the root directory of this source tree.
 
 use super::{super::Air, FieldElement, Vec};
+use utils::iterators::StreamingIterator;
 
 /// A set of execution trace rows required for evaluation of transition constraints.
 /// It is passed in as one of the parameters into
 /// [Air::evaluate_transition()](crate::Air::evaluate_transition) function.
 pub trait EvaluationFrame<E: FieldElement> {
-    type Chunk<'a>
-    where
-        Self: 'a;
-
     /// Creates an empty frame
     fn new<A: Air<BaseField = E>>(air: &A) -> Self;
 
@@ -20,7 +17,7 @@ pub trait EvaluationFrame<E: FieldElement> {
     fn from_rows(rows: Vec<Vec<E>>) -> Self;
 
     /// Fills the frame using the provided column iterator
-    fn read_from<'a, I: Iterator<Item = Self::Chunk<'a>>>(&'a mut self, columns: I, step: usize);
+    fn read_from<'a, I: StreamingIterator>(&mut self, columns: I, step: usize);
 
     /// Returns the specified row
     fn row<'a>(&'a self, index: usize) -> &'a [E];
@@ -41,11 +38,6 @@ pub struct DefaultEvaluationFrame<E: FieldElement> {
 impl<E: FieldElement> DefaultEvaluationFrame<E> {}
 
 impl<E: FieldElement> EvaluationFrame<E> for DefaultEvaluationFrame<E> {
-    type Chunk<'a>
-    where
-        Self: 'a,
-    = &'a [E];
-
     // CONSTRUCTORS
     // --------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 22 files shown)


air/src/air/transition/frame.rs
```diff
@@ -5,33 +5,116 @@
...
-// EVALUATION FRAME
-// ================================================================================================
 /// A set of execution trace rows required for evaluation of transition constraints.
-///
-/// In the current implementation, an evaluation frame always contains two consecutive rows of the
-/// execution trace. It is passed in as one of the parameters into
+/// It is passed in as one of the parameters into
 /// [Air::evaluate_transition()](crate::Air::evaluate_transition) function.
+pub trait EvaluationFrame<E: FieldElement> {
+    type Chunk<'a>
```
maxgillett, line 12: What is the best way to handle indexing of the `Chunk` associated type?

air/src/air/transition/frame.rs
```diff
@@ -5,33 +5,116 @@
...
+    where
+        Self: 'a;
+
+    fn new(width: usize) -> Self;
```
maxgillett, line 16: Can we move the EvaluationFrame constructors outside of the trait definition?

prover/src/trace/trace_lde.rs
```diff
@@ -69,32 +69,30 @@ impl<E: FieldElement> TraceLde<E> {
...
-        let next_lde_step = (lde_step + self.blowup()) % self.trace_len();
+        // TODO: handle next step
+        let _next_lde_step = (lde_step + self.blowup()) % self.trace_len();
 
         // copy main trace segment values into the frame
-        self.main_segment_lde
-            .read_row_into(lde_step, frame.current_mut());
-        self.main_segment_lde
-            .read_row_into(next_lde_step, frame.next_mut());
+        frame.read_segment_into(lde_step, &self.main_segment_lde.columns);
```
maxgillett, line 82: I think it should be the responsibility of the `EvaluationFrame` to read in the segment, especially if only certain row/column combinations are used. Is there a better way to expose the values in the main segment (and aux segments)?

prover/src/constraints/evaluator.rs
```diff
@@ -148,7 +148,7 @@ impl<'a, A: Air, E: FieldElement<BaseField = A::BaseField>> ConstraintEvaluator<
         fragment: &mut EvaluationTableFragment<E>,
     ) {
         // initialize buffers to hold trace values and evaluation results at each step;
-        let mut main_frame = EvaluationFrame::new(trace.main_trace_width());
+        let mut main_frame = A::Frame::new(trace.main_trace_width());
```
maxgillett, line 151: Can we move the EvaluationFrame constructors outside of the trait definition while still initializing a buffer outside of the loop below?

air/src/air/transition/frame.rs
```diff
@@ -5,33 +5,116 @@
...
+
+    fn new(num_columns: usize) -> Self {
+        // TODO: Move constructor outside of trait definition?
+        assert!(
+            num_columns > 0,
+            "number of columns must be greater than zero"
+        );
+        ChunkedEvaluationFrame {
+            data: vec![],
+            window_size: 0,
```
maxgillett, line 59: For now I am assuming equal window sizes for current and next rows

air/src/air/transition/frame.rs
```diff
@@ -5,33 +5,116 @@
...
+    fn read_segment_into(&mut self, step: usize, segment: &Vec<Vec<E>>);
+    fn current<'a>(&'a self) -> Self::Chunk<'a>;
+    fn next<'a>(&'a self) -> Self::Chunk<'a>;
+}
+
+/// Contains two consecutive (multi)row chunks of the execution trace.
+#[derive(Debug, Clone)]
+pub struct ChunkedEvaluationFrame<E: FieldElement> {
+    data: Vec<Vec<E>>, // row major indexing
+    window_size: usize,
```
maxgillett, line 27: For now I am assuming equal window sizes for current and next rows

prover/src/constraints/evaluator.rs
bobbinth: We could put a method for building "blank" frame into the `Air` trait (so, `Air` trait becomes kind of like a factory for evaluation frames). But I'm not seeing why having a constructor on the frame itself is an issue.

One thing I'd potentially do differently is to change the signature of the constructor to take a reference an instance of an `Air` struct. For example, something like:
```Rust
fn new<A: Air<BaseField = Self::BaseField>>(air: &A) -> Self;
```

prover/src/trace/trace_lde.rs
bobbinth: I agree that it should be the responsibility of the frame to read data from the trace. But this method should be generic enough to accommodate requirements of the prover (when we read frames from the original trace and from trace LDE) and the verifier (when we read frames from the proof).

Maybe taking an iterator over slices (where each slice is a column) could work. For example, something like:
```Rust
fn read_from<'a, I: Iterator<Item = &'a [E]>>(&mut self, columns: I, step: usize)
```  

Though, I think the above will have some issues with lifetimes.

bobbinth:
I haven't reviewed this in detail yet - but a few preliminary thoughts:

* I wonder if we can do away with a concept of "current" and "next" rows entirely. Specific implementations of the evaluation frame could still have them, but as far as the `Air` trait is concerned, it just needs to pass the evaluation frame to the `evaluate_transition()` method. There are a few other places where the "shape" of the frame is important (e.g., for constraint composition), but I wonder if these can be generalized as well.
* It may make sense to have two different types of evaluation frames: for for the main trace and one for the auxiliary trace segments.
* One of the things which is frame dependent is divisor for transition constraints. Right now we use the same divisor for all transition constraints, but we might need to make it so that either a frame or the `Air` trait builds divisor(s) for transition constraints.

irakliyk:
I haven't reviewed this in detail yet - but a few preliminary thoughts:

* I wonder if we can do away with a concept of "current" and "next" rows entirely. Specific implementations of the evaluation frame could still have them, but as far as the `Air` trait is concerned, it just needs to pass the evaluation frame to the `evaluate_transition()` method. There are a few other places where the "shape" of the frame is important (e.g., for constraint composition), but I wonder if these can be generalized as well.
* It may make sense to have two different types of evaluation frames: for for the main trace and one for the auxiliary trace segments.
* One of the things which is frame dependent is divisor for transition constraints. Right now we use the same divisor for all transition constraints, but we might need to make it so that either a frame or the `Air` trait builds divisor(s) for transition constraints.

prover/src/constraints/evaluator.rs
irakliyk: We could put a method for building "blank" frame into the `Air` trait (so, `Air` trait becomes kind of like a factory for evaluation frames). But I'm not seeing why having a constructor on the frame itself is an issue.

One thing I'd potentially do differently is to change the signature of the constructor to take a reference an instance of an `Air` struct. For example, something like:
```Rust
fn new<A: Air<BaseField = Self::BaseField>>(air: &A) -> Self;
```

prover/src/trace/trace_lde.rs
irakliyk: I agree that it should be the responsibility of the frame to read data from the trace. But this method should be generic enough to accommodate requirements of the prover (when we read frames from the original trace and from trace LDE) and the verifier (when we read frames from the proof).

Maybe taking an iterator over slices (where each slice is a column) could work. For example, something like:
```Rust
fn read_from<'a, I: Iterator<Item = &'a [E]>>(&mut self, columns: I, step: usize)
```  

Though, I think the above will have some issues with lifetimes.

prover/src/constraints/evaluator.rs
maxgillett: The motivation was to allow each struct to potentially customize the arguments needed for its constructor. I think your suggestion solves that by giving access to the `AirContext`, which should be all the information the constructor needs.

air/src/air/transition/frame.rs
```diff
@@ -3,80 +3,81 @@
...
 /// [Air::evaluate_transition()](crate::Air::evaluate_transition) function.
+pub trait EvaluationFrame<E: FieldElement> {
+    /// Creates an empty frame
+    fn new<A: Air<BaseField = E>>(air: &A) -> Self;
+
+    /// Creates an frame instantiated from the provided rows
+    fn from_rows(rows: Vec<Vec<E>>) -> Self;
+
+    /// Fills the frame using the provided column iterator
+    fn read_from<'a, I: StreamingIterator>(&mut self, columns: I, step: usize);
```
maxgillett, line 20: Tried this in an attempt to solve the lifetime errors from `fn read_from<'a, I: Iterator<Item = &'a [E]>>(&mut self, columns: I, step: usize)`, but stuck on how to handle the associated type from the iterator.

prover/src/constraints/evaluator.rs
```diff
@@ -234,8 +235,8 @@ impl<'a, A: Air, E: FieldElement<BaseField = A::BaseField>> ConstraintEvaluator<
 
             // evaluate boundary constraints; the results go into remaining slots of the
             // evaluations buffer
-            let main_state = main_frame.current();
-            let aux_state = aux_frame.current();
+            let main_state = main_frame.row(step);
+            let aux_state = aux_frame.row(step);
             self.boundary_constraints.evaluate_all(
                 main_state,
                 aux_state,
```
maxgillett, line 242: Not sure how to handle the different type for auxiliary state (`&[E]` instead of `&[E::BaseField]`) in the `evaluate_all` function.

air/src/air/transition/frame.rs
```diff
@@ -3,80 +3,81 @@
...
 
-// EVALUATION FRAME
-// ================================================================================================
 /// A set of execution trace rows required for evaluation of transition constraints.
-///
-/// In the current implementation, an evaluation frame always contains two consecutive rows of the
-/// execution trace. It is passed in as one of the parameters into
+/// It is passed in as one of the parameters into
 /// [Air::evaluate_transition()](crate::Air::evaluate_transition) function.
+pub trait EvaluationFrame<E: FieldElement> {
```
irakliyk, line 12: I think right now we assume that the frame rows are contiguous. Maybe there is a relatively easy way to relax this condition. For example, we could add the following function to the trait:
```Rust
fn offsets() -> &'static [usize];
```

And then for a default implementation this function could return a reference to the following constant:
```Rust
const FRAME_OFFSETS: [usize; 2] = [0, 1];
```

And custom implementations could return different arrays. For example:
```Rust
const FRAME_OFFSETS: [usize; 2] = [0, 1, 3, 31];
```
Would mean that the frame consists of 4 rows: current, next, 3 rows forward, and 31 rows forward.

These are just initial thoughts. I haven't really considered all pros and cons, and there might be a better way of doing this.

irakliyk (commented):
Thanks again for all the work! This is not a full review - just a few random comments/thoughts.

air/src/air/transition/frame.rs
```diff
@@ -3,80 +3,81 @@
...
+    /// Creates an empty frame
+    fn new<A: Air>(air: &A) -> Self;
+
+    /// Creates an frame instantiated from the provided rows
+    fn from_rows(rows: Vec<Vec<E>>) -> Self;
```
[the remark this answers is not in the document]
irakliyk: We also could add a function `fn to_table(&self) -> Table<E>;` which would convert the frame into a table. This might come in handing when we deal with OOD composition.

air/src/air/transition/frame.rs
irakliyk: btw, this would also provide a default implementation of `row_count()` function. It could look like:
```Rust
fn row_count(&self) -> usize {
    Self::offsets().len()
}
```

A separate nit: would prefer `num_rows()` instead of `row_count()` (that's the convention I've been trying to use throughout the library).

air/src/air/transition/frame.rs
irakliyk: One thought is that maybe instead of passing in an iterator (which needs to be mutable), we pass in something else which provides a view into a 2D structure (which doesn't need to be mutable). For example, let's say we have a `Matrix` trait (a different name is probably needed since we already have `Matrix` struct in the prover) which provides the following interface:

```Rust
pub trait Matrix<E: FieldElement> {

    fn num_cols(&self) -> usize;
    fn num_rows(&self) -> usize;

    fn get(&self, col_idx: usize, row_idx: usize) -> E;
}
```

And then `read_from()` can have the following signature:
```Rust
fn read_from<M: Matrix<E>>(&mut self, data: M, step: usize);
```

And then we'd need to implement `Matrix` trait for w/e struct needs to be passed into `read_from()`.

air/src/air/transition/frame.rs
maxgillett: Good idea. I've also added `FRAME_SHIFT` to indicate how many steps in the trace the frame should skip over at each evaluation. It might be a good idea to indicate not just the rows that need to be read but also columns (I mentioned this in the original issue), but I think this is a good start.

air/src/air/transition/frame.rs
maxgillett: This seems like a cleaner approach. I've called it `TableReader` for now, but am open to better names.  I've also added a `blowup` argument to the signature so that the row indices can be computed for the LDE trace.

air/src/air/transition/frame.rs
irakliyk: Looks good! Two small comments:

* I don't think we can have `FRAME_OFFSETS` as a constant at the trait level as it limits the number of items in the array. Instead, we can make `offsets()` function be a required so that a specific implementation of the trait can adapt it as needed.
* We probably don't need 'shift()' function as `FRAME_SHIFT` constant can be just as easily accessed.

air/src/air/transition/frame.rs
```diff
@@ -3,80 +3,96 @@
...
+    fn new<A: Air>(air: &A) -> Self;
+
+    /// Creates a new frame instantiated from the provided row-major table
+    fn from_table(table: Table<E>) -> Self;
+
+    /// Convert frame to a row-major table
+    fn to_table(&self) -> Table<E>;
+
+    /// Reads selected trace rows from the supplied data into the frame
+    fn read_from<R: TableReader<E>>(&mut self, data: R, step: usize, blowup: usize);
```
4l0n50, line 27: What is `blowup` here? What would be the blowup in the original case of two consecutive rows?

verifier/src/lib.rs
```diff
@@ -169,7 +169,7 @@ pub fn verify<AIR: Air>(
 /// attests to a correct execution of the computation specified by the provided `air`.
 fn perform_verification<A, E, H>(
     air: A,
-    mut channel: VerifierChannel<E, H>,
+    mut channel: VerifierChannel<E, H, A::Frame<E>, A::AuxFrame<E>>,
```
4l0n50, line 172: I think the type parameter of `A::AuxFrame` might be not necessarily equal to the one of `A::Frame`

4l0n50: Nope, my mistake. From the point of view of the verifier both frames are defined over the same field (I was thinking on the prover where they are defined over different fields)

air/src/air/mod.rs
```diff
@@ -156,6 +156,9 @@ pub trait Air: Send + Sync {
     /// This could be any type as long as it can be serialized into a sequence of bytes.
     type PublicInputs: Serializable;
 
+    type Frame<E: FieldElement>: EvaluationFrame<E>;
+    type AuxFrame<E: FieldElement>: EvaluationFrame<E>;
```
4l0n50, line 160: Are you trying to use generic associated types? because I get many errors of the type `error[E0658]: generic associated types are unstable` when trying to implement the `Air` traits.

maxgillett: Yes, you'll need to compile using rust nightly. It sounds like we might wait to merge this in until it is stabilized (according to @irakliyk this should happen within the next couple months). I should note that while the code does compile right now, it's still incomplete and not properly handling row offsets and frame shifts yet. Hoping to finish it up within the next week, and will let you know when it's ready for review.

irakliyk:
Hmm - I didn't realize that deleting the other branch would close the PR. Sorry! Could you rebase this PR from the `main` branch?

maxgillett:
I just rebased from `main`. Are you able to reopen this pull request so I can change the base branch to `main`? You might have to briefly restore the `multi-segment-trace` branch.

PR #83, opened 2022-04-13
Better handling of transition constraint degrees

irakliyk:
This PR improves handling of transition constraint degrees. Specifically, previously the blowup factor required to accommodate transition constraint degrees was computed too conservatively. For example, for transition constraints of degree 5, the required blowup factor would have been computed as 8.

This is not strictly needed because divisor polynomial for transition constraints has degree on the order of trace length (specifically, `trace_length - 1`). Thus, once we take the divisor into account, the blowup factor needed to accommodate degree 5 constraints could be reduced to 4.


pgrinaway:
Quick question about this and the required blowup factor (wasn't sure where else to put it and didn't seem to merit another issue): Since in the code the required blowup factor is calculated as `self.base + self.cycles.len() - 1`, does this mean that for the Rescue example (with a cycle length of 16 and a degree of 3), the minimum blowup factor is the next power of two to 18, which is 32? Thanks in advance, and apologies if this is a silly question.

irakliyk:
@pgrinaway - the `self.cycles.len()` part refers to the number of periodic columns used in a constraint (not the lengths of cycles within them). So, if a constraint uses one periodic column, and has a base degree 3, we'll need a blowup factor of 4.

pgrinaway:
Thanks! Silly misreading on my part.

PR #84, opened 2022-04-14
Custom transition constraint exemptions

irakliyk:
This PR will add ability to specify custom transition constraint exemptions.

Right now, transition constraints are applied to all execution trace rows except for the last one. With this PR, we will be able to specify how many rows (> 1) to "exempt" at the end of the trace. This would be useful for implementing perfect zero knowledge as well as for other purposes.


Jasleen1 (approved):
Looks good (and clever!), thank you!

PR #85, opened 2022-04-19
Include RAPs example with Rescue Chain

Nashtare:
This PR includes an example leveraging RAPs, by decomposing a Rescue chain hash in two parallel executions of half the original length.
It requires changes on the OOD frame handling on the verifier side (when reseeding the public coin) to be workable.


examples/src/rescue_raps/custom_trace_table.rs
```diff
@@ -0,0 +1,496 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#![allow(unused)]
```
irakliyk, line 6: I'd prefer to remove the unused functionality from this module. We may bring it in in the future, but for now, the simpler we can make the test (and the PR) the better. For example, I'm not sure we need to bother with trace fragments for this.

examples/src/rescue/rescue.rs
```diff
@@ -45,6 +45,14 @@ pub fn apply_round(state: &mut [BaseElement], step: usize) {
...
 }
 
+pub fn apply_round_parallel(multi_state: &mut [BaseElement], step: usize) {
+    debug_assert_eq!(multi_state.len() % STATE_WIDTH, 0);
+
+    for state in multi_state.chunks_mut(STATE_WIDTH) {
+        apply_round(state, step)
+    }
+}
+
```
irakliyk, line 55: This function is specific for the `rescue_rap` example - maybe we should move it there?

irakliyk (changes requested):
Huge thank you for working on this example! I left a coupe of minor comments inline, but my main feedback is below.

I wasn't able to come up with a simple way to identify the first and the last row which would not affect performance. We can probably do this by replacing the periodic columns with 3 real trace columns - bit I think this will complicate the example considerably which will make understanding of the main intent quite a bit more difficult.

So, I'm wondering if we could change the example somewhat. We would still have two hash chains, but they would work as follows: the chains would "absorb" a list of values but in different order - and the proof would basically attest that both chains absorb the same values. More specifically:

Let's say we have some values `A`, `B`, `C`, and `D` (these values don't need to be publicly known). Chain 1 would be computed as follows:
```
h1 = hash(0, A)
h2 = hash(h1, B)
h3 = hash(h2, C)
h4 = hash(h3, D)
```

Chain 2 could be computed as follow (but could also be in any other order):
```
g1 = hash(0, B)
g2 = hash(g1, C)
g3 = hash(g2, D)
g4 = hash(g3, A)
```

Both `h4` and `g4` would be public inputs (enforced via boundary constraints). Thus, the proof should convince the verifier that both `h4` and `g4` were derived by hashing the same set of values, but in different order.

Thank you again, and let me know what you think about this example.

examples/src/rescue_raps/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+fn merge(values: &[[BaseElement; 2]; 2], result: &mut [BaseElement]) {
+    let mut state = BaseElement::zeroed_vector(STATE_WIDTH);
+    state[..2].copy_from_slice(&values[0]);
+    state[2..].copy_from_slice(&values[1]);
+    for i in 0..NUM_HASH_ROUNDS {
+        rescue::apply_round(&mut state, i);
+    }
+    result.copy_from_slice(&state[..2]);
+}
```
irakliyk, line 144: I think the naming is a bit confusing here. Usually, when we say `merge` we mean to merge two independent values, in which case, the capacity elements of the state are initialized to zeros.

Probably a more appropriate name would be `absorb`. In such cases we take the previous state and add values to be absorbed into the rate portion of the state. It could look something like this:

```Rust
fn absorb(state: &mut [BaseElement; STATE_WIDTH], values: [BaseElement; 2]) {
    state[0] += values[0];
    state[1] += values[1];
    for i in 0..NUM_HASH_ROUNDS {
        rescue::apply_round(&mut state, i);
    }
}
```

It doesn't make a big difference for example purposes, but if possible - would be great to update it.

examples/src/rescue_raps/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mut permuted_values = [BaseElement::ZERO; 2];
+    let mut tmp = [BaseElement::ZERO; 2];
+    let mut permuted_tmp = [BaseElement::ZERO; 2];
+
+    // Start the hash chain
+    for (seed, pseed) in seeds.iter().zip(permuted_seeds) {
+        merge(&[values, *seed], &mut tmp);
+        merge(&[permuted_values, *pseed], &mut permuted_tmp);
+        values.copy_from_slice(&tmp);
+        permuted_values.copy_from_slice(&permuted_tmp);
```
irakliyk, line 160: If we change to the `absorb` function as described in the other comment, this could be changed to something like:

```Rust
absorb(&mut state, seed);
absorb(&mut permuted_state, speed);
```

Where `state` and `permuted_state` would track the state of the hashchains.

examples/src/rescue_raps/custom_trace_table.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        let mut row = unsafe { uninit_vector(self.width()) };
+        self.read_row_into(0, &mut row);
+        let mut aux_columns = vec![vec![E::ZERO; self.length()]; self.aux_trace_width()];
+
+        aux_columns[0][0] = rand_elements[0] * row[2].into() + rand_elements[1] * row[3].into();
+        aux_columns[1][0] = rand_elements[0] * row[6].into() + rand_elements[1] * row[7].into();
+
+        // Permutation argument column
+        aux_columns[2][0] = E::ONE;
```
irakliyk, line 232: This can probably be a single column - but I think having 3 column makes the example simpler to digest. Maybe add a comment that 3 column are technically not required and are used for illustrative purposes?

examples/src/rescue_raps/custom_trace_table.rs
```diff
@@ -0,0 +1,253 @@
...
+/// For computations which consist of many small independent computations, we can generate the
+/// execution trace of the entire computation by building fragments of the trace in parallel,
+/// and then joining these fragments together.
+///
+/// For this purpose, `RapTraceTable` struct exposes [fragments()](RapTraceTable::fragments)
+/// method, which takes fragment length as a parameter, breaks the execution trace into equally
+/// sized fragments, and returns an iterator over these fragments. You can then use fragment's
+/// [fill()](RapTraceTableFragment::fill) method to fill all fragments with data in parallel.
+/// The semantics of the fragment's [RapTraceTableFragment::fill()] method are identical to the
+/// semantics of the [RapTraceTable::fill()] method.
```
irakliyk, line 51: This section (and some of the above comments) are outdated. We can probably remove most of them and just have a simple description.

irakliyk (changes requested):
Thank you! Pretty much everything looks good. I left a couple of inline comments - the biggest one of which is about how we compute hash chains.

irakliyk:
Hmm - I didn't realize that deleting the other branch would close the PR. Sorry! Could you rebase this PR from the `main` branch?

examples/src/rescue_raps/custom_trace_table.rs
Nashtare: Yes sorry! I've removed some others as well.

examples/src/rescue_raps/custom_trace_table.rs
Nashtare: Yes I had kept them for illustrative purposes, comment added!

examples/src/rescue_raps/mod.rs
Nashtare: Ah yes, sorry for the naming. I changed the example. This requires one extra periodic column, to retrieve the inserted values as we don't reset registers but add them instead, though I believe this is still fine.

PR #76, opened 2022-02-28
update `ExecutionTrace` to latest `TraceTable` struct in README example

bajpai244:
As per the latest v0.3.0, the struct `ExecutionTrace` must be replaced
with `TraceTable` in the example of the README.md file.

The PR updates the example and will make the example consistent with the latest v0.3.0 of Winterfell.

PR #86, opened 2022-04-26
Include RAPs example with Rescue Chain

Nashtare:
Successor of #85.
Includes a RAPs example with Rescue. The example computes two hash chains in parallel, each absorbing a sequence of random elements that are a permutation of each other.


examples/src/rescue_raps/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    let mut state = [BaseElement::ZERO; STATE_WIDTH];
+    let mut permuted_state = [BaseElement::ZERO; STATE_WIDTH];
+
+    // Start the hash chain
+    for (seed, permuted_seed) in seeds.iter().zip(permuted_seeds) {
+        absorb(&mut state, seed);
+        absorb(&mut permuted_state, permuted_seed);
+    }
+
+    [state, permuted_state]
```
irakliyk, line 158: I'm wondering if we should return just the first two elements of each state here. For example, something like:

```Rust
[[state[0], state[1]], [permuted_state[0], permuted_state[1]]]
```

Not hugely important for this example, but in theory, having the entire hasher state be a public input would make the hash chain invertible.

examples/src/rescue_raps/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+        options.to_proof_options(42, 4),
+    ))
+}
+
+pub struct RescueRapsExample {
+    options: ProofOptions,
+    chain_length: usize,
+    seeds: Vec<[BaseElement; 2]>,
+    permuted_seeds: Vec<[BaseElement; 2]>,
+    result: [[BaseElement; STATE_WIDTH]; 2],
```
irakliyk, line 51: Related to the other comment: I wonder if the `result` should be `[[BaseElement; 2]; 2]`. This way, public inputs would include only the rate portion of the state (rather than the entire state).

examples/src/rescue_raps/custom_trace_table.rs
```diff
@@ -0,0 +1,237 @@
...
+use winterfell::{
+    math::{log2, FieldElement, StarkField},
+    EvaluationFrame, Matrix, Trace, TraceInfo, TraceLayout,
+};
+
+// RAP TRACE TABLE
+// ================================================================================================
+/// A concrete implementation of the [Trace] trait supporting custom RAPs.
+///
+/// This implementation supports concurrent trace generation and should be sufficient for most use
```
irakliyk, line 16: We should probably remove a reference to "concurrent trace generation" here.

examples/src/rescue_raps/air.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    fn get_assertions(&self) -> Vec<Assertion<Self::BaseField>> {
+        // Assert ending values of the hash chain
+        let last_step = self.trace_length() - 1;
+        vec![
+            Assertion::single(0, last_step, self.result[0][0]),
+            Assertion::single(1, last_step, self.result[0][1]),
+            Assertion::single(4, last_step, self.result[1][0]),
+            Assertion::single(5, last_step, self.result[1][1]),
+        ]
```
irakliyk, line 222: We should probably enforce that columns 2, 3, and 6, 7 are set to zeros at the first step (at the beginning of the hash chain).

irakliyk (changes requested):
Thank you! (and sorry again for deleting the other branch). I left a few more small comments inline. After these, we should be good to merge.

Nashtare:
Argh sorry, I rushed these last changes before going to vacation, guess I should have been more careful! 😅 
hope everything is alright now! I also updated the READMEs (in examples/ to describe this new one, and in the main one that superseeds #76). I kept the wording `execution trace` in sentences as it still makes sense to describe it as so, only the struct name should be adapted to the new one I think.

examples/src/rescue_raps/air.rs
Nashtare: Right, addressed!

irakliyk (approved):
No worries! And a huge thank you again for working on this.

Everything looks great! I'll merge this PR.

PR #91, opened 2022-04-30
Custom evaluation frame

maxgillett:
Continuation of PR #81 and issue #80


examples/src/custom_frames/fib1/air.rs
```diff
@@ -0,0 +1,72 @@
...
+        &self.context
+    }
+
+    fn evaluate_transition<E: FieldElement + From<Self::BaseField>>(
+        &self,
+        frame: &Self::Frame<E>,
+        _periodic_values: &[E],
+        result: &mut [E],
+    ) {
+        let row_0 = frame.row(0);
```
4l0n50, line 49: What about adding methods `current` and  `next` for the  `DefaultEvaluationFrame`? I think in that way you can keep the examples and simple Air programs almost unchanged.

examples/src/fibonacci/fib2/air.rs
```diff
@@ -41,12 +44,12 @@ impl Air for FibAir {
...
     fn evaluate_transition<E: FieldElement + From<Self::BaseField>>(
         &self,
-        frame: &EvaluationFrame<E>,
+        frame: &Self::Frame<E>,
         _periodic_values: &[E],
         result: &mut [E],
     ) {
-        let current = frame.current();
-        let next = frame.next();
+        let current = frame.row(0);
```
4l0n50, line 51: What about adding methods current and next for DefaultEvaluationFrame? I think in that way you can keep the examples and simple Air programs almost unchanged.

maxgillett: Good point. I'll add these back in.

examples/src/custom_frames/fib1/air.rs
```diff
@@ -0,0 +1,72 @@
...
+    type Frame<E: FieldElement> = FibEvaluationFrame<E>;
+    type AuxFrame<E: FieldElement> = FibEvaluationFrame<E>;
+
+    // CONSTRUCTOR
+    // --------------------------------------------------------------------------------------------
+    fn new(trace_info: TraceInfo, pub_inputs: Self::BaseField, options: ProofOptions) -> Self {
+        let degrees = vec![TransitionConstraintDegree::new(1)];
+        assert_eq!(TRACE_WIDTH, trace_info.width());
+        let context =
+            AirContext::new(trace_info, degrees, 3, options).set_num_transition_exemptions(2);
```
4l0n50, line 32: Maybe the addition of the exemptions shouldn't be done here nor through the Air context. By setting the number of exemptions you're basically modifying the divisor of `TrasitionConstraints` returned by `Air::get_transition_constraints`. Maybe the best that `get_transtition_constraints` is rewritten as follows
```
fn get_transition_constraints<E: FieldElement<BaseField = Self::BaseField>>(
        &self,
        composition_coefficients: &[(E, E)],
    ) -> TransitionConstraints<E> {
         // build constraint divisor; the same divisor applies to all transition constraints
        let divisor = ConstraintDivisor::from_transition(
            self.context().trace_len(),
            2,
        );
        TransitionConstraints::new(self.context(), divisor, composition_coefficients)
    }
```
Note that this also requires a small change on `TransitionConstraints::new` and in the default definition of  `get_transition_constraints`

4l0n50: I forgot to mention that I also proposed to move the computation of the divisor because now it depends on `Self::Frame::FRAME_SHIFT`. The method `ConstraintDivisor::from_transition` should now receive the frame shift instead of the exemptions, and compute the divisor which vanishes over the set `{g^0, g^1,...,g^{(trace_lenght-frame_legth)/(frame_shift)}}` (I guess this should be also a sparse poly).

maxgillett: I realized that a frame shift of 1 is actually sufficient for my purposes right now (implementing the Cairo constraints), so I went ahead and removed it from the PR with the intention of adding it back in at a later time. However, I do need to specify transition-specific divisors to make progress. I believe in my case that the number of exemptions will be a function of the public inputs. Maybe a vector of exemptions can be stored in `AirContext`, and `get_transition_constraints` can then use this to compute the appropriate divisors?

maxgillett: I still need to write an additional test case, but I've pushed my changes if you want to take a look. I've added columns in the `ConstraintEvaluationTable` for each unique `TransitionConstraintGroup` (as determined by the tuple `(constraint_evaluation_degree, divisor_degree)`). As a consequence, there is now an extra column present in this table in the event that all constraint divisors are the same and any aux columns are present, as there is no merging between main and aux constraint groups. There is also a slight performance hit when merging constraints into a `TransitionConstraintGroup` in the case that all divisors are the same, as division of the divisor occurs at each step of the accumulator when folding, and not at the end. Both of these could be remedied, but I don't have a sense of whether the optimization is worth it right now.

air/src/air/transition/mod.rs
```diff
@@ -31,7 +31,7 @@ pub struct TransitionConstraints<E: FieldElement> {
     main_constraint_degrees: Vec<TransitionConstraintDegree>,
     aux_constraints: Vec<TransitionConstraintGroup<E>>,
     aux_constraint_degrees: Vec<TransitionConstraintDegree>,
-    divisor: ConstraintDivisor<E::BaseField>,
+    divisors: Vec<ConstraintDivisor<E::BaseField>>,
```
4l0n50, line 34: You're setting the number of divisors as the number of degrees? (and I think in the end all the divisors would be the same). I was thinking more about a `main_divisor` and `aux_divisor` or at least, many constraints should share the same divisor.  Otherwise it might be heavy to the verifier (don't know how much but I'm worried because is linear in the number of divisors)

examples/src/custom_frames/fib1/air.rs
4l0n50: > I realized that a frame shift of 1 is actually sufficient for my purposes right now (implementing the Cairo constraints),

Without periodic values?

air/src/air/transition/mod.rs
maxgillett: Yes, but they are grouped together according to divisor degree when evaluating, so there isn't unnecessary work being done. You're right that in most cases constraints should share the same divisor, but I have a need to specify different divisors in a small fraction of constraints.

examples/src/custom_frames/fib1/air.rs
4l0n50: and how many rows?

4l0n50: Sorry! I wanted to sat columns! :smile: I was getting mad trying to figure out how could you do it with just two columns (but ~16 rows) and no frame_shifts.

air/src/air/transition/mod.rs
maxgillett: And I'm mistaken again -- it is possible to prove execution with just single `main_divisor` and `aux_divisor` for the Cairo constraints. I'm still wondering if there are use cases that call for a more granular control of the divisors. As I mentioned above I don't think my implementation is adding too much unnecessary work, but an argument could be made for simplifying things if there aren't any use cases for this.

irakliyk:
Closing as stale.

PR #95, opened 2022-06-01
Add check to speed up RP by 50 %

Sword-Smith:
This extra check speeds up Rescue Prime on my DELL XPS 16GB RAM with 50
%.

Executing `cargo criterion hash` gives me:

Before this commit:
hash_rp64_256 (random)  time:   [31.142 us 31.200 us 31.292 us]

After this commit:
hash_rp64_256 (random)  time:   [17.937 us 17.951 us 17.966 us]

My best guess for why this significant speedup happens is that the
multiplication in `add` can sometimes be avoided.

### DIFF
--- a/math/src/field/f64/mod.rs
+++ b/math/src/field/f64/mod.rs
@@ -253,7 +253,16 @@ impl Add for BaseElement {
     #[allow(clippy::suspicious_arithmetic_impl)]
     fn add(self, rhs: Self) -> Self {
         let (result, over) = self.0.overflowing_add(rhs.as_int());
-        Self(result.wrapping_sub(M * (over as u64)))
+        let mut val = result.wrapping_sub(M * (over as u64));
+        // For some reason, this `if` codeblock improves NTT runtime by ~10 % and
+        // Rescue prime calculations with up to 45 % for `hash_pair`. I think it has
+        // something to do with a compiler optimization but I actually don't
+        // understand why this speedup occurs.
+        if val > M - 1 {
+            val -= M;
+        }
+
+        Self(val)
     }
 }
 


irakliyk:
Interesting! On Apple M1 Pro I'm actually seeing this change making Rescue Prime run about 5% slower. I'll need to check on an x86 machine. But if this benefit shows up there, we could probably put a feature flag to add this modification only on x86 architecture.

One question: which version of rustc are you running this with? I'm using 1.61.0.

irakliyk:
By the way, the command I use to run benchmarks is `cargo bench hash` (not `cargo criterion hash`).

Nashtare:
I'm also getting a nice improvement on `hash_rp64_256 (random)`. Down 39% when running `cargo bench hash` and down 43% when running `RUSTFLAGS="-C target-cpu=native" cargo bench hash`. I am on an x86 arch (running on Ubuntu 22.04).

Sword-Smith:
> Interesting! On Apple M1 Pro I'm actually seeing this change making Rescue Prime run about 5% slower. I'll need to check on an x86 machine. But if this benefit shows up there, we could probably put a feature flag to add this modification only on x86 architecture.
> 
> One question: which version of rustc are you running this with? I'm using 1.61.0.

The observed speedup is confirmed on my setup with version `1.61.0` and also by running `cargo bench hash` -- `bench` gives me the same speedup as `criterion`.

irakliyk:
I ran some more benchmarks, and here are my results (all results are for `hash_rp64_256 (random)` benchmark, showing before and after results):

* My original benchmark on M1 Pro (Mac OS): 8.6 us -> 9 us (about 5% degradation)
* x86 Windows (AMD Ryzen 9 5950X): 8.8 us -> 9.7 us (about 10% degradation)
* x86 Ubuntu (AMD EPYC 7R13): 28 us -> 16 us (about 43% improvement)

I'm not quite sure what to make of these numbers. It is obvious that the compiler is not doing its job correctly on AMD EPYC 7R13 Ubuntu because 28 us is way too slow to begin with. This extra bit of code helps it somehow, but it I'm not sure if this is related to the OS or to the machine architecture. It does seem like the issue may be caused by the branch predictor behavior, because even on AMD EPYC 7R13 the "cached" version of the benchmark yields 13 us result.

It is also possible that the benchmark itself is flawed in some way, and maybe in a context of a more complex computation,  the differences would not appear.

irakliyk:
@Sword-Smith - given #101 should we close this PR?

Sword-Smith:
> @Sword-Smith - given #101 should we close this PR?

I think we should. I should compare both on my setup, but let's close. I'll reopen if I find a speedup.

irakliyk:
Sounds good! Thank you!

Sword-Smith:
Related:
https://github.com/rust-lang/rust/issues/106490

PR #97, opened 2022-06-15
Enlarge exemption limit

Nashtare:
This (tiny) PR increases the transition constraint exemption limit in `AirContext` by one.

I believe that in some cases, the current limit may be a problem (for instance a trace where the 2^k rows are necessary, which then needs to be extended to the next power of two 2^{k+1}). In such cases, the current limit won't be sufficient, because the last evaluation frame to be checked on the proving side will overlap between the actual trace and the dummy values.

### DIFF
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -254,11 +254,11 @@ impl<B: StarkField> AirContext<B> {
             n > 0,
             "number of transition exemptions must be greater than zero"
         );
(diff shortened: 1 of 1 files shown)


Nashtare:
(Clippy failing is not due to the changes induced by this PR, but I can make it go green here if needed)

irakliyk:
Thank you! Looks good to me. I'll fix the clippy issue in a separate commit.

irakliyk (approved):

PR #99, opened 2022-07-01
Custom doubling

Nashtare:
This PR overrides the default `double()` method of the `FieldElement` trait for `f64`, `f62`, and related extensions. 
This yielded on my machine a speed-up of respectively 22.1% / 37.9% / 27.5% for `f64` simple/quad/cube and 7.3% / 7.2% / 4.2% for `f62` simple/quad/cube.
I haven't found a significant improvement on `f128` hence I left it untouched. I also added the `double()` operation to the field benchmarks for comparison with the `add` methods.

### DIFF
--- a/math/benches/field.rs
+++ b/math/benches/field.rs
@@ -54,6 +54,11 @@ where
         bench.iter(|| black_box(x) + black_box(y))
     });
 
+    group.bench_function("double", |bench| {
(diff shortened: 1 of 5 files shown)


irakliyk (approved):
Looks good! Thank you so much!

PR #100, opened 2022-07-03
Matrix and FFT optimization

maxgillett:
This PR addresses https://github.com/novifinancial/winterfell/issues/93 with these major changes:
- Introduces a `Matrix` trait in the `winter-math` package along with implementations for row and column-major tables (`RowMajorTable`, `ColumnMajorTable` and `RowMajorRefTable`).
- Adds Matrix-friendly FFT algorithms for polynomial interpolation and evaluation to `winter-math`.
- Replaces the `Matrix` struct in the `winter-prover` package with a `Table` struct that uses `RowMajorTable` to store and retrieve data, and modifies the polynomial methods to use the FFT algorithms that operate on the whole table, not column-by-column.

Preliminary tests show that polynomial interpolation applied on the whole table is up to ~30% faster on an M1 (see `math/benches/matrix.rs` for parameters). Polynomial evaluation remains to be tested, and more extensive benchmarking should be done using [stark-bench](https://github.com/bobbinth/stark-bench).

The following topics need to be discussed:
- All matrix functionality is currently contained in a single `Matrix` trait, which makes it impossible for both row and column-major tables to implement all required methods. For example, a row-major table cannot expose the required contiguous reference in `get_column`. This is currently addressed by using the `unimplemented` macro for problematic methods, and implementing column accessor methods and iterators on the prover's `Table` struct that copy data to return columns. One solution is to break the `Matrix` trait into multiple subtraits that define row and column accessors separately, and by both reference and value.
- In the FFT functions that are constrained to operate on `Matrix`, sub-matrices backed by mutable chunks of the underlying data are sometimes needed, so a `RowMajorRefTable` type is introduced that stores a reference to a subslice of the original data. This table type is currently constructed by calling a `as_ref_table` method on the `Matrix` trait, which is not compatible with a column-major table.  Several solutions are possible:
  - Use a GAT to specify the reference table type in this method.
  - Remove the ColumnMajorTable altogether as it does not appear to be needed, and refactor the `Matrix` trait as proposed above.
- The `Table` struct in `winter-air` should be refactored to use the newly implemented table in the `math` package. This could be added now, or left for a future PR.


irakliyk:
Superseded by #142 and #163.

PR #102, opened 2022-07-05
f64 exp ct

Nashtare:
This PR makes f64 `exp()` method constant-time to match the module description, and add a helper method `exp7()` for Rescue-Prime S-Box.


irakliyk (approved):
Looks great! Thank you!

PR #103, opened 2022-07-13
Relax field requirement of print_trace() utility method

Nashtare:
This tiny PR relaxes the requirement for `f128::BaseElement` in the `print_trace()` utility method so that it can be used for debugging Air programs on arbitrary fields implementing the `StarkField` trait.


irakliyk (approved):
All looks good! Thank you!

PR #104, opened 2022-07-13
feat: MDS mult. using FFT & special matrix

Al-Kindi-0:
This PR implements an optimization, discussed in the setting of the Poseidon hash function [here](https://www.youtube.com/watch?v=HVGaK3xv0Mw), and relies on FFT-based fast matrix-vector multiplication techniques for circulant matrices. See [here](http://web.mit.edu/18.06/www/Spring17/Circulant-Matrices.pdf) for a light exposition of the ideas used.
By choosing an MDS matrix of a very special form, we are able to implement several optimizations on top of the FFT-based multiplication. More precisely, the MDS matrix we chose has components that are small powers of two in "frequency domain" and some of these powers are even equal to zero, provided we scale the (i)FFT appropriately. This translates to multiplications, in frequency domain, being substituted with shifts or even removed entirely. Moreover, the small powers of two entries permit very efficient scheduling of modular reductions. The special matrix we use was found using very optimized code that was developed by the Polygon Zero team. The current implementation benefited, in addition, of the discussions and insights of Hamish Ivey-Law and Jacqueline Nabaglo.
Overall, the resulting improvements are around 20% compared to the original implementation.


irakliyk:
Nice! I'm actually observing a 26% improvement on an M1 Pro: 6.9 us -> 5.1 us!

@Nashtare - curious if you are seeing a similar improvement as well.

Nashtare:
I am getting a 42% (and 60%) improvement on Ubuntu22.04, from 13.171us to 7.645us (cached version) and from 22.302us to 9.8897us (random version). But we've seen weird benchmarking results with Rescue on Linux distribs (and it's overly slow compared to the M1 so I'm not sure to which extent this is relevant).

However I'm not sure to encourage this kind of changes. Unlike Poseidon (that I'm aware of?), Rescue-Prime enjoys a thorough description (https://eprint.iacr.org/2020/1143) including constants generation specification, which I would tend to favor against performance improvements. But I guess this falls back to personal appreciation, so if we judge the improvement to be more important then let it be! :)

irakliyk:
I've tested on a bigger computation (Merkle tree construction), and I'm getting a similar 26% improvement over previous implementation.

In general, I agree with your sentiment, but in this case, we are already deviating from the reference (i.e., using 7 rounds, changing position of rate and capacity) - so, I think a bit more deviation (while does not change the core construction) is OK. Hopefully soon we'll have a reference specifically tailored to this field such that we won't need to deviate at all.

Al-Kindi-0:
> I am getting a 42% (and 56%) improvement on Ubuntu22.04, from 13.171us to 7.645us (cached version) and from 22.302us to 9.8897us (random version). But we've seen weird benchmarking results with Rescue on Linux distribs (and it's overly slow compared to the M1 so I'm not sure to which extent this is relevant).
> 
> However I'm not sure to encourage this kind of changes. Unlike Poseidon (that I'm aware of?), Rescue-Prime enjoys a thorough description (https://eprint.iacr.org/2020/1143) including constants generation specification, which I would tend to favor against performance improvements. But I guess this falls back to personal appreciation, so if we judge the improvement to be more important then let it be! :)

I agree that one should be cautious with updates touching the specs but the current implementation is not different in any way from the original one with regards to the specs. The only difference is with respect to the MDS matrix but the specs in the Rescue prime paper don't put any constraints on it as long as it is MDS. Also, there are no constraints on how one generates it, as opposed to the other constants.

swasilyev:
Just a reminder. The recent version of Poseidon paper warns (https://eprint.iacr.org/2019/458.pdf, p. 6):
> Avoiding Insecure Matrices. 
We emphasize that not every MDS matrix provides the same level of security. In particular...

Al-Kindi-0:
Section 4.2 in https://eprint.iacr.org/2020/1143

irakliyk:
> Just a reminder. The recent version of Poseidon paper warns (https://eprint.iacr.org/2019/458.pdf, p. 6):

Thanks! Based on my understanding of how Rescue Prime works, unlike for Poseidon, any MDS matrix satisfies Rescue Prime requirements.

Al-Kindi-0:
> Section 4.2 in https://eprint.iacr.org/2020/1143

Indeed, that's the reference I mentioned here.

irakliyk (approved):

Al-Kindi-0:
> All looks good! Thank you again!

Thank you!

Nashtare:
@Al-Kindi-0 Out of curiosity, do you have the script that generated this MDS matrix, with those given coefficients? Thanks!

Al-Kindi-0:
Hey @Nashtare !
The above matrix was selected from a list of MDS matrices which are small powers of two in the frequency domain. The selection criteria was the matrix having the fewest and lightest shifts in frequency domain. I defined lightest to be less than 16.
As for the code generating the list, I had to introduce a restriction on the search space so as to make the runtime manageable on my old laptop. This should be easy to remove with a bit more thinking.
I would be happy to share the code that I used to generate the list as well as the transcript that I got. I was actually thinking of making the code faster and more systematic as I believe this would be useful in other settings as well so maybe you can provide some suggestions on that :)

Nashtare:
Thanks @Al-Kindi-0 for the detailed answer! And yes I'd be happy to have access to the code to do some noodling around it.
Actually, I think it would even make sense to refer to it (if it's public) in the related winterfell code.

Al-Kindi-0:
Sounds good! I will make the code public on my Github account once I have cleaned it up a bit and then do a PR to update the documentation with the link included.

PR #105, opened 2022-08-23
Performance improvement in Matrix::interpolate_columns

jimpo:
Instead of cloning the entire matrix with a single .clone(), which will clone each column vector sequentially, do each column clone
inside the parallelized loop. This has a significant performance impact when run with concurrency.

Running the [stark-bench](https://github.com/bobbinth/stark-bench) benchmarks on a 64-core AWS machine (c7g.16xlarge) leads to a ~57% improvement in the interpolation step. UPDATE: The parameters I used to obtain this result were `stark-bench -c 255 -n 23 -b 2`.

### DIFF
--- a/prover/src/matrix.rs
+++ b/prover/src/matrix.rs
@@ -150,10 +150,14 @@ impl<E: FieldElement> Matrix<E> {
     ///   coefficients of a degree `num_rows - 1` polynomial.
     pub fn interpolate_columns(&self) -> Self {
         let inv_twiddles = fft::get_inv_twiddles::<E::BaseField>(self.num_rows());
-        // TODO: get ride of cloning by introducing another version of fft::interpolate_poly()
-        let mut result = self.clone();
-        iter_mut!(result.columns).for_each(|column| fft::interpolate_poly(column, &inv_twiddles));
-        result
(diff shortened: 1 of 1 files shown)


Nashtare:
I'm not getting any performance improvement surprisingly. I'm on Ubuntu22.04, Intel Core i7-9750H (12 cores) running at 2.60GHz.

irakliyk:
@jimpo - very cool! Thank you looking into this!

@Nashtare - I think the improvement will only be noticeable at relatively high core count. The fact that iNTT performance took almost twice as long as I expected on high-core counts was puzzling me for a while. I never noticed this issue locally on an 10-core MBP, but when running on 64 cores, the relative performance between iNTT and NTT always looked odd.

Nashtare:
@irakliyk Good to know, hadn't noticed it before. I tried running the stark-bench with the default parameters provided `./target/release/stark-bench -c 100 -n 23 -b 8` on a 112-cores machine, and got indeed a 20% speed-up in the interpolation step!

jimpo:
Updated note with the parameters I used. Result is more dramatic with more columns b/c the change it to parallelize over the number of columns.

irakliyk (approved):

PR #106, opened 2022-08-29
Merkle path unbatch

Al-Kindi-0:
This PR proposes an implementation of a `into_paths` method for `BatchMerkleProof`.
This method takes a batch proof and outputs the individual Merkle path proofs.

### DIFF
--- a/crypto/src/merkle/proofs.rs
+++ b/crypto/src/merkle/proofs.rs
@@ -255,6 +255,157 @@ impl<H: Hasher> BatchMerkleProof<H> {
         v.remove(&1).ok_or(MerkleTreeError::InvalidProof)
     }
 
+    /// Computes the uncompressed Merkle paths which aggregate to this proof.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// * No indexes were provided (i.e., `indexes` is an empty slice).
+    /// * Number of provided indexes is greater than 255.
+    /// * Number of provided indexes does not match the number of leaf nodes in the proof.
+    pub fn into_paths(self, indexes: &[usize]) -> Result<Vec<Vec<H::Digest>>, MerkleTreeError> {
+        if indexes.is_empty() {
+            return Err(MerkleTreeError::TooFewLeafIndexes);
+        }
+        if indexes.len() > MAX_PATHS {
+            return Err(MerkleTreeError::TooManyLeafIndexes(
+                MAX_PATHS,
+                indexes.len(),
+            ));
+        }
+        if indexes.len() != self.leaves.len() {
+            return Err(MerkleTreeError::InvalidProof);
+        }
+
+        let mut partial_tree_map = BTreeMap::new();
+
+        for (&i, leaf) in indexes.iter().zip(self.leaves.iter()) {
+            partial_tree_map.insert(i + (1 << (self.depth)), *leaf);
+        }
+
+        let mut buf = [H::Digest::default(); 2];
+        let mut v = BTreeMap::new();
+
+        // replace odd indexes, offset, and sort in ascending order
+        let original_indexes = indexes;
+        let index_map = super::map_indexes(indexes, self.depth as usize)?;
+        let indexes = super::normalize_indexes(indexes);
+        if indexes.len() != self.nodes.len() {
+            return Err(MerkleTreeError::InvalidProof);
+        }
+
+        // for each index use values to compute parent nodes
+        let offset = 2usize.pow(self.depth as u32);
+        let mut next_indexes: Vec<usize> = Vec::new();
+        let mut proof_pointers: Vec<usize> = Vec::with_capacity(indexes.len());
+        for (i, index) in indexes.into_iter().enumerate() {
+            // copy values of leaf sibling leaf nodes into the buffer
+            match index_map.get(&index) {
+                Some(&index1) => {
+                    if self.leaves.len() <= index1 {
+                        return Err(MerkleTreeError::InvalidProof);
+                    }
+                    buf[0] = self.leaves[index1];
+                    match index_map.get(&(index + 1)) {
+                        Some(&index2) => {
+                            if self.leaves.len() <= index2 {
+                                return Err(MerkleTreeError::InvalidProof);
+                            }
+                            buf[1] = self.leaves[index2];
+                            proof_pointers.push(0);
+                        }
+                        None => {
+                            if self.nodes[i].is_empty() {
+                                return Err(MerkleTreeError::InvalidProof);
+                            }
+                            buf[1] = self.nodes[i][0];
+                            proof_pointers.push(1);
+                        }
+                    }
+                }
+                None => {
+                    if self.nodes[i].is_empty() {
+                        return Err(MerkleTreeError::InvalidProof);
+                    }
+                    buf[0] = self.nodes[i][0];
+                    match index_map.get(&(index + 1)) {
+                        Some(&index2) => {
+                            if self.leaves.len() <= index2 {
+                                return Err(MerkleTreeError::InvalidProof);
+                            }
+                            buf[1] = self.leaves[index2];
+                        }
+                        None => return Err(MerkleTreeError::InvalidProof),
+                    }
+                    proof_pointers.push(1);
+                }
+            }
+
+            // hash sibling nodes into their parent and add it to partial_tree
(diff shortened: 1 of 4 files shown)


crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,168 @@ impl<H: Hasher> BatchMerkleProof<H> {
...
+        let mut partial_tree_map = BTreeMap::new();
+
+        for i in indexes {
+            if let Some(leaf) = leaves.pop() {
+                partial_tree_map.insert(*i + (1 << (self.depth)), leaf);
+            } else {
+                return Err(MerkleTreeError::TooManyLeafIndexes(
+                    leaves.len(),
+                    indexes.len(),
+                ));
```
irakliyk, line 286: Maybe this is something we can check just once before entering the loop? Also, it might be possible to simplify this loop as follows:
```Rust
for (&i, leaf) in indexes.iter().zip(leaves.iter()) {

}
```
This should also let us avoid cloning and reversing leaves above.

crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,168 @@ impl<H: Hasher> BatchMerkleProof<H> {
...
+            let parent_index = (offset + index) >> 1;
+            v.insert(parent_index, parent);
+            next_indexes.push(parent_index);
+            partial_tree_map.insert(parent_index, parent);
+        }
+
+        // iteratively move up, until we get to the root
+        for _ in 1..self.depth {
+            let indexes = next_indexes.clone();
+            next_indexes.truncate(0);
```
irakliyk, line 361: nit: would prefer `.clear()` vs. `.truncate(0)`.

crypto/src/merkle/proofs.rs
```diff
@@ -343,3 +505,15 @@ impl<H: Hasher> BatchMerkleProof<H> {
 fn are_siblings(left: usize, right: usize) -> bool {
     left & 1 == 0 && right - 1 == left
 }
+
+/// Computes the Merkle path from the computed (partial) tree.
+pub fn get_path<H: Hasher>(index: usize, tree: &BTreeMap<usize, <H as Hasher>::Digest>, depth: usize) -> Vec<H::Digest> {
+    let mut index = index + (1 << depth);
+    let mut proof = vec![*tree.get(&index).unwrap()];
```
irakliyk, line 512: nit: I'd prefer not to use `.unwrap()` (here and two lines below) in non-test code. If we know that `get` will always succeed, we could use `.expect()` instead. Or, if `get` could fail, we should return an error instead.

crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,168 @@ impl<H: Hasher> BatchMerkleProof<H> {
...
+                next_indexes.push(parent_index);
+                partial_tree_map.insert(parent_index, parent);
+
+                i += 1;
+            }
+        }
+        let mut result = vec![];
+        for i in original_indexes {
+            result.push(get_path::<H>(*i, &partial_tree_map, self.depth as usize).to_vec());
+        }
```
irakliyk, line 415: Is the `to_vec()` call at the end needed here? Also, maybe this could be simplified as:
```Rust
let result = original_indexes
    .iter()
    .map(|&i| get_path::<H>(*i, &partial_tree_map, self.depth as usize)).collect::<Vec<_>>();
```

crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,168 @@ impl<H: Hasher> BatchMerkleProof<H> {
...
+                if node_index & 1 != 0 {
+                    buf[0] = sibling;
+                    buf[1] = *node;
+                    partial_tree_map.insert(node_index ^ 1, sibling);
+                } else {
+                    buf[0] = *node;
+                    buf[1] = sibling;
+                    partial_tree_map.insert(node_index ^ 1, sibling);
+                }
+                let parent = H::merge(&buf);
```
irakliyk, line 401: Both branches seem to put the same thing into the partial tree map (`partial_tree_map.insert(node_index ^ 1, sibling)`). If this is correct, you might be able to simplify this as:
```Rust
partial_tree_map.insert(node_index ^ 1, sibling);
let parent = if node_index & 1 != 0 {
    H::merge(&[sibling, *node])
} else {
    H::merge(&[*node, sibling])
};
```

crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,168 @@ impl<H: Hasher> BatchMerkleProof<H> {
...
+                    };
+                    i += 1;
+                } else {
+                    let pointer = proof_pointers[i];
+                    if self.nodes[i].len() <= pointer {
+                        return Err(MerkleTreeError::InvalidProof);
+                    }
+                    sibling = self.nodes[i][pointer];
+                    proof_pointers[i] += 1;
+                }
```
irakliyk, line 383: You might be able to simplify this a little as:
```Rust
let sibling = if i + 1 < indexes.len() && indexes[i + 1] == sibling_index {
    i += 1;
    match v.get(&sibling_index) {
        Some(sibling) => *sibling,
        None => return Err(MerkleTreeError::InvalidProof),
    }
} else {
    let pointer = proof_pointers[i];
    if self.nodes[i].len() <= pointer {
        return Err(MerkleTreeError::InvalidProof);
    }
    proof_pointers[i] += 1;
    self.nodes[i][pointer]
};
```

irakliyk (changes requested):
Looks good! Thank you! I left a few comments inline - mostly to improve code ergonomics.

Also, would be great to fix the rustfmt error.

crypto/src/merkle/proofs.rs
Al-Kindi-0: Indeed!

crypto/src/merkle/proofs.rs
Al-Kindi-0: Is it Ok to add another error variant in the `MerkleTreeError` enum? I think it makes sense to return an error of the kind "partial tree was malformed" as it is the most appropriate in this situation.
Also I forgot but this helper function shouldn't be `pub`.

irakliyk: I'd probably just return `InvalidProof` variant. My assumption is that partial tree would end up being malformed only if the original batch proof was malformed, right?

Al-Kindi-0: Yes, if it is used exactly like we do now inside `into_paths`.

crypto/src/merkle/proofs.rs
Al-Kindi-0: Indeed!

crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,158 @@ impl<H: Hasher> BatchMerkleProof<H> {
         v.remove(&1).ok_or(MerkleTreeError::InvalidProof)
     }
 
+    /// Computes the uncompressed Merkle paths which aggregate to this proof.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// * No indexes were provided (i.e., `indexes` is an empty slice).
+    /// * Number of provided indexes is greater than 255.
```
irakliyk, line 263: We should probably add another error condition here which is when the number of indexes does not match the number of leaf nodes in the proof.

crypto/src/merkle/proofs.rs
```diff
@@ -255,6 +255,158 @@ impl<H: Hasher> BatchMerkleProof<H> {
...
+                i += 1;
+            }
+        }
+
+        let result: Result<Vec<Vec<H::Digest>>, MerkleTreeError> = original_indexes
+            .iter()
+            .map(|&i| get_path::<H>(i, &partial_tree_map, self.depth as usize))
+            .collect();
+
+        result
```
irakliyk, line 407: nit: since we return the result immediately after computing it, we could simplify this as:
```Rust
original_indexes
    .iter()
    .map(|&i| get_path::<H>(i, &partial_tree_map, self.depth as usize))
    .collect()
```

irakliyk (changes requested):
Looks good! Thank you! Just a couple of very minor comments inline.

irakliyk (approved):
All looks good! Thank you!

Al-Kindi-0:
Thank you!

PR #110, opened 2022-09-20
Fri extended to folding factor 2

Al-Kindi-0:
This PR extends the FRI protocol to cover the minimal possible folding factor i.e. 2.


irakliyk (approved):
Looks good! Thank you!

Issue #52, opened 2021-08-18
Add Rescue Prime as a hash function option for prover/verifier


irakliyk:
Rescue Prime hash function has been implemented as a part of #50. However, it is not yet integrated into the prover and verifier.

There is one challenging aspect here: `RP62_248` works only in `f62` field, however, prover and verifier expect the hash function to be generic over all fields. It would be nice if we could do some kind of "type narrowing" - e.g. if there is a mismatch between base field of AIR and hash function, a runtime error would be thrown, otherwise everything will work as expected. But I'm not sure if this is currently possible in Rust.

Nashtare:
Ideally I think we would like to abstract it behind a single `RescuePrime` additional type in the `HashFunction` enum. In `crypto/src/hash/`, we could then have a single `pub struct RescuePrime<B: StarkField>(PhantomData<B>);` defined, in addition to the existing sha3 and blake3 hash functions, which relies internally on the concrete implementation of Rescue for the given field.
This would require implementing a Rescue instantiation (of same conjectured security) for each existing field in `math/src/field` (currently f128 missing) but would then allow for an easier abstraction on the prover / verifier side. However I think we would still need a generic implementation, which without `#![feature(specialization)]` is not allowed. I don't know of a way to deal with this on the stable channel.

irakliyk:
One thing I'm wonder is whether we should make hash function a generic parameter for prover/verifier. This way, users of the library would be able to provide any hash function implementations which comply with `ElementHasher` interface.

I think on the prover side this shouldn't be too difficult to do. But the verifier side might be more challenging (depending on how sophisticated we want to make it).

irakliyk:
A few more thoughts on this.

On the prover side, there are a couple of options:
We could add something like `HashFn` associated type to the `Prover` trait:
```Rust
type HashFn: ElementHasher<BaseField = Self::BaseField>;
```
Another option is to make `Prover` trait generic:
```Rust
trait Foo<H: ElementHasher<BaseField = Self::BaseField>>
```

I probably prefer the former approach as I think it would be more flexible - but open to suggestions.


On the verifier side, we'll need to update the signature of the `verify()` function to look something like this:
```Rust
pub fn verify<AIR: Air, H: ElementHasher<BaseField = AIR::BaseField>>(
    proof: StarkProof,
    pub_inputs: AIR::PublicInputs,
) -> Result<(), VerifierError>
```

[...]

PR #111, opened 2022-10-16
refactor: hash function generic for prover/verifier

0xKanekiKen:
This PR addresses #52.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,7 @@ impl Prover for WorkProver {
     type BaseField = BaseElement;
     type Air = WorkAir;
     type Trace = TraceTable<Self::BaseField>;
+    type HashFn = Blake3_256<Self::BaseField>;
 
     // Our public inputs consist of the first and last value in the execution trace.
     fn get_pub_inputs(&self, trace: &Self::Trace) -> PublicInputs {
@@ -297,7 +298,6 @@ pub fn prove_work() -> (BaseElement, StarkProof) {
         32, // number of queries
         8,  // blowup factor
         0,  // grinding factor
-        HashFunction::Blake3_256,
         FieldExtension::None,
         8,   // FRI folding factor
         128, // FRI max remainder length
@@ -318,7 +318,7 @@ pub fn verify_work(start: BaseElement, result: BaseElement, proof: StarkProof) {
     // The number of steps and options are encoded in the proof itself, so we
     // don't need to pass them explicitly to the verifier.
     let pub_inputs = PublicInputs { start, result };
-    match winterfell::verify::<WorkAir>(proof, pub_inputs) {
+    match winterfell::verify::<WorkAir, Blake3_256<Self::BaseField>>(proof, pub_inputs) {
         Ok(_) => println!("yay! all good!"),
         Err(_) => panic!("something went terribly wrong!"),
     }
--- a/air/src/air/divisor.rs
+++ b/air/src/air/divisor.rs
@@ -255,7 +255,7 @@ mod tests {
     fn constraint_divisor_equivalence() {
         let n = 8_usize;
         let g = BaseElement::get_root_of_unity(n.trailing_zeros());
-        let k = 4 as u32;
+        let k = 4_u32;
         let j = n as u32 / k;
 
         // ----- periodic assertion divisor, no offset --------------------------------------------
@@ -288,7 +288,7 @@ mod tests {
         // ----- periodic assertion divisor, with offset ------------------------------------------
 
         // create a divisor for assertion which repeats every 2 steps starting at step 1
-        let offset = 1u32;
+        let offset = 1_u32;
         let assertion = Assertion::periodic(0, offset as usize, j as usize, BaseElement::ONE);
         let divisor = ConstraintDivisor::from_assertion(&assertion, n);
         assert_eq!(
@@ -318,8 +318,8 @@ mod tests {
         }
 
         // create a divisor for assertion which repeats every 4 steps starting at step 3
-        let offset = 3u32;
-        let k = 2 as u32;
+        let offset = 3_u32;
+        let k = 2_u32;
         let j = n as u32 / k;
         let assertion = Assertion::periodic(0, offset as usize, j as usize, BaseElement::ONE);
         let divisor = ConstraintDivisor::from_assertion(&assertion, n);
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -53,7 +53,7 @@ pub enum FieldExtension {
 ///    `num_queries * log2(blowup_factor) + grinding_factor`.
 ///
 /// Another important parameter in defining STARK security level, which is not a part of [ProofOptions]
-/// is a hash function used in the protocol. The proof soundness of a STARK proof is limited by the
+/// is the hash function used in the protocol. The soundness of a STARK proof is limited by the
 /// collision resistance of the hash function used by the protocol. For example, if a hash function
 /// with 128-bit collision resistance is used, soundness of a STARK proof cannot exceed 128 bits.
 #[derive(Debug, Clone, Eq, PartialEq)]
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -10,30 +10,6 @@ use utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serial
 // TYPES AND INTERFACES
 // ================================================================================================
 
-/// Defines a set of available hash function for STARK protocol.
-///
-/// Choice of a hash function has a direct impact on proof generation time, proof size, and proof
-/// soundness. In general, sounds of the proof is bounded by the collision resistance of the hash
-/// function used by the protocol.
-#[repr(u8)]
-#[derive(Debug, Copy, Clone, Eq, PartialEq)]
-pub enum HashFunction {
-    /// BLAKE3 hash function with 192 bit output.
-    ///
-    /// When this function is used in the STARK protocol, proof security cannot exceed 96 bits.
-    Blake3_192 = 1,
-
-    /// BLAKE3 hash function with 256 bit output.
-    ///
-    /// When this function is used in the STARK protocol, proof security cannot exceed 128 bits.
-    Blake3_256 = 2,
-
-    /// SHA3 hash function with 256 bit output.
-    ///
-    /// When this function is used in the STARK protocol, proof security cannot exceed 128 bits.
-    Sha3_256 = 3,
-}
-
 /// Defines an extension field for the composition polynomial.
 ///
 /// Choice of a field for a composition polynomial may impact proof soundness, and can also have
@@ -63,27 +39,28 @@ pub enum FieldExtension {
 /// These parameters have a direct impact on proof soundness, proof generation time, and proof
 /// size. Specifically:
 ///
-/// 1. Hash function - proof soundness is limited by the collision resistance of the hash function
-///    used by the protocol. For example, if a hash function with 128-bit collision resistance is
-///    used, soundness of a STARK proof cannot exceed 128 bits.
-/// 2. Finite field - proof soundness depends on the size of finite field used by the protocol.
+/// 1. Finite field - proof soundness depends on the size of finite field used by the protocol.
 ///    This means, that for small fields (e.g. smaller than ~128 bits), field extensions must be
 ///    used to achieve adequate security. And even for ~128 bit fields, to achieve security over
 ///    100 bits, a field extension may be required.
-/// 3. Number of queries - higher values increase proof soundness, but also increase proof size.
-/// 4. Blowup factor - higher values increase proof soundness, but also increase proof generation
+/// 2. Number of queries - higher values increase proof soundness, but also increase proof size.
+/// 3. Blowup factor - higher values increase proof soundness, but also increase proof generation
 ///    time and proof size. However, higher blowup factors require fewer queries for the same
 ///    security level. Thus, it is frequently possible to increase blowup factor and at the same
 ///    time decrease the number of queries in such a way that the proofs become smaller.
-/// 5. Grinding factor - higher values increase proof soundness, but also may increase proof
+/// 4. Grinding factor - higher values increase proof soundness, but also may increase proof
 ///    generation time. More precisely, proof soundness is bounded by
 ///    `num_queries * log2(blowup_factor) + grinding_factor`.
+///
+/// Another important parameter in defining STARK security level, which is not a part of [ProofOptions]
+/// is a hash function used in the protocol. The proof soundness of a STARK proof is limited by the
+/// collision resistance of the hash function used by the protocol. For example, if a hash function
+/// with 128-bit collision resistance is used, soundness of a STARK proof cannot exceed 128 bits.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct ProofOptions {
     num_queries: u8,
     blowup_factor: u8,
     grinding_factor: u8,
-    hash_fn: HashFunction,
     field_extension: FieldExtension,
     fri_folding_factor: u8,
     fri_max_remainder_size: u8, // stored as power of 2
@@ -118,7 +95,6 @@ impl ProofOptions {
         num_queries: usize,
         blowup_factor: usize,
         grinding_factor: u32,
-        hash_fn: HashFunction,
         field_extension: FieldExtension,
         fri_folding_factor: usize,
         fri_max_remainder_size: usize,
@@ -146,7 +122,6 @@ impl ProofOptions {
             num_queries: num_queries as u8,
             blowup_factor: blowup_factor as u8,
             grinding_factor: grinding_factor as u8,
-            hash_fn,
             field_extension,
             fri_folding_factor: fri_folding_factor as u8,
             fri_max_remainder_size: fri_max_remainder_size.trailing_zeros() as u8,
@@ -186,16 +161,6 @@ impl ProofOptions {
         self.grinding_factor as u32
     }
 
-    /// Returns a hash functions to be used during STARK proof construction.
-    ///
-    /// Security of a STARK proof is bounded by collision resistance of the hash function used
-    /// during proof construction. For example, if collision resistance of a hash function is
-    /// 128 bits, then soundness of a proof generated using this hash function cannot exceed
-    /// 128 bits.
-    pub fn hash_fn(&self) -> HashFunction {
-        self.hash_fn
-    }
-
     /// Specifies whether composition polynomial should be constructed in an extension field
     /// of STARK protocol.
     ///
@@ -227,7 +192,6 @@ impl Serializable for ProofOptions {
         target.write_u8(self.num_queries);
         target.write_u8(self.blowup_factor);
         target.write_u8(self.grinding_factor);
-        target.write(self.hash_fn);
         target.write(self.field_extension);
         target.write_u8(self.fri_folding_factor);
         target.write_u8(self.fri_max_remainder_size);
@@ -244,7 +208,6 @@ impl Deserializable for ProofOptions {
             source.read_u8()? as usize,
             source.read_u8()? as usize,
             source.read_u8()? as u32,
-            HashFunction::read_from(source)?,
             FieldExtension::read_from(source)?,
             source.read_u8()? as usize,
             2usize.pow(source.read_u8()? as u32),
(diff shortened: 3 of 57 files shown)


air/src/proof/mod.rs
```diff
@@ -94,27 +87,6 @@ impl StarkProof {
...
-            get_conjectured_security(
-                self.context.options(),
-                self.context.num_modulus_bits(),
-                self.lde_domain_size() as u64,
-            )
-        } else {
-            // TODO: implement provable security estimation
-            unimplemented!("proven security estimation has not been implement yet")
-        }
-    }
```
irakliyk, line 116: We should not delete this method - but instead find a way to pass required hash function info into it.

examples/src/main.rs
```diff
@@ -69,7 +69,6 @@ fn main() {
 
     let proof_bytes = proof.to_bytes();
     debug!("Proof size: {:.1} KB", proof_bytes.len() as f64 / 1024f64);
-    debug!("Proof security: {} bits", proof.security_level(true));
```
irakliyk, line 72: We still need to print out security level of the proof somehow.

prover/src/lib.rs
```diff
@@ -161,30 +162,18 @@ pub trait Prover {
         // figure out which version of the generic proof generation procedure to run. this is a sort
         // of static dispatch for selecting two generic parameter: extension field and hash function.
         match self.options().field_extension() {
-            FieldExtension::None => match self.options().hash_fn() {
-                HashFunction::Blake3_256 => self.generate_proof::<Self::BaseField, Blake3_256<Self::BaseField>>(trace),
-                HashFunction::Blake3_192 => self.generate_proof::<Self::BaseField, Blake3_192<Self::BaseField>>(trace),
-                HashFunction::Sha3_256 => self.generate_proof::<Self::BaseField, Sha3_256<Self::BaseField>>(trace),
-            },
+            FieldExtension::None => self.generate_proof::<Self::BaseField, Self::HashFn>(trace),
```
irakliyk, line 165: We probably don't need to pass `Self::HashFn` to `self.generate_proof()` as we can access `Self::HashFn` from within `self.generate_proof()`.

irakliyk (commented):
Thank you! I left a couple of comments inline. But I also wonder if this approach limits flexibility.

For example, it seems like we might be losing the ability to specify hash functions at runtime - though, maybe there are ways around it. How would we define a computation and then specify different hash functions when generating its proofs?

irakliyk:
@0xKanekiKen - one way to add missing info to hash function definition is to add a constant to the `Hasher` trait. Something like:
```Rust
pub trait Hasher {
    const COLLISION_RESISTANCE: u32;
    ...
}
```

0xKanekiKen:
Thanks @irakliyk for the review! I've actioned on the changes you've suggested. Please let me know your thoughts on the latest PR.

prover/src/lib.rs
0xKanekiKen: thanks!

air/src/proof/mod.rs
0xKanekiKen: As per you suggestion, I've defined a const COLLISION_RESISTANCE parameter to the `hasher` trait:

```
pub trait Hasher {
    const COLLISION_RESISTANCE: u32;
    ...
}
```

I've then defined a collision resistance parameter in `StarkProof` struct which we can use to get the security level of the hash function being used in the proof:
```
pub struct StarkProof {
    /// Collision resistance of the hash function being used.
    pub collision_resistance: u32,
}
```

crypto/src/hash/mod.rs
```diff
@@ -23,12 +23,16 @@ pub use rescue::{Rp62_248, Rp64_256};
...
 /// This trait defined hash procedures for the following inputs:
 /// * A sequence of bytes.
+/// * Collision resistance of the hash function.
 /// * Two digests - this is intended for use in Merkle tree constructions.
 /// * A digests and a u64 value - this intended for use in PRNG or PoW contexts.
 pub trait Hasher {
     /// Specifies a digest type returned by this hasher.
     type Digest: Digest;
 
+    /// Collision resistance of the hash function being used.
```
irakliyk, line 33: nit: I would say: "Collision resistance of this hash function measured in bits."

air/src/options.rs
```diff
@@ -63,27 +39,23 @@ pub enum FieldExtension {
 /// These parameters have a direct impact on proof soundness, proof generation time, and proof
 /// size. Specifically:
 ///
-/// 1. Hash function - proof soundness is limited by the collision resistance of the hash function
-///    used by the protocol. For example, if a hash function with 128-bit collision resistance is
-///    used, soundness of a STARK proof cannot exceed 128 bits.
```
irakliyk, line 68: I would probably not remove this as it is an import part of security level considerations. We can move this to the end, and make it clear that hash function is not a part of the options. Something along the lines: "Another important parameter in defining STARK security level, which is not a part of [ProofOptions] is a hash function used in the protocol...".

air/src/proof/mod.rs
```diff
@@ -63,6 +63,8 @@ pub struct StarkProof {
     pub ood_frame: OodFrame,
     /// Low-degree proof for a DEEP composition polynomial.
     pub fri_proof: FriProof,
+    /// Collision resistance of the hash function being used.
+    pub collision_resistance: u32,
```
irakliyk, line 67: I would not encode collision resistance into the proof. Instead, maybe we could pass the hash function as a generic parameter into the `security_level()` function? For example, something like this:
```Rust
pub fn security_level<H: Hasher>(&self, conjectured: bool) -> u32
```

prover/src/lib.rs
```diff
@@ -130,6 +128,9 @@ pub trait Prover {
     /// Execution trace of the computation described by this prover.
     type Trace: Trace<BaseField = Self::BaseField>;
 
+    /// Hash function to be used described by this prover.
```
irakliyk, line 131: nit: there seems to be a typo in this comment.

examples/src/fibonacci/fib2/mod.rs
```diff
@@ -98,10 +98,13 @@ impl Example for FibExample {
     }
 
     fn verify(&self, proof: StarkProof) -> Result<(), VerifierError> {
-        winterfell::verify::<FibAir>(proof, self.result)
+        winterfell::verify::<FibAir, <FibProver as Prover>::HashFn>(proof, self.result)
```
irakliyk, line 101: Could we define `Blake3` as a type at the top of this module and then use it both here and in the `prover.rs`? It could look something like this:
```Rust
type Blake3 = winterfell::crypto::hashers::Blake3_256<BaseElement>;
```

examples/src/lib.rs
```diff
@@ -73,18 +69,11 @@ impl ExampleOptions {
             3 => FieldExtension::Cubic,
             val => panic!("'{}' is not a valid field extension option", val),
         };
-        let hash_fn = match self.hash_fn.as_str() {
-            "blake3_192" => HashFunction::Blake3_192,
-            "blake3_256" => HashFunction::Blake3_256,
-            "sha3_256" => HashFunction::Sha3_256,
-            val => panic!("'{}' is not a valid hash function option", val),
-        };
```
irakliyk, line 81: We shouldn't get rid of these. Instead, we should figure out how we can update examples to support instantiations over multiple hash functions.

While we are trying to figure out how to do this, it might be easier to turn off all examples except for one (e.g., `fib2`). Once we have the `fib2` example working with multiple hash functions, we can re-enable all other examples.

irakliyk (changes requested):
Looks good! Thank you! I left a few more comments inline.

examples/src/lib.rs
0xKanekiKen: I've commented out all the other examples in relevant places and refactored the `fib2` example to support instantiations over multiple hash functions. As per our discussion, I've declared both `FibProver` and `FibExample` generically to accept type `ElementHasher`. Please see [62ca1c8](https://github.com/novifinancial/winterfell/pull/111/commits/62ca1c881c0193401878fbf5f2175a7e8ac3eda9).

examples/src/fibonacci/fib2/mod.rs
0xKanekiKen: We might not need to do this as we are making it accept any generic hash function now.

air/src/proof/mod.rs
0xKanekiKen: I've updated the code as per your suggestions.

air/src/options.rs
0xKanekiKen: Done

examples/src/fibonacci/fib2/prover.rs
```diff
@@ -3,20 +3,28 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use std::marker::PhantomData;
+
+use winterfell::crypto::ElementHasher;
```
irakliyk, line 8: nit: `ElementHasher` should come from `super` (and merged with the line below); `PhantomData` should come from core.

examples/src/fibonacci/fib2/tests.rs
```diff
@@ -3,22 +3,33 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use crate::fibonacci::Blake3_256;
```
irakliyk, line 6: nit: this can probably come from `super` and be merged with the line below.

examples/src/fibonacci/fib2/mod.rs
```diff
@@ -29,21 +30,26 @@ const TRACE_WIDTH: usize = 2;
 // FIBONACCI EXAMPLE
 // ================================================================================================
 
-pub fn get_example(options: ExampleOptions, sequence_length: usize) -> Box<dyn Example> {
-    Box::new(FibExample::new(
-        sequence_length,
-        options.to_proof_options(28, 8),
-    ))
+pub fn get_example(options: &ExampleOptions, sequence_length: usize) -> Box<dyn Example> {
```
irakliyk, line 33: We might need to return a `Result` from here. This would be useful in cases when a given example does not support the requested hash function.

irakliyk (changes requested):
Looks good! Thank you! I left a few small comments inline.

Also, I think it would be good to add an example which works with `Rp64_256` hash functions. Maybe we could have `FibSmall` example which would be instantiated over `f64` field, and then it could use SHA256, BLAKE3, and Rp64_256 hash functions as options.

examples/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub struct ExampleOptions {
 
     /// Hash function used in the protocol
     #[structopt(short = "h", long = "hash_fn", default_value = "blake3_256")]
-    hash_fn: String,
+    pub hash_fn: String,
```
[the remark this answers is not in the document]
0xKanekiKen: I've updated the function to handle the Hash function using an Enum.

examples/src/fibonacci/fib2/mod.rs
0xKanekiKen: Make sense

README.md
```diff
@@ -318,7 +318,7 @@ pub fn verify_work(start: BaseElement, result: BaseElement, proof: StarkProof) {
     // The number of steps and options are encoded in the proof itself, so we
     // don't need to pass them explicitly to the verifier.
     let pub_inputs = PublicInputs { start, result };
-    match winterfell::verify::<WorkAir>(proof, pub_inputs) {
+    match winterfell::verify::<WorkAir, WorkProver::Hash>(proof, pub_inputs) {
```
irakliyk, line 321: Instead of referring to `WorkProver` here, I'd prefer using the hash function directly. Or, if that's too long, maybe create a type alias for it.

air/src/air/divisor.rs
```diff
@@ -255,7 +255,7 @@ mod tests {
     fn constraint_divisor_equivalence() {
         let n = 8_usize;
         let g = BaseElement::get_root_of_unity(n.trailing_zeros());
-        let k = 4 as u32;
+        let k = 4u32;
```
irakliyk, line 258: nit: let's do it as `4_u32` here and also below.

crypto/src/hash/mod.rs
```diff
@@ -23,12 +23,16 @@ pub use rescue::{Rp62_248, Rp64_256};
 ///
 /// This trait defined hash procedures for the following inputs:
 /// * A sequence of bytes.
+/// * Collision resistance of the hash function measured in bits.
```
irakliyk, line 26: This bullet point feels out of place here (it does not describe a hashing procedure). I'll put it into a separate paragraph after the list. Or we can get rid of this altogether.

Also, there is a typo on line 24 (which was there before this PR): "defined" should be "defines".

examples/src/fibonacci/fib2/mod.rs
```diff
@@ -29,21 +31,43 @@ const TRACE_WIDTH: usize = 2;
...
+        HashFunction::Sha3_256 => Ok(Box::new(FibExample::<Sha3_256>::new(
+            sequence_length,
+            options,
+        ))),
+        HashFunction::Rp62_248 => {
+            Err("The hash function won't function in this example".to_string())
+        }
+        HashFunction::Rp64_256 => {
+            Err("The hash function won't function in this example".to_string())
+        }
```
irakliyk, line 58: Instead of listing out all unsupported options, why not do add a catch-all clause (i.e., "_ => ...")?

Also, I'd change the error message to "The specified hash function cannot be used with this example."

examples/src/lib.rs
```diff
@@ -166,3 +195,37 @@ pub enum ExampleType {
...
+
+    /// SHA3 hash function with 256 bit output.
+    ///
+    /// When this function is used in the STARK protocol, proof security cannot exceed 128 bits.
+    Sha3_256,
+
+    /// Rescue Prime hash function with 248 bit output. It only works in `f62` field.
+    ///
+    /// When this function is used in the STARK protocol, proof security cannot exceed 124 bits.
+    Rp62_248,
```
irakliyk, line 225: Let's get rid of this option. I'm not sure it adds much given we have `Rp64_256` supported.

examples/src/lib.rs
```diff
@@ -166,3 +195,37 @@ pub enum ExampleType {
         num_signers: usize,
     },
 }
+
+/// Defines a set of available hash function for STARK protocol.
```
irakliyk, line 199: I would change this to something like: "Defines a set of hash functions available for the provided examples. Some examples may not support all listed hash functions."

examples/src/lib.rs
```diff
@@ -73,28 +77,53 @@ impl ExampleOptions {
...
 
 #[derive(StructOpt, Debug)]
 //#[structopt(about = "available examples")]
 pub enum ExampleType {
+    /// Compute a Fibonacci sequence using trace table with 2 registers in `f64` basefield.
+    FibSmall {
+        /// Length of Fibonacci sequence; must be a power of two
+        #[structopt(short = "n", default_value = "67072")]
+        sequence_length: usize,
+    },
```
irakliyk, line 126: nit: let's move this after `Mulfib8`. Also, let's say just "field" instead of "basefield".

examples/src/main.rs
```diff
@@ -30,37 +30,43 @@ fn main() {
...
         ExampleType::LamportA { num_signatures } => {
-            lamport::aggregate::get_example(options, num_signatures)
+            lamport::aggregate::get_example(&options, num_signatures)
         }
         #[cfg(feature = "std")]
         ExampleType::LamportT { num_signers } => {
-            lamport::threshold::get_example(options, num_signers)
+            lamport::threshold::get_example(&options, num_signers)
         }
     };
```
irakliyk, line 65: Can we put `.expect()` here to avoid doing `unwrap()` on line 69?

examples/src/lib.rs
```diff
@@ -73,28 +77,53 @@ impl ExampleOptions {
...
+            ),
             hash_fn,
-            field_extension,
-            self.folding_factor,
-            256,
         )
     }
+
+    /// Returns security level of the input proof in bits.
+    pub fn get_security_level(&self, proof: &StarkProof) -> usize {
```
irakliyk, line 104: nit: let's change the name to `get_proof_security_level()`.

examples/src/fibonacci/utils.rs
```diff
@@ -17,6 +20,18 @@ pub fn compute_fib_term(n: usize) -> BaseElement {
...
+    let mut t0 = F64::ONE;
+    let mut t1 = F64::ONE;
+
+    for _ in 0..(n - 1) {
+        t1 = t0 + t1;
+        core::mem::swap(&mut t0, &mut t1);
+    }
+
+    t1
+}
```
irakliyk, line 33: Instead of creating a new function, we can probably make `compute_fib_term()` function generic.

irakliyk (changes requested):
Looks good! Thank you! I left a few more comments inline.

examples/src/lib.rs
0xKanekiKen: done

examples/src/main.rs
0xKanekiKen: I've added `.expect()` to all the example types.

examples/src/fibonacci/utils.rs
0xKanekiKen: done

examples/src/fibonacci/utils.rs
```diff
@@ -5,9 +5,9 @@
 
 use winterfell::math::{fields::f128::BaseElement, FieldElement};
 
-pub fn compute_fib_term(n: usize) -> BaseElement {
-    let mut t0 = BaseElement::ONE;
-    let mut t1 = BaseElement::ONE;
+pub fn compute_fib_term<T: FieldElement>(n: usize) -> T {
+    let mut t0 = T::ONE;
+    let mut t1 = T::ONE;
```
irakliyk, line 10: minor nit: we usually use `E` for field elements.

0xKanekiKen: done, thanks

examples/src/main.rs
```diff
@@ -29,38 +29,57 @@ fn main() {
...
+            ExampleType::LamportA { num_signatures } => {
+                lamport::aggregate::get_example(&options, num_signatures)
+                    .expect("The example failed to initialize.")
+            }
+            #[cfg(feature = "std")]
+            ExampleType::LamportT { num_signers } => {
+                lamport::threshold::get_example(&options, num_signers)
+                    .expect("The example failed to initialize.")
+            }
+        };
```
irakliyk, line 78: Is it possible to put a single `.expect()` here (right after `}`) - or does this not work?

examples/src/lib.rs
```diff
@@ -3,8 +3,11 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use fibonacci::{Blake3_192, Blake3_256, Sha3_256};
```
irakliyk, line 6: It feels a bit weird that we import these here from one of the examples. Should we define them here and import them in examples instead? Or does this create issues?

examples/src/vdf/regular/mod.rs
```diff
@@ -3,10 +3,15 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-use crate::{Example, ExampleOptions};
+use crate::{
+    fibonacci::{Blake3_192, Blake3_256, Sha3_256},
```
irakliyk, line 7: Similar to the last comment: it feels odd that we define these in fib examples and them import them from there into other places. We should probably put them into a more central place.

examples/src/lib.rs
0xKanekiKen: Agreed, makes sense. I've moved it to the the `lib.rs`. Please see [fa8e7c4](https://github.com/novifinancial/winterfell/pull/111/commits/fa8e7c408cd9ab80ae3367d8c68b87c9e2231ce1).

examples/src/main.rs
0xKanekiKen: It's working, thanks! Please see [fa8e7c4](https://github.com/novifinancial/winterfell/pull/111/commits/fa8e7c408cd9ab80ae3367d8c68b87c9e2231ce1).

irakliyk (approved):
All looks good! Thank you!

@Al-Kindi-0 @Nashtare - would you mind taking a look as well, if you get a chance? Example files probably don't need to be reviewed (with the exception for maybe `fib_small` example) - but it would be helpful to get another set of eyes on changes in the prover and verifier.

Nashtare:
Sure, I'll have a look later today!

examples/src/lib.rs
```diff
@@ -119,6 +143,12 @@ pub enum ExampleType {
         #[structopt(short = "n", default_value = "1048576")]
         sequence_length: usize,
     },
+    /// Compute a Fibonacci sequence using trace table with 2 registers in `f64` field.
+    FibSmall {
+        /// Length of Fibonacci sequence; must be a power of two
+        #[structopt(short = "n", default_value = "67072")]
```
Nashtare, line 149: That is not a power of two (and hence the default case fails). The closest one is the one below
```suggestion
        #[structopt(short = "n", default_value = "65536")]
```

Nashtare (changes requested):
Looks good to me! Only one tiny change in the instantiation of the new example, the default sequence length is wrong (and hence the example is failing). Otherwise only 2 nitpick comments that you can discard.

Al-Kindi-0:
Same here, looks good to me.

examples/src/lib.rs
0xKanekiKen: Thanks, I've updated it in the latest commit. Please see [80c03a8](https://github.com/novifinancial/winterfell/pull/111/commits/80c03a80aaf7359be794e12aeb93f2d4256b900d). I checked the default case in my local system, and I was able to generate the proof of the default case.

air/src/options.rs
```diff
@@ -63,27 +39,28 @@ pub enum FieldExtension {
...
 ///    time and proof size. However, higher blowup factors require fewer queries for the same
 ///    security level. Thus, it is frequently possible to increase blowup factor and at the same
 ///    time decrease the number of queries in such a way that the proofs become smaller.
-/// 5. Grinding factor - higher values increase proof soundness, but also may increase proof
+/// 4. Grinding factor - higher values increase proof soundness, but also may increase proof
 ///    generation time. More precisely, proof soundness is bounded by
 ///    `num_queries * log2(blowup_factor) + grinding_factor`.
+///
+/// Another important parameter in defining STARK security level, which is not a part of [ProofOptions]
+/// is a hash function used in the protocol. The proof soundness of a STARK proof is limited by the
```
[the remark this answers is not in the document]
0xKanekiKen: done

0xKanekiKen:
Thanks, @irakliyk, @Al-Kindi-0, and @Nashtare, for the review! I've worked on the changes you suggested. If everything is fine, then we can go ahead and merge this PR.

examples/src/lib.rs
Nashtare: Great!

PR #116, opened 2022-10-31
Changed visibility of members for batched merkle proof

Jasleen1:
Modifying the visibility for members of the `BatchMerkleProof` struct. Found that this was needed in another application of the crypto crate.


irakliyk (approved):
Looks good! Thank you!

PR #118, opened 2022-11-11
Eliminate exponentiation from prover constraint evaluation

irakliyk:
This PR eliminates exponentiations from evaluating constraints on the prover side. Depending on a computation, this could result in significant speed-up in the constraint evaluation step. For example, for Lamport signature example, the speed up is about 12% on my machine.


irakliyk:
@Nashtare, @Al-Kindi-0 - would appreciate a review if you have a chance (for some reason, I can't mark you as reviewers).

Al-Kindi-0:
Thanks @irakliyk, very nice improvement! The implementation looks good to me.

Nashtare (approved):
Loos good to me as well!

irakliyk:
@Al-Kindi-0 @Nashtare Thank you!

PR #121, opened 2022-11-29
Correction of docs for f64 field

Al-Kindi-0:
Correction of a mistake in the documentation related to Montgomery representation.


irakliyk (approved):
Thank you! Looks good!

Issue #93, opened 2022-05-24
Optimization of Matrix struct


irakliyk:
Currently, [Matrix](https://github.com/novifinancial/winterfell/blob/main/prover/src/matrix.rs) struct stores trace and polynomial evaluation data in a **column-major** order. This allows us to run LDE for each column independently, but presents challenges during constraint evaluation where we need to process two sequential rows. Thus, we spend some time reading rows from the trace by copying values from each column into an `EvaluationFrame` struct.

A better approach could be to store data in a `Matrix` in a **row-major** order. This has the following advantages:
1. `EvaluationFrame` can just store pointers to rows in the trace - no need to move any data around.
2. Running LDE and FFT over multiple columns in parallel will actually be faster - probably up to 20% faster than doing it column by column as we do now.

irakliyk:
Closed by #124, #142, #163, and #168.

PR #124, opened 2022-12-05
New FftInputs trait

0xKanekiKen:
This PR partly addresses #93. It introduces a `FftInput` trait in the `maths` crate, which defines the shape of the input for the FFT computation. This trait is implemented on `[E]`in this PR.

### DIFF
--- a/math/src/fft/fft_inputs.rs
+++ b/math/src/fft/fft_inputs.rs
@@ -1,5 +1,8 @@
-use super::StarkField;
-use crate::FieldElement;
+use super::{permute_index, FieldElement, StarkField};
+
+// CONSTANTS
+// ================================================================================================
+const MAX_LOOP: usize = 256;
 
 // FFTINPUTS TRAIT
 // ================================================================================================
@@ -28,6 +31,29 @@ pub trait FftInputs<B: StarkField> {
     ///
     /// elem_i = elem_i * offset
     fn shift_by(&mut self, offset: B);
+
+    /// Permutes the elements in this input using the permutation defined by the given
+    /// permutation index. The permutation index is a number between 0 and `self.len() - 1`
+    /// that specifies the permutation to apply to the input. The permutation is applied
+    /// in place, so the input is replaced with the result of the permutation. The permutation
+    /// is applied by swapping elements in the input.
+    fn permute(&mut self) {
+        let n = self.len();
+        for i in 0..n {
+            let j = permute_index(n, i);
+            if j > i {
+                self.swap(i, j);
+            }
+        }
+    }
+
+    /// Applies the FFT to this input. The FFT is applied in place, so the input is
+    /// replaced with the result of the FFT. The `twiddles` parameter specifies the
+    /// twiddle factors to use for the FFT. The `twiddles` parameter must be a slice
+    /// of length `self.len() / 2`.
+    fn fft_in_place(&mut self, twiddles: &[B]) {
+        fft_in_place(self, twiddles, 1, 1, 0);
+    }
 }
 
 /// Implements FftInputs for a slice of field elements.
@@ -76,6 +102,53 @@ where
         let offset = E::from(offset);
         for d in self.iter_mut() {
             *d *= offset;
+        }
+    }
+}
+
+// CORE FFT ALGORITHM
+// ================================================================================================
+
+/// In-place recursive FFT with permuted output.
+///
+/// Adapted from: https://github.com/0xProject/OpenZKP/tree/master/algebra/primefield/src/fft
+pub(super) fn fft_in_place<B, I>(
+    values: &mut I,
+    twiddles: &[B],
+    count: usize,
+    stride: usize,
+    offset: usize,
+) where
+    B: StarkField,
+    I: FftInputs<B> + ?Sized,
+{
+    let size = values.len() / stride;
+    debug_assert!(size.is_power_of_two());
+    debug_assert!(offset < stride);
+    debug_assert_eq!(values.len() % size, 0);
+
+    // Keep recursing until size is 2
+    if size > 2 {
+        if stride == count && count < MAX_LOOP {
+            fft_in_place(values, twiddles, 2 * count, 2 * stride, offset);
+        } else {
+            fft_in_place(values, twiddles, count, 2 * stride, offset);
+            fft_in_place(values, twiddles, count, 2 * stride, offset + stride);
+        }
+    }
+
+    for offset in offset..(offset + count) {
+        I::butterfly(values, offset, stride);
+    }
+
+    let last_offset = offset + size * stride;
+    for (i, offset) in (offset..last_offset)
+        .step_by(2 * stride)
+        .enumerate()
+        .skip(1)
+    {
+        for j in offset..(offset + count) {
+            I::butterfly_twiddle(values, twiddles[i], j, stride);
         }
     }
 }
--- a/math/src/fft/fft_inputs.rs
+++ b/math/src/fft/fft_inputs.rs
@@ -19,7 +19,12 @@ pub trait FftInputs<B: StarkField> {
     /// twiddle factor into a larger NTT.
     fn butterfly_twiddle(&mut self, twiddle: B, offset: usize, stride: usize);
 
-    /// Swaps the element at index i with the element at index j.
+    /// Swaps the element at index i with the element at index j. Specifically:
+    ///
+    /// elem_i <-> elem_j
+    ///
+    /// # Panics
+    /// Panics if i or j are out of bounds.
     fn swap(&mut self, i: usize, j: usize);
 
     /// Multiplies every element in this input by a series of increment. Specifically:
@@ -33,10 +38,15 @@ pub trait FftInputs<B: StarkField> {
     fn shift_by(&mut self, offset: B);
 
     /// Permutes the elements in this input using the permutation defined by the given
-    /// permutation index. The permutation index is a number between 0 and `self.len() - 1`
-    /// that specifies the permutation to apply to the input. The permutation is applied
-    /// in place, so the input is replaced with the result of the permutation. The permutation
-    /// is applied by swapping elements in the input.
+    /// permutation index.
+    ///
+    /// The permutation index is a number between 0 and `self.len() - 1` that specifies the
+    /// permutation to apply to the input. The permutation is applied in place, so the input
+    /// is replaced with the result of the permutation. The permutation is applied by swapping
+    /// elements in the input.
+    ///
+    /// # Panics
+    /// Panics if the permutation index is out of bounds.
     fn permute(&mut self) {
         let n = self.len();
         for i in 0..n {
@@ -47,10 +57,13 @@ pub trait FftInputs<B: StarkField> {
         }
     }
 
-    /// Applies the FFT to this input. The FFT is applied in place, so the input is
-    /// replaced with the result of the FFT. The `twiddles` parameter specifies the
-    /// twiddle factors to use for the FFT. The `twiddles` parameter must be a slice
-    /// of length `self.len() / 2`.
+    /// Applies the FFT to this input.
+    ///
+    /// The FFT is applied in place, so the input is replaced with the result of the FFT. The
+    /// `twiddles` parameter specifies the twiddle factors to use for the FFT.
+    ///
+    /// # Panics
+    /// Panics if length of the `twiddles` parameter is not self.len() / 2.
     fn fft_in_place(&mut self, twiddles: &[B]) {
         fft_in_place(self, twiddles, 1, 1, 0);
     }
@@ -112,13 +125,8 @@ where
 /// In-place recursive FFT with permuted output.
 ///
 /// Adapted from: https://github.com/0xProject/OpenZKP/tree/master/algebra/primefield/src/fft
-pub(super) fn fft_in_place<B, I>(
-    values: &mut I,
-    twiddles: &[B],
-    count: usize,
-    stride: usize,
-    offset: usize,
-) where
+fn fft_in_place<B, I>(values: &mut I, twiddles: &[B], count: usize, stride: usize, offset: usize)
+where
     B: StarkField,
     I: FftInputs<B> + ?Sized,
 {
@@ -137,10 +145,12 @@ pub(super) fn fft_in_place<B, I>(
         }
     }
 
+    // Apply butterfly operations.
     for offset in offset..(offset + count) {
         I::butterfly(values, offset, stride);
     }
 
+    // Apply butterfly operations with twiddle factors.
     let last_offset = offset + size * stride;
     for (i, offset) in (offset..last_offset)
         .step_by(2 * stride)
--- a/math/src/fft/fft_inputs.rs
+++ b/math/src/fft/fft_inputs.rs
@@ -1,0 +1,81 @@
+use super::StarkField;
+use crate::FieldElement;
+
+// FFTINPUTS TRAIT
+// ================================================================================================
+
+/// Defines the interface that must be implemented by the input to fft_in_place method.
+pub trait FftInputs<B: StarkField> {
+    /// Returns the number of elements in this input.
+    fn len(&self) -> usize;
+
+    /// Combines the result of smaller number theoretic transform into a larger NTT.
+    fn butterfly(&mut self, offset: usize, stride: usize);
+
+    /// Combines the result of smaller number theoretic transform multiplied with a
+    /// twiddle factor into a larger NTT.
+    fn butterfly_twiddle(&mut self, twiddle: B, offset: usize, stride: usize);
+
+    /// Swaps the element at index i with the element at index j.
+    fn swap(&mut self, i: usize, j: usize);
+
+    /// Multiplies every element in this input by a series of increment. Specifically:
+    ///
+    /// elem_i = elem_i * offset * increment^i
+    fn shift_by_series(&mut self, offset: B, increment: B);
+
+    /// Multiplies every element in this input by `offset`. Specifically:
+    ///
+    /// elem_i = elem_i * offset
+    fn shift_by(&mut self, offset: B);
+}
+
+/// Implements FftInputs for a slice of field elements.
+impl<B, E> FftInputs<B> for [E]
+where
+    B: StarkField,
+    E: FieldElement<BaseField = B>,
+{
+    fn len(&self) -> usize {
+        self.len()
+    }
+
+    #[inline(always)]
+    fn butterfly(&mut self, offset: usize, stride: usize) {
+        let i = offset;
+        let j = offset + stride;
+        let temp = self[i];
+        self[i] = temp + self[j];
+        self[j] = temp - self[j];
+    }
+
+    #[inline(always)]
+    fn butterfly_twiddle(&mut self, twiddle: B, offset: usize, stride: usize) {
+        let i = offset;
+        let j = offset + stride;
+        let temp = self[i];
+        self[j] = self[j].mul_base(twiddle);
+        self[i] = temp + self[j];
+        self[j] = temp - self[j];
+    }
+
+    fn swap(&mut self, i: usize, j: usize) {
+        self.swap(i, j)
+    }
+
+    fn shift_by_series(&mut self, offset: E::BaseField, increment: E::BaseField) {
+        let mut offset = E::from(offset);
+        let increment = E::from(increment);
+        for d in self.iter_mut() {
+            *d *= offset;
+            offset *= increment;
+        }
+    }
+
+    fn shift_by(&mut self, offset: E::BaseField) {
+        let offset = E::from(offset);
+        for d in self.iter_mut() {
+            *d *= offset;
+        }
+    }
+}
--- a/math/src/fft/mod.rs
+++ b/math/src/fft/mod.rs
@@ -435,8 +435,8 @@ where
         "multiplicative subgroup of size {} does not exist in the specified base field",
         values.len()
     );
-    FftInputs::fft_in_place(values, twiddles);
-    FftInputs::permute(values);
+    values.fft_in_place(twiddles);
+    values.permute();
 }
 
 // TWIDDLES
--- a/math/src/fft/mod.rs
+++ b/math/src/fft/mod.rs
@@ -12,6 +12,7 @@
 //! `n` is the domain size.
 
 use crate::{
+    fft::fft_inputs::FftInputs,
     field::{FieldElement, StarkField},
     utils::{get_power_series, log2},
 };
@@ -434,8 +435,8 @@ where
         "multiplicative subgroup of size {} does not exist in the specified base field",
         values.len()
     );
-    serial::fft_in_place(values, twiddles, 1, 1, 0);
-    serial::permute(values);
+    FftInputs::fft_in_place(values, twiddles);
+    FftInputs::permute(values);
 }
 
 // TWIDDLES
@@ -590,7 +591,7 @@ fn permute<E: FieldElement>(v: &mut [E]) {
         #[cfg(feature = "concurrent")]
         concurrent::permute(v);
     } else {
-        serial::permute(v);
+        FftInputs::permute(v);
     }
 }
 
--- a/math/src/fft/mod.rs
+++ b/math/src/fft/mod.rs
@@ -16,6 +16,7 @@ use crate::{
     utils::{get_power_series, log2},
 };
 
+mod fft_inputs;
 mod serial;
 
 #[cfg(feature = "concurrent")]
--- a/math/src/fft/serial.rs
+++ b/math/src/fft/serial.rs
@@ -3,10 +3,8 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-use crate::{
-    field::{FieldElement, StarkField},
-    utils::log2,
-};
+use super::fft_inputs::FftInputs;
+use crate::{field::StarkField, utils::log2, FieldElement};
 use utils::{collections::Vec, uninit_vector};
 
 // CONSTANTS
@@ -58,7 +56,7 @@ where
             fft_in_place(chunk, twiddles, 1, 1, 0);
         });
 
-    permute(&mut result);
+    permute(result.as_mut_slice());
     result
 }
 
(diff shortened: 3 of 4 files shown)


math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,103 @@
...
+#[cfg(feature = "concurrent")]
+use utils::rayon::{self, slice::{Chunks, ChunksMut as ParChunksMut}};
+
+pub trait FftInputs<B: StarkField, E: FieldElement<BaseField = B>> {
+
+    fn evaluate(chunk: &mut Self, p: &Self, offset: B, factor: B, skip: usize);
+
+    fn interpolate(eval: &mut Self, domain_offset: B, offset: &mut B);
+
+    fn chunks_mut(&mut self, chunk_size: usize) -> ChunksMut<'_, E>;
```
0xKanekiKen, line 14: I was thinking of returning `Self` instead of a field element here- `ChunksMut<'_, Self>`. This way, we would get `chunks` of type `[Ffttrait]` in `evaluate_poly_with_offset`, which we can use in `fft_in_place`. While implementing this trait for `[E]`, the `chunks_mut` return a struct of type `ChunksMut<'_, E>` instead of `ChunksMut<'_, [E]>` due to which I had to introduce additionals generics to handle this.

0xKanekiKen:
This PR has been broken down further into 3 new PRs. The first PR(current one) introduces the `FftInput` trait in `serial.rs` expect for `evaluate_poly_with_offset`.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Combines the result of smaller discrete fourier transforms multiplied with a
+    /// twiddle factor into a larger DFT.
+    fn butterfly_twiddle(&mut self, twiddle: B, offset: usize, stride: usize);
+
+    /// Swaps the element at index i with the element at index j.
+    fn swap(&mut self, i: usize, j: usize);
+
+    /// Helper method used to replace each element in fftinputs with respective `poly`
+    /// item times `init_offset`, where the `init_offset` gets updated by `offset_factor`
+    /// right after each element update.
```
irakliyk, line 24: This will need to be update after the change described above is made, but I'd also add concrete formulas for what this operation does.

math/src/fft/serial.rs
```diff
@@ -67,45 +65,40 @@ where
...
-
-    let domain_offset = E::inv(domain_offset.into());
-    let mut offset = E::inv((evaluations.len() as u64).into());
-    for coeff in evaluations.iter_mut() {
-        *coeff *= offset;
-        offset *= domain_offset;
-    }
+    let domain_offset = B::inv(domain_offset);
+    let mut offset = B::inv((evaluations.len() as u64).into());
+    I::shift_by(evaluations, None, domain_offset, &mut offset);
```
irakliyk, line 91: nit: this can probably be just:
```Rust
evaluations.shift_by(domain_offset, offset);
```

irakliyk (changes requested):
Thank you! Looks good! I left a few comments inline.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,76 @@
...
+    ///
+    /// NOTE: When there's no `poly`, the item in fftinputs are updated with their product
+    /// with `init_offset` and `init_offset` gets updated by `offset_factor` after the update.
+    fn shift_by(&mut self, poly: Option<&Self>, offset_factor: B, init_offset: &mut B);
```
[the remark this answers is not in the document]
0xKanekiKen: > I would probably not have `poly` parameter here and instead have another function which takes this extra parameter as non-optional.

I will introduce this new function in the next PR where we would need to implement `chunk_mut` in `FftInputs` to handle `poly_evaluation_with_offset`.

>Also, for init_offset we should probably just use type B as whether it is mutable or not is an implementation detail.

>Lastly, maybe flip the order of parameters so that init_offset comes before offset_factor.

I've actioned on them in the commit cdacf86.

math/src/fft/serial.rs
0xKanekiKen: yes, thanks!

0xKanekiKen:
Thanks @irakliyk for the review. Benchmark runs show that `f64/fft_interpolate_poly/simple/1048576` has 
dropped by ~2.5%, whereas the rest all seem to be at the same the performance level.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,92 @@
+use super::StarkField;
+use crate::FieldElement;
+
+// FFTINPUTS TRAIT
+// ================================================================================================
+
+/// Defines the shape of the input passed as input for FFT computation.
+pub trait FftInputs<B: StarkField> {
+    /// Returns the length of this fftinputs.
```
irakliyk, line 9: nit: I would change this to: "Returns the number of elements in this input."

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Helper method used to replace each element in fftinputs with its product with
+    /// `init_offset. The mathematical formula is as follows:
+    ///
+    /// $$
+    /// elem_{i} = elem_{i} * init_offset
+    /// $$
+    ///
+    /// where elem is the element of the fftinput trait and i is the index of the element in
+    /// the fftindex.
```
irakliyk, line 43: I would change this to:
```
Multiplies every element in this input by `offset`. Specifically:

elem_{i} = elem_{i} * offset
```
Also, `init_offset` will not render correctly in the formula as underscores are used for subscripts. You can probably drop `$$` and it will be fine.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn shift_by_no_offset(&mut self, init_offset: B);
```
irakliyk, line 44: I would rename this to just `shift_by()`. Also, I don't think we need to call the parameter `init_offset` as there is only one offset. So, probably just `offset` will be fine.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// after the element update. The mathematical formula is as follows:
+    ///
+    /// $$
+    /// elem_{i} = elem_{i} * init_offset_{i}
+    /// init_offset_{i+1} = init_offset_{i} * offset_factor
+    /// $$
+    ///
+    /// where elem is the element of the fftinput trait and i is the index of the element in
+    /// the fftindex.
+    fn shift_by_with_offset(&mut self, init_offset: B, offset_factor: B);
```
irakliyk, line 33: I would update the docs to be more inline with what I wrote in the previous comment. Also, I think the formula is:
```
elem_i = elem_i * init_offset * offset_factor^i
```
Lastly, I wonder if a better signature would be:
```Rust
fn shift_by_series(&mut self, offset: B, increment: B);
```
Or maybe just use letter and explain in the comments similar to how it is done [here](https://github.com/novifinancial/winterfell/blob/main/math/src/utils/mod.rs#L48-L51).

math/src/fft/tests.rs
```diff
@@ -18,42 +18,46 @@ use utils::collections::Vec;
 fn fft_in_place() {
     // degree 3
     let n = 4;
-    let mut p = rand_vector(n);
+    let mut binding = rand_vector(n);
+    let p = binding.as_mut_slice();
```
irakliyk, line 22: why do we need to do it like this? Wouldn't doing `polynom::eval_many(p.as_mut_slice(), &domain);` not work?

irakliyk (changes requested):
Thank you! Looks good! I left a few small comments inline.

math/src/fft/fft_inputs.rs
0xKanekiKen: done

math/src/fft/tests.rs
0xKanekiKen: It will work. I didn't think of it then. Thanks!

0xKanekiKen:
@irakliyk Thanks for the review! I did a benchmark analysis(this PR vs Main) on a Macbook Pro with M1 chip. 

```
f64/fft_evaluate_poly/simple/262144
                        time:   [5.2651 ms 5.2752 ms 5.2851 ms]
                        change: [-2.0300% -1.7274% -1.4336%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
f64/fft_evaluate_poly/simple/524288
                        time:   [11.131 ms 11.144 ms 11.153 ms]
                        change: [-3.4868% -2.3287% -1.2115%] (p = 0.00 < 0.05)
                        Performance has improved.
f64/fft_evaluate_poly/simple/1048576
                        time:   [23.450 ms 23.481 ms 23.495 ms]
                        change: [-3.8105% -2.6461% -1.5983%] (p = 0.00 < 0.05)
                        Performance has improved.
f64/fft_evaluate_poly/with_offset/262144
                        time:   [5.7104 ms 5.7136 ms 5.7150 ms]
                        change: [-3.9474% -3.2772% -2.5663%] (p = 0.00 < 0.05)
                        Performance has improved.
f64/fft_evaluate_poly/with_offset/524288
                        time:   [11.916 ms 11.926 ms 11.935 ms]
                        change: [-1.2804% -0.6625% -0.1137%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
f64/fft_evaluate_poly/with_offset/1048576
                        time:   [24.904 ms 24.910 ms 24.913 ms]
                        change: [-0.7624% -0.5275% -0.3145%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low severe
  1 (10.00%) low mild

f64_quad/fft_evaluate_poly/simple/262144
                        time:   [8.1970 ms 8.2128 ms 8.2233 ms]
                        change: [-0.2712% +0.1428% +0.5941%] (p = 0.54 > 0.05)
                        No change in performance detected.
f64_quad/fft_evaluate_poly/simple/524288
                        time:   [17.144 ms 17.158 ms 17.166 ms]
                        change: [-2.9753% -1.6194% -0.5066%] (p = 0.01 < 0.05)
                        Change within noise threshold.
f64_quad/fft_evaluate_poly/simple/1048576
                        time:   [36.466 ms 36.507 ms 36.524 ms]
                        change: [-2.7131% -1.8831% -1.0936%] (p = 0.00 < 0.05)
                        Performance has improved.
f64_quad/fft_evaluate_poly/with_offset/262144
                        time:   [8.1432 ms 8.1475 ms 8.1499 ms]
                        change: [-2.9101% -2.1345% -1.4438%] (p = 0.00 < 0.05)
                        Performance has improved.
f64_quad/fft_evaluate_poly/with_offset/524288
                        time:   [17.079 ms 17.095 ms 17.106 ms]
                        change: [-0.1825% +0.0339% +0.2405%] (p = 0.77 > 0.05)
                        No change in performance detected.
f64_quad/fft_evaluate_poly/with_offset/1048576
                        time:   [36.127 ms 36.137 ms 36.143 ms]
                        change: [-0.5137% -0.2235% +0.0047%] (p = 0.13 > 0.05)
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low severe

f64_cube/fft_evaluate_poly/simple/262144
                        time:   [11.248 ms 11.266 ms 11.278 ms]
                        change: [-4.0939% -3.0034% -1.7962%] (p = 0.00 < 0.05)
                        Performance has improved.
f64_cube/fft_evaluate_poly/simple/524288
                        time:   [23.778 ms 23.803 ms 23.820 ms]
                        change: [-3.1312% -1.9150% -0.6785%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild
f64_cube/fft_evaluate_poly/simple/1048576
                        time:   [49.413 ms 49.537 ms 49.694 ms]
                        change: [+0.5457% +1.1376% +1.7086%] (p = 0.00 < 0.05)
                        Change within noise threshold.
f64_cube/fft_evaluate_poly/with_offset/262144
                        time:   [11.423 ms 11.708 ms 11.875 ms]
                        change: [+1.5217% +3.2481% +5.2445%] (p = 0.00 < 0.05)
                        Performance has regressed.
f64_cube/fft_evaluate_poly/with_offset/524288
                        time:   [23.256 ms 23.310 ms 23.406 ms]
                        change: [-0.0482% +0.6838% +1.2725%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 3 outliers among 10 measurements (30.00%)
  1 (10.00%) low severe
  1 (10.00%) low mild
  1 (10.00%) high severe
f64_cube/fft_evaluate_poly/with_offset/1048576
                        time:   [48.939 ms 49.095 ms 49.275 ms]
                        change: [-2.6747% -1.5169% -0.2587%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild

f64/fft_interpolate_poly/simple/262144
                        time:   [5.4510 ms 5.4724 ms 5.5139 ms]
                        change: [-0.3636% +0.3893% +1.3029%] (p = 0.41 > 0.05)
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
f64/fft_interpolate_poly/simple/524288
                        time:   [11.786 ms 11.981 ms 12.091 ms]
                        change: [+0.1043% +1.6921% +3.4173%] (p = 0.06 > 0.05)
                        No change in performance detected.
f64/fft_interpolate_poly/simple/1048576
                        time:   [24.381 ms 24.553 ms 24.719 ms]
                        change: [+0.8572% +1.5401% +2.1405%] (p = 0.00 < 0.05)
                        Change within noise threshold.
f64/fft_interpolate_poly/with_offset/262144
                        time:   [6.3199 ms 6.4513 ms 6.6823 ms]
                        change: [+0.8724% +2.5547% +5.1051%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
f64/fft_interpolate_poly/with_offset/524288
                        time:   [13.217 ms 13.351 ms 13.526 ms]
                        change: [+1.0659% +1.8573% +2.8208%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
f64/fft_interpolate_poly/with_offset/1048576
                        time:   [27.683 ms 27.927 ms 28.246 ms]
                        change: [+2.4457% +3.8879% +5.3281%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

fft_get_twiddles/262144 time:   [1.8241 ms 1.8379 ms 1.8446 ms]
                        change: [-0.0129% +0.5954% +1.2336%] (p = 0.09 > 0.05)
                        No change in performance detected.
fft_get_twiddles/524288 time:   [3.7273 ms 3.7544 ms 3.7771 ms]
                        change: [+0.4761% +1.7563% +3.0816%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
fft_get_twiddles/1048576
                        time:   [7.4698 ms 7.5297 ms 7.5925 ms]
                        change: [+1.1288% +1.8560% +2.7234%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

math/src/fft/serial.rs
```diff
@@ -18,10 +16,10 @@ const MAX_LOOP: usize = 256;
 
 /// Evaluates polynomial `p` in-place over the domain of length `p.len()` in the field specified
 /// by `B` using the FFT algorithm.
-pub fn evaluate_poly<B, E>(p: &mut [E], twiddles: &[B])
+pub fn evaluate_poly<B, I>(p: &mut I, twiddles: &[B])
```
irakliyk, line 19: Since we can't apply this change to all public functions (i.e., specifically we can't make it work for `evaluate_poly_with_offset()`), we should roll back this and similar changes.

The only two functions which we should change signatures for here are: `permute()` and `fft_in_place()`.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,85 @@
+use super::StarkField;
+use crate::FieldElement;
+
+// FFTINPUTS TRAIT
+// ================================================================================================
+
+/// Defines the interface for a type that can be used as input to an FFT. It is used to abstract away
+/// the details of how the FFT is implemented, so that the same code can be used for both serial and
+/// parallel FFTs. This trait is not intended to be implemented by external types. Instead, it is
+/// used to define the interface that the FFT implementations expect.
```
irakliyk, line 10: I think the comment here needs to be updated.

irakliyk (changes requested):
Thank you! I left a couple more comments inline.

math/src/fft/serial.rs
0xKanekiKen: Understood, I've rolled back the changes in `serial.rs`. https://github.com/facebook/winterfell/pull/124/commits/da95291284bebe52e7713cd5666b82275b86b173 only introduces the `FftInput` trait in methods `permute` and `fft_in_place`. I was wondering if I should also introduce it similarly in `concurrent.rs` in this PR.

math/src/fft/fft_inputs.rs
0xKanekiKen: Let me know if the below is fine?

`Defines the interface that must be implemented by the input to fft_in_place method.`

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,55 @@
+use super::StarkField;
+use crate::FieldElement;
+
+// FFTINPUTS TRAIT
+// ================================================================================================
+
+/// Defines the interface that must be implemented by the input to fft_in_place method.
```
irakliyk, line 7: The description looks good - but I would keep the `shift_by()` methods we had before. I think they simplifies code it on there places.

math/src/fft/tests.rs
```diff
@@ -20,40 +20,40 @@ fn fft_in_place() {
     let n = 4;
     let mut p = rand_vector(n);
     let domain = build_domain(n);
-    let expected = polynom::eval_many(&p, &domain);
+    let expected = polynom::eval_many(p.as_mut_slice(), &domain);
     let twiddles = super::get_twiddles::<BaseElement>(n);
-    super::serial::fft_in_place(&mut p, &twiddles, 1, 1, 0);
-    super::permute(&mut p);
+    super::serial::fft_in_place(p.as_mut_slice(), &twiddles, 1, 1, 0);
+    super::permute(p.as_mut_slice());
```
irakliyk, line 26: is there a reason not to revert these changes?

0xKanekiKen: We have to  pass a slice as an argument in `fft_in_place`. Rest all I think can be reverted back.

math/src/fft/fft_inputs.rs
0xKanekiKen: Then we can use these methods in `interpolate_poly` and `interpolate_poly_with_offset` as follows:

```
pub fn interpolate_poly<B, E>(evaluations: &mut [E], inv_twiddles: &[B])
where
    B: StarkField,
{
    fft_in_place(evaluations, inv_twiddles, 1, 1, 0);
    let inv_length = B::inv((evaluations.len() as u64).into());

    // Use fftinputs shift_by on evaluations.
    FftInputs::shift_by(evaluations, inv_length);
    permute(evaluations);
}

pub fn interpolate_poly_with_offset<B, E>(
    evaluations: &mut [E],
    inv_twiddles: &[B],
    domain_offset: B,
) where
    B: StarkField,
{
    fft_in_place(evaluations, inv_twiddles, 1, 1, 0);
    permute(evaluations);

    let domain_offset = B::inv(domain_offset);
    let offset = B::inv((evaluations.len() as u64).into());

    // Use fftinputs's shift_by_series on evaluations.
    FftInputs::shift_by_series(evaluations, offset, domain_offset);
}

```

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,81 @@
...
+pub trait FftInputs<B: StarkField> {
```
irakliyk, line 8: Looks good! One thing I wonder if we should put `permute()` as a function on this trait. It would be an implemented  function here. Something like:
```Rust
fn permute(&mut self) {
    let n = self.len();
    for i in 0..n {
        let j = permute_index(n, i);
        if j > i {
            self.swap(i, j);
        }
    }
}
```

irakliyk: Or, maybe we could put `fft_in_place()` here as well. Something like:
```Rust
fn fft_in_place(&mut self, twiddles: &[B]) {
    fft_in_place(self, twiddles, 1, 1, 0);
}
```
And then `fft_in_place()` function could be a helper function in this module.

0xKanekiKen: We have to also move the concurrent part of these method in `FftInput` crate. We can do it in the next PR.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+// CORE FFT ALGORITHM
+// ================================================================================================
+
+/// In-place recursive FFT with permuted output.
+///
+/// Adapted from: https://github.com/0xProject/OpenZKP/tree/master/algebra/primefield/src/fft
+pub(super) fn fft_in_place<B, I>(
```
irakliyk, line 115: We probably don't need to make this public any more.

math/src/fft/mod.rs
```diff
@@ -433,8 +435,8 @@ where
         "multiplicative subgroup of size {} does not exist in the specified base field",
         values.len()
     );
-    serial::fft_in_place(values, twiddles, 1, 1, 0);
-    serial::permute(values);
+    FftInputs::fft_in_place(values, twiddles);
+    FftInputs::permute(values);
```
irakliyk, line 439: nit: I'd prefer to use the method notation (here and in other places):
```Rust
values.fft_in_place(twiddles);
values.permute();
```

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,154 @@
...
+            if j > i {
+                self.swap(i, j);
+            }
+        }
+    }
+
+    /// Applies the FFT to this input. The FFT is applied in place, so the input is
+    /// replaced with the result of the FFT. The `twiddles` parameter specifies the
+    /// twiddle factors to use for the FFT. The `twiddles` parameter must be a slice
+    /// of length `self.len() / 2`.
```
irakliyk, line 53: nit: rustdoc does not render monolithic paragraphs like this very well. It is better to have a short intro sentence (1 or  2 lines at most) and then more detailed description. For example:

```
Applies the FFT to this input.

The FFT is applied in place, so the input is  replaced with the result of the FFT. The
`twiddles` parameter specifies the twiddle factors to use for the FFT.

# Panics
Panics if length of the `twiddles` parameter is not self.len() / 2.
```

irakliyk (changes requested):
Looks good! I left a couple more simple comments inline.

math/src/fft/fft_inputs.rs
0xKanekiKen: I was talking about these two methods in `concurrent.rs`.

- [permute](https://github.com/0xKanekiKen/winterfell/blob/add-fftinput-trait/math/src/fft/concurrent.rs#L95)
- [split_radix_fft](https://github.com/0xKanekiKen/winterfell/blob/add-fftinput-trait/math/src/fft/concurrent.rs#L123)

math/src/fft/tests.rs
```diff
@@ -22,8 +23,8 @@ fn fft_in_place() {
     let domain = build_domain(n);
     let expected = polynom::eval_many(&p, &domain);
     let twiddles = super::get_twiddles::<BaseElement>(n);
-    super::serial::fft_in_place(&mut p, &twiddles, 1, 1, 0);
-    super::permute(&mut p);
+    p.as_mut_slice().fft_in_place(&twiddles);
+    p.as_mut_slice().permute();
```
irakliyk, line 27: Do we actually need to do `as_mut_slice()` here? Seems like it works fine on line 67 below without it.

math/src/fft/serial.rs
```diff
@@ -55,10 +49,10 @@ where
                 *d = (*c).mul_base(factor);
                 factor *= offset;
             }
-            fft_in_place(chunk, twiddles, 1, 1, 0);
+            chunk.fft_in_place(twiddles);
         });
 
-    permute(&mut result);
+    result.as_mut_slice().permute();
```
irakliyk, line 55: Do we need `as_mut_slice()` here?

math/src/fft/tests.rs
0xKanekiKen: yes- we don't need to do `as_mut_slice()`. I missed this here. Thanks.

math/src/fft/fft_inputs.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn shift_by(&mut self, offset: B);
+
+    /// Permutes the elements in this input using the permutation defined by the given
+    /// permutation index.
+    ///
+    /// The permutation index is a number between 0 and `self.len() - 1` that specifies the
+    /// permutation to apply to the input.
+    ///
+    /// The permutation is applied in place, so the input is replaced with the result of the
+    /// permutation. The permutation is applied by swapping elements in the input.
```
irakliyk, line 47: minor nit: I'd combine these paragraphs. It is only the first paragraph that needs to be short.

math/src/fft/fft_inputs.rs
irakliyk: Yep - makes sense!

irakliyk (changes requested):
Looks good! Thank you! Just a couple of minor nits inline.

math/src/fft/serial.rs
0xKanekiKen: nope, removed it. Thanks.

irakliyk (approved):
All looks good! Thank you!

PR #127, opened 2022-12-12
Simpler Montgomery reduction for u64

Al-Kindi-0:
This PR proposes a simpler method to convert a field element in Montgomery form to canonical form. The current method uses Montgomery reduction where the input is expected to be u128 (i.e. the result of a multiplication). Since we know that field elements are stored as u64, we can use a simpler method to compute $\frac{x}{2^{64}}\textit{ mod }p$.


irakliyk (approved):
Cool improvement! Thank you!

Issue #126, opened 2022-12-08
More efficient remainder verification in FRI


Al-Kindi-0:
Currently, the last step in FRI verification is  https://github.com/novifinancial/winterfell/blob/e43b75d365e03a71288ede25a6c2c9d9dc021a4d/fri/src/verifier/mod.rs#L307-L319

This is inefficient in, at least, two ways:

1. The commitment to the remainder is a Merkle tree. This is inefficient because we don't really use the property of vector-commitment that a Merkle tree provides in the above implementation and we can in fact get away with the weaker notion of set-commitment, provided the set is composed of elements of the form `(position,remainder[position])`.
2. Combined with the previous point, the verifier can, instead of checking that `remainder[position] == evaluation`, just compute the set commitment to `(position,evaluations[position])` and then check that the two set-commitments are equal. This assumes that the queries at the initial layer are drawn in such a way so as the last folded positions span the range `[0, remainder.len() - 1]` . I assume this is already implemented in Winterfell but I haven't checked.

As for the set-commitment, a sequential hash of `remainder[i]` for $i$ in increasing order might be sufficient.

irakliyk:
I think the first part will definitely work. The second part will also kind of work, but I doubt savings would be more than about 500 bytes in most cases.

The reason for the second part is that we still need to send most of the remainder. We can skip only positions which we get from folding. For example, if the remainder size is 128 elements, and and the proof contains 30 queries, then the best we could hope for is that 30 of 128 elements can be skipped. If we are in a 128-bit field, these 30 skipped elements would result in proof size reduction of 30 * 16 = 480 bytes.

irakliyk:
The first point of this was addressed in #128

irakliyk:
There is actually another optimization we can do which makes the second point here irrelevant. Specifically, instead of sending the actual codeword for the remainder (which we then need to interpolate into a polynomial in coefficient form), we can just send the polynomial in coefficient form. Then, the verifier would evaluate the polynomial and make sure the resulting evaluations match the commitment.

Depending on the blowup factor, this could result in signifiant savings (i.e.,g 1KB or more).

Al-Kindi-0:
[...]

PR #128, opened 2022-12-14
More efficient remainder commitment

Al-Kindi-0:
This PR proposes a solution to point 1 in #126 .

### DIFF
--- a/fri/src/prover/mod.rs
+++ b/fri/src/prover/mod.rs
@@ -94,6 +94,7 @@ where
 {
     options: FriOptions,
     layers: Vec<FriLayer<B, E, H>>,
+    remainder: FriRemainder<E>,
     _channel: PhantomData<C>,
 }
 
@@ -102,6 +103,8 @@ struct FriLayer<B: StarkField, E: FieldElement<BaseField = B>, H: Hasher> {
     evaluations: Vec<E>,
     _base_field: PhantomData<B>,
 }
+
+struct FriRemainder<E: FieldElement>(Vec<E>);
 
 // PROVER IMPLEMENTATION
 // ================================================================================================
@@ -120,6 +123,7 @@ where
         FriProver {
             options,
             layers: Vec::new(),
+            remainder: FriRemainder(vec![]),
             _channel: PhantomData,
         }
     }
@@ -146,6 +150,7 @@ where
     /// Clears a vector of internally stored layers.
     pub fn reset(&mut self) {
         self.layers.clear();
+        self.remainder.0.clear();
     }
 
     // COMMIT PHASE
@@ -170,8 +175,8 @@ where
         );
 
         // reduce the degree by folding_factor at each iteration until the remaining polynomial
-        // is small enough; + 1 is for the remainder
-        for _ in 0..self.options.num_fri_layers(evaluations.len()) + 1 {
+        // is small enough
+        for _ in 0..self.options.num_fri_layers(evaluations.len()) {
             match self.folding_factor() {
                 2 => self.build_layer::<2>(channel, &mut evaluations),
                 4 => self.build_layer::<4>(channel, &mut evaluations),
@@ -181,9 +186,10 @@ where
             }
         }
 
+        self.set_remainder(channel, &mut evaluations);
+
         // make sure remainder length does not exceed max allowed value
-        let last_layer = &self.layers[self.layers.len() - 1];
-        let remainder_size = last_layer.evaluations.len();
+        let remainder_size = self.remainder.0.len();
         debug_assert!(
             remainder_size <= self.options.max_remainder_size(),
             "last FRI layer cannot exceed {} elements, but was {} elements",
@@ -209,12 +215,19 @@ where
         // projection to reduce the degree of evaluations by N
         let alpha = channel.draw_fri_alpha();
         *evaluations = apply_drp(&transposed_evaluations, self.domain_offset(), alpha);
-
         self.layers.push(FriLayer {
             tree: evaluation_tree,
             evaluations: flatten_vector_elements(transposed_evaluations),
             _base_field: PhantomData,
         });
+    }
+
+    /// Creates a FriRemainder from a vector of `evaluations` representing the remainder and
+    /// commits to the latter by hashing sequentially `evaluations`.
+    fn set_remainder(&mut self, channel: &mut C, evaluations: &mut [E]) {
+        let commitment = <H as ElementHasher>::hash_elements(evaluations);
+        channel.commit_fri_layer(commitment);
+        self.remainder = FriRemainder(evaluations.to_vec());
     }
 
     // QUERY PHASE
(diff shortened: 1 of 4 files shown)


fri/src/prover/mod.rs
```diff
@@ -103,6 +104,11 @@ struct FriLayer<B: StarkField, E: FieldElement<BaseField = B>, H: Hasher> {
     _base_field: PhantomData<B>,
 }
 
+struct FriLastLayer<B: StarkField, E: FieldElement<BaseField = B>> {
+    evaluations: Vec<E>,
+    _base_field: PhantomData<B>,
+}
```
irakliyk, line 110: Do we actually need `_base_field` here? Maybe we can simplify this as:
```Rust
struct FriLastLayer<E: FieldElement>(Vec<E>);
```
Also, I wonder if we should call it just `FriRemainder`?

fri/src/prover/mod.rs
```diff
@@ -171,7 +181,7 @@ where
 
         // reduce the degree by folding_factor at each iteration until the remaining polynomial
         // is small enough; + 1 is for the remainder
-        for _ in 0..self.options.num_fri_layers(evaluations.len()) + 1 {
+        for _ in 0..self.options.num_fri_layers(evaluations.len()) {
```
irakliyk, line 184: The comment above this line needs to be updated (I think we can just remove the "+1 ... " part of the comment.

fri/src/prover/mod.rs
```diff
@@ -209,14 +228,23 @@ where
...
-
         self.layers.push(FriLayer {
             tree: evaluation_tree,
             evaluations: flatten_vector_elements(transposed_evaluations),
             _base_field: PhantomData,
         });
     }
 
+    /// Builds a last FRI layer by hashing sequentially `evaluations`.
+    fn build_last_layer<const N: usize>(&mut self, channel: &mut C, evaluations: &mut [E]) {
```
irakliyk, line 239: It doesn't look like we use `N` type parameter here for anything - do we need it?

fri/src/prover/mod.rs
```diff
@@ -229,48 +257,41 @@ where
...
-            for j in 0..folding_factor {
-                remainder[i + n * j] = last_values[i * folding_factor + j];
+        if !self.layers.is_empty() {
+            let mut positions = positions.to_vec();
+            let mut domain_size = self.layers[0].evaluations.len();
+            let folding_factor = self.options.folding_factor();
+
+            // for all FRI layers, except the last one, record tree root, determine a set of query
+            // positions, and query the layer at these positions.
+            //let mut layers = Vec::with_capacity(self.layers.len());
```
irakliyk, line 269: nit: this commented out line can probably be deleted.

fri/src/prover/mod.rs
```diff
@@ -229,48 +257,41 @@ where
...
+                };
+
+                layers.push(proof_layer);
+                domain_size /= folding_factor;
             }
         }
 
+        // use the remaining polynomial values directly as proof; last layer values contain
+        // remainder.
+        let last_values = &self.last_layer.evaluations.to_vec();
```
irakliyk, line 289: nit: we can probably use `.clone()` here instead of `.to_vec()`.

Also, we probably don't need to do `.to_vec()` again on line 294.

fri/src/verifier/channel.rs
```diff
@@ -96,24 +96,19 @@ pub trait VerifierChannel<E: FieldElement> {
...
         let remainder = self.take_fri_remainder();
 
-        // build remainder Merkle tree
-        let remainder_values = transpose_slice(&remainder);
-        let hashed_values = hash_values::<Self::Hasher, E, N>(&remainder_values);
-        let remainder_tree = MerkleTree::<Self::Hasher>::new(hashed_values)
-            .map_err(|err| VerifierError::RemainderTreeConstructionFailed(format!("{err}")))?;
+        // sequential hash of the remainder
+        let set_commitment =
+            <<Self as VerifierChannel<E>>::Hasher as ElementHasher>::hash_elements(&remainder);
```
irakliyk, line 109: Do we still need `RemainderTreeConstructionFailed` error variant somewhere? If not, I'd delete it.

fri/src/prover/mod.rs
```diff
@@ -229,48 +257,41 @@ where
...
+
         // clear layers so that another proof can be generated
         self.reset();
```
irakliyk, line 292: In `reset()` we should probably clear the remainder as well.

fri/src/prover/mod.rs
```diff
@@ -229,48 +257,41 @@ where
     /// # Panics
     /// Panics is the prover state is clean (no FRI layers have been build yet).
     pub fn build_proof(&mut self, positions: &[usize]) -> FriProof {
-        assert!(
-            !self.layers.is_empty(),
-            "FRI layers have not been built yet"
-        );
```
irakliyk, line 235: I'd still keep a similar check here. What we can assert is that remainder should not be empty at this point. If it is, we haven't built FRI layers yet.

irakliyk (commented):
Nice improvement! Thank you! I left some small comments inline.

fri/src/prover/mod.rs
Al-Kindi-0: Sounds good, I use `remainder` throughout the code now instead.

fri/src/prover/mod.rs
Al-Kindi-0: Yes, thank you!

fri/src/prover/mod.rs
Al-Kindi-0: We don't need it since we are not building trees anymore. Fixed it!

fri/src/prover/mod.rs
Al-Kindi-0: Yes, thank you.

fri/src/prover/mod.rs
Al-Kindi-0: Done!

fri/src/verifier/channel.rs
Al-Kindi-0: Yes, fixed!

fri/src/prover/mod.rs
Al-Kindi-0: Yes, fixed!

fri/src/prover/mod.rs
Al-Kindi-0: Makes sense, I included the same assert with the same message again.

fri/src/prover/mod.rs
```diff
@@ -209,14 +223,20 @@ where
...
-
         self.layers.push(FriLayer {
             tree: evaluation_tree,
             evaluations: flatten_vector_elements(transposed_evaluations),
             _base_field: PhantomData,
         });
     }
 
+    /// Builds a last FRI layer by hashing sequentially `evaluations`.
+    fn build_remainder(&mut self, channel: &mut C, evaluations: &mut [E]) {
```
irakliyk, line 234: nit: since we are not really building the remainder here - would a more appropriate name for this be `set_remainder()`?

fri/src/prover/mod.rs
```diff
@@ -230,47 +250,44 @@ where
...
+                };
+
+                layers.push(proof_layer);
+                domain_size /= folding_factor;
             }
         }
 
+        // use the remaining polynomial values directly as proof; last layer values contain
+        // remainder.
+        let last_values = self.remainder.0.clone();
```
irakliyk, line 285: nit: let's change `last_values` to `remainder`?

fri/src/prover/mod.rs
```diff
@@ -181,9 +186,18 @@ where
...
         }
 
+        // compute commitment to last layer
+        match self.folding_factor() {
+            2 => self.build_remainder(channel, &mut evaluations),
+            4 => self.build_remainder(channel, &mut evaluations),
+            8 => self.build_remainder(channel, &mut evaluations),
+            16 => self.build_remainder(channel, &mut evaluations),
+            _ => unimplemented!("folding factor {} is not supported", self.folding_factor()),
+        }
```
irakliyk, line 196: This seems to be doing the same thing regardless of the branch. Should we change it simply to:
```Rust
self.build_remainder(channel, &mut evaluations);
```

irakliyk (changes requested):
Looks good! Thank you! Just a couple more minor comments.

fri/src/prover/mod.rs
Al-Kindi-0: We can use your suggestion as I couldn't come up with a better name than `set_remainder`.

fri/src/prover/mod.rs
Al-Kindi-0: Yes, that's more consistent.

fri/src/prover/mod.rs
Al-Kindi-0: Yes, fixed!

fri/src/prover/mod.rs
irakliyk: Seems like this didn't get changed in the last commit?

fri/src/prover/mod.rs
```diff
@@ -181,9 +186,11 @@ where
             }
         }
 
+        self.build_remainder(channel, &mut evaluations);
+
         // make sure remainder length does not exceed max allowed value
-        let last_layer = &self.layers[self.layers.len() - 1];
-        let remainder_size = last_layer.evaluations.len();
+        let remainder = &self.remainder;
+        let remainder_size = remainder.0.len();
```
irakliyk, line 193: nit: let's merge this into a single line as:
```Rust
let remainder_size = self.remainder.0.len();
```

Al-Kindi-0: Done!

fri/src/prover/mod.rs
Al-Kindi-0: Yes, changed now. Thanks!

fri/src/prover/mod.rs
```diff
@@ -230,43 +243,40 @@ where
...
+                };
+
+                layers.push(proof_layer);
+                domain_size /= folding_factor;
             }
         }
 
+        // use the remaining polynomial values directly as proof; last layer values contain
+        // remainder.
+        let remainder = self.remainder.0.clone();
```
irakliyk, line 278: one last comment: I think the comment here needs to be updated. probably removing "last year values contain remainder" could be good enough.

Al-Kindi-0: Fixed, thanks!

irakliyk (approved):
All looks good! Thank you!

PR #129, opened 2022-12-31
Rescue and griffin with Jive compression mode

Nashtare:
This PR aims at providing additional options for algebraic hash functions support, in particular in the light of two recent works:

- [Anemoi hash function and Jive compression mode](https://eprint.iacr.org/2022/840.pdf)
- [Griffin hash function](https://eprint.iacr.org/2022/403.pdf)

The PR implements two new instantiations of hash functions, more suited to Winterfell's purpose as tailored towards Merkle tree compression vs arbitrary data slice hashing:

- an instantiation of Rescue-Prime over f64 using Jive compression mode: it follows the same specification changes as `Rp64_256`, except that this instance has a state width of 8 (instead of 12). In particular the security margin has also been reduced to 40% instead of 50% to obtain 7 rounds.
- an instantiation of Griffin over f64 using Jive compression mode: hence also having a state width of 8. The security margin has been reduced to 15% instead of 20% to also obtain 7 rounds. However, (see below), the modified matrix for the linear layer is MDS and has a larger branch factor, which could alleviate some potential concerns regarding the margin reduction.

Both instances use the same MDS matrix, crafted in the spirit of `Rp64_256` matrix, i.e. a circulant matrix of size 8 with coefficients in frequency domain that are powers of two, to rely on FFTs for efficient matrix-vector product in the linear layer. Its first row is `[23, 8, 13, 10, 7, 6, 21, 8]`. I cannot thank @Al-Kindi-0 enough for helping me through the process and providing helper code. The best Christmas gift I got this year!

_Note on the MDS matrix choice_: I was actually planning on having some deterministic process to generate a valid one, but by taking the one of `Rp64_256` and trimming the first subvector to test my code, I realized the obtained matrix was actually MDS.

As for performance, I benchmarked the two functions (and compared with `Rp64_256`) for two variants of the fib-small example:

- Large but faster instance:
`./target/release/winterfell -e 3 -b 2 -q 112 --hash_fn HASH_FN fib-small -n 262144`

- Small but slower instance:
`./target/release/winterfell -e 3 -b 8 -q 42 --hash_fn HASH_FN fib-small -n 262144`

each targeting 128-bit security level.
The results on my machine (Ubuntu 22.04 - Intel® Core™ i7-9750H CPU @ 2.60GHz × 12) are the following:

- Large instance:
-- `Rp64_256`: Prover: 3100 ms / Verifier: 60 ms
-- `RpJive64_256`: Prover: 2100 ms / Verifier: 39 ms (i.e. 1.48x / 1.54x faster than `Rp64_256`)
-- `GriffinJive64_256`: Prover: 550 ms / Verifier: 17 ms (i.e. 5.64x / 3.53x faster than `Rp64_256`)

- Small instance:
-- `Rp64_256`: Prover: 10400 ms / Verifier: 35 ms 
-- `RpJive64_256`: Prover: 8400 ms / Verifier: 23 ms (i.e. 1.24x / 1.52x faster than `Rp64_256`)
-- `GriffinJive64_256`: Prover: 2000 ms / Verifier: 8 ms (i.e. 5.20x / 4.38x faster than `Rp64_256`)

I couldn't find some reference implementation of Griffin in Python for easily testing the code, so I made one [here](https://github.com/Nashtare/griffin-hash).

 I guess this will take time to review / discuss as it's holidays season, but I wanted to share this before 2022 ends!
 Happy new year! 🥳


Nashtare:
In particular, and in addition to these two illustrative instantiations, the Jive compression mode could perfectly be combined with the recent [Rescue-Prime Optimized](https://eprint.iacr.org/2022/1577.pdf) instantiation used in Miden to further improve the performance of the VM, until perhaps the community has enough confidence in using recent, more efficient constructions like Griffin.

crypto/README.md
```diff
@@ -15,6 +16,7 @@ Rescue hash function is implemented according to the Rescue Prime [specification
...
   - Instead of using Vandermonde matrices as a standard way of generating an MDS matrix as described in Rescue Prime paper, we use a methodology developed by Polygon Zero to find an MDS matrix with coefficients which are small powers of two in frequency domain. This allows us to dramatically reduce MDS matrix multiplication time. Using a different MDS matrix does not affect security of the hash function as any MDS matrix satisfies Rescue Prime construction (as described in section 4.2 of the paper).
+* The `RPJive64_256` instantiation of Rescue Prime using Jive as compression mode implements similar modifications than `Rp64_256`. However, because of the use of Jive, the output of the hash function is not the same when we hash 8 field elements as a sequence of elements using `hash_elements()` function and when we compress 8 field elements into 4 (e.g., for building a Merkle tree) using the 2-to-1 Jive compression mode.
```
irakliyk, line 19: nit: should it be "... similar modifications as ..."?

crypto/README.md
```diff
@@ -34,16 +44,34 @@ The parameters used to instantiate the functions are:
   - S-Box degree: 3.
   - Target security level: 124-bits.
 
+### Griffin hash function implementation
+Griffin hash function is implemented according to the Griffin [specifications](https://eprint.iacr.org/2022/403.pdf) with the following differences:
+* We set the number of rounds to 7, which implies a 15% security margin instead of the 20% margin used in the specifications (a 20% margin rounds up to 8 rounds), similarly to the motivation on Rescue above.
+* When hashing a sequence of elements, we do not append Fp(1) followed by Fp(0) elements to the end of the sequence as padding. Instead, we initialize one of the capacity elements to the number of elements to be hashed, and pad the sequence with Fp(0) elements only. However, this also means that our instantiation of Griffin cannot be used in a stream mode as the number of elements to be hashed must be known upfront.
```
irakliyk, line 50: There are actually some concerns about potential security degradation from using this padding rule. It might be better to use a padding rule similar to the one used in RPO.

crypto/benches/hash.rs
```diff
@@ -19,6 +19,9 @@ type Sha3Digest = <Sha3 as Hasher>::Digest;
 
 type Rp62_248Digest = <Rp62_248 as Hasher>::Digest;
 type Rp64_256Digest = <Rp64_256 as Hasher>::Digest;
+type RpJive64_256Digest = <RpJive64_256 as Hasher>::Digest;
+
+type GriffinJive64_256Digest = <GriffinJive64_256 as Hasher>::Digest;
```
irakliyk, line 24: minor nit: this blank line can probably be removed.

crypto/README.md
```diff
@@ -34,16 +44,34 @@ The parameters used to instantiate the functions are:
...
-| CPU                         | BLAKE3_256 | SHA3_256 | RP64_256 | RP62_248 |
-| --------------------------- | :--------: | :------: | :------: | :------: |
-| Apple M1 Pro                | 76 ns      | 227 ns   | 5.1 us   | 7.1 us   |
-| AMD Ryzen 9 5950X @ 3.4 GHz | 62 ns      | 310 ns   | 5.2 us   | 6.9 us   |
-| Core i9-9980KH @ 2.4 GHz    | 66 ns      | 400 ns   | -        | 6.6 us   |
-| Core i5-7300U @ 2.6 GHz     | 81 ns      | 540 ns   | -        | 9.5 us   |
-| Core i5-4300U @ 1.9 GHz     | 106 ns     | 675 ns   | -        | 13.9 us  |
+| CPU                         | BLAKE3_256 | SHA3_256 | RP64_256 | RPJ64_256 | RP62_248 | GriffinJ64_256 |
+| --------------------------- | :--------: | :------: | :------: | :-------: | :------: | :------------: |
+| Apple M1 Pro                | 76 ns      | 227 ns   | 5.1 us   | TODO us   | 7.1 us   | TODO us        |
```
irakliyk, line 70: On an M1, I'm getting the following results:

- RPJ64_256: 3.8 us
- GriffinJ64_256: 2.2 us

crypto/src/hash/rescue/rp64_256_jive/mod.rs
```diff
@@ -0,0 +1,749 @@
...
+        // apply the Rescue permutation and apply the final Jive summation
+        Self::apply_permutation(&mut state);
+
+        let mut result = [BaseElement::ZERO; DIGEST_SIZE];
+        for (i, r) in result.iter_mut().enumerate() {
+            *r = initial_state[i]
+                + initial_state[DIGEST_SIZE + i]
+                + state[i]
+                + state[i + DIGEST_SIZE];
+        }
```
irakliyk, line 200: Since the same code is also used on lines 228 - 234, we should probably put this into a stand-alone function. Maybe something like `apply_jive_summation()`?

Also minor nit: on line 197 we use `DIGEST_SIZE + i` but on line 199 we use `i + DIGEST_SIZE`. I'd probably make them consistent.

irakliyk (changes requested):
This is very cool! Thank you for doing this.

This is not a full review yet - I haven't gotten to reviewing Griffin yet. But I've included a few small comments for the other parts for now.

crypto/README.md
Nashtare: Thanks, will update it!

crypto/README.md
Nashtare: The Hirose padding? Sounds fair, we've used this also in Anemoi.
Should I wait until further discussion is made or update it right away?

crypto/src/hash/rescue/rp64_256_jive/mod.rs
Nashtare: `Since the same code is also used on lines 228 - 234, we should probably put this into a stand-alone function. Maybe something like apply_jive_summation()?`

Definitely. I had the same remark for MDS related functions (the fft2 / fft4 operations).

crypto/src/hash/mod.rs
```diff
@@ -14,7 +14,10 @@ mod sha;
 pub use sha::Sha3_256;
 
 mod rescue;
-pub use rescue::{Rp62_248, Rp64_256};
+pub use rescue::{Rp62_248, Rp64_256, RpJive64_256};
+
+mod griffin;
+pub use griffin::GriffinJive64_256;
```
Nashtare, line 20: I was also not sure about the final naming on the module and the struct.
Same goes for the instantiation of RP.

crypto/README.md
Nashtare: The Hirose padding? Sounds fair, we've used this also in Anemoi.
Should I wait until further discussion is made or update it right away?

crypto/src/hash/griffin/griffin64_256_jive/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+///   mode as the number of elements to be hashed must be known upfront.
+/// * Instead of using the suggested matrix as described in Griffin paper, we use a methodology
+///   developed by Polygon Zero to find an MDS matrix with coefficients which are small powers
+///   of two in frequency domain. This allows us to dramatically reduce matrix multiplication
+///   time. We claim without proof that using a different (MDS) matrix does not affect security of
+///   as the branching number is actually increased (9 instead of 6).
+///
+/// The parameters used to instantiate the function are:
+/// * Field: 64-bit prime field with modulus 2^64 - 2^32 + 1.
+/// * State width: 12 field elements.
```
Al-Kindi-0, line 71: 8 instead of 12?

Al-Kindi-0 (commented):
Left only few nit-comments, else looks good to me!
As for the new padding rule, maybe we can do it in a separate PR and do it for all concerned hash functions?

crypto/README.md
```diff
@@ -34,16 +44,34 @@ The parameters used to instantiate the functions are:
...
+| AMD Ryzen 9 5950X @ 3.4 GHz | 62 ns      | 310 ns   | 5.2 us   | TODO us   | 6.9 us   | TODO us        |
+| Core i9-9980KH @ 2.4 GHz    | 66 ns      | 400 ns   | -        | -         | 6.6 us   | -              |
+| Core i5-7300U @ 2.6 GHz     | 81 ns      | 540 ns   | -        | -         | 9.5 us   | -              |
+| Core i5-4300U @ 1.9 GHz     | 106 ns     | 675 ns   | -        | -         | 13.9 us  | -              |
 
 As can be seen from the table, BLAKE3 is by far the fastest hash function, while our implementation of Rescue Prime is roughly 100x slower than BLAKE3 and about 20x slower than SHA3.
```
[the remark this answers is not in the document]
Nashtare: (also if we refer to RP64_256, it seems it's rather roughly 70x slower than BLAKE3 (exactly 67), compared to 100x)

Nashtare:
> As for the new padding rule, maybe we can do it in a separate PR and do it for all concerned hash functions?

I can also take care of it in this PR if desired.

crypto/src/hash/griffin/griffin64_256_jive/mds_freq.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    let [s0, s2, s4, s6] = ifft4_real((v0, v1, v2));
+    let [s1, s3, s5, s7] = ifft4_real((v4, v5, v6));
+
+    [s0, s1, s2, s3, s4, s5, s6, s7]
+}
+
+// We use the real FFT to avoid redundant computations. See https://www.mdpi.com/2076-3417/12/9/4700
+#[inline(always)]
+fn fft2_real(x: [u64; 2]) -> [i64; 2] {
```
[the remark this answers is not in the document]
Nashtare: @irakliyk what do you think about this? Should we group only the fft2 / fft4 methods? Or have a module for each different MDS matrices as well? (as the two new hash instantiations use the same 8x8 MDS matrix).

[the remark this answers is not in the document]
irakliyk: @Nashtare - I like this idea! Maybe we create another module under `fft` called `real`? But I'm also fine with other suggestions.

crypto/README.md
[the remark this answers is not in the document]
irakliyk: Yes, agreed - it is probably a good idea to re-write this sentence. Maybe we can say that algebraic hashes are between x (Griffin) and 70x (Rescue Prime) slower than BLAKE3 - or something similar.

irakliyk:
> As for the new padding rule, maybe we can do it in a separate PR and do it for all concerned hash functions?

I'd probably do it in this PR for at least the newly added hash functions, and we could do the other ones in a subsequent PR.

crypto/src/hash/griffin/griffin64_256_jive/mds_freq.rs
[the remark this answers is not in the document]
irakliyk: I also wonder if we should somehow consolidate MDS matrixes. i.e., both `GriffingJive` and `RpJive` seem to use the same matrix and we define it twice.

crypto/src/hash/rescue/rp64_256_jive/mod.rs
```diff
@@ -0,0 +1,751 @@
...
+            Self::apply_permutation(&mut state);
+        }
+
+        // return the first 4 elements of the state as hash result
+        ElementDigest::new(state[DIGEST_RANGE].try_into().unwrap())
+    }
+}
+
+#[inline(always)]
+fn apply_jive_summation(
```
irakliyk, line 265: I wonder if this should go into `RpJive64_256` and be a public function (similar to `apply_round()`). Might be useful for building execution traces and maybe for defining AIR constraints as well.

crypto/src/hash/griffin/griffin64_256_jive/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+            Self::apply_permutation(&mut state);
+        }
+
+        // return the first 4 elements of the state as hash result
+        ElementDigest::new(state[DIGEST_RANGE].try_into().unwrap())
+    }
+}
+
+#[inline(always)]
+fn apply_jive_summation(
```
irakliyk, line 246: I wonder if this should go into `GriffinJive64_256` and be a public function (similar to `apply_round()`). Might be useful for building execution traces and maybe for defining AIR constraints as well.

irakliyk (changes requested):
Looks good! Thank you! (and sorry again for the delay). I left a few small comments inline. Besides these, I think the following is left:

1. Putting some parts of MDS matrix multiplication in frequency domain into a common location.
2. Potentially putting same MDS matrixes into a common location.
3. Updating the padding rule for the newly added hash functions.
4. Minor updates to readme (and I still need to run the benchmarks on an AMD machine).

Let me know if I missed anything.

Nashtare:
@irakliyk I've addressed all the comments. Regarding the move of fft / mds related components:

- I put the fft helper methods in fft/real_64, specifying the size instead of making them generic as they are only used for the MDS products, I can adapt it if required
- I created an mds module within crypto/hash grouping the different mds matrice and product implementations

I've also updated the griffin SAGE implementation with Hirose padding in case it is needed for review (I had to update one test in griffin impl with the change, which is not there in the other hash functions where we test only the underlying permutation).
Let me know if there's anything else you'd like to change (though I believe now it should be only cosmetic changes, as we've reviewed all the core components).

irakliyk:
@Nashtare - thank you! While looking through it I noticed that the last commit also adds Anemoi hash function? Was that intentional? (asking because it's not mentioned in the docs or description).

irakliyk (commented):
All looks great! Thank you! The changes are exactly what I would have done as well. I haven't yet reviewed the new Anemoi code as I wasn't sure if the commit was intentional - but once we figure out what to do with it, we can merge this PR.

Also, I've run the benchmarks on the AMD machine and get the following:
- RPJ64_256: 3.9 us
- Griffin64_256: 1.5 us

Nashtare:
Oups sorry, I had added it some time ago to check Merkle tree benches but I was not planning to include it, as its not as well tailored for STARKs than Griffin or RP. Let me remove it 😅

irakliyk (approved):
All looks good! Thank you very much! (and sorry for such huge delays).

Issue #96, opened 2022-06-07
Indexing outside of bounds winterfell v0.3.0


nattrad:
Hello,

I tried running the STARK VDF (https://github.com/bobbinth/stark-vdf) using winterfell and when I run with `cargo run`, i get this: 

![image](https://user-images.githubusercontent.com/107082480/172498240-3c34b6e5-8192-48cc-a2e1-87c9a2536c18.png)

To fix I ran it with `cargo run --release` and get the following output:

![image](https://user-images.githubusercontent.com/107082480/172498365-75e4b28d-3bd7-4dc3-9b6c-89b0d748498c.png)

irakliyk:
Fixed by #130

PR #130, opened 2023-01-04
Concurrent Execution Panic Fix

0xKanekiKen:
This fix addresses #96 and [0xPolygonMiden/miden-vm/#602](https://github.com/0xPolygonMiden/miden-vm/issues/602). I've summarized the fix [here](https://github.com/0xPolygonMiden/miden-vm/issues/602#issuecomment-1370845674).


irakliyk (approved):
Good catch! Thank you for the fix!

If I understood correctly, when we have a single fragment `i` and `step` are the same - so, all works as expected. But in multi-threaded evaluation, we have multiple fragments and for any fragment beyond the first one `i` and `step` are different, and thus would cause on overflow.

0xKanekiKen:
Yes, you got it right! I should elaborate it a bit more on the issue why the code was not panicking when we have a single fragment.

PR #132, opened 2023-01-18
Fix concurrent fft

### LINKED ISSUE #131: Build fails with concurrent feature
PR https://github.com/facebook/winterfell/pull/124 broke the build using concurrent feature, by changing the scope of the `fft_in_place` function (originally in `fft/serial` module) to private, and moving it to the `fft/fft_inputs` module without updating the calls in the `fft/concurrent` module.

I think this already happened in the past that a change breaking builds with parallelization went unnoticed in the main branch, perhaps we should consider having a check in the CI?

Nashtare:
This PR fixes build with `concurrent` feature.
@irakliyk I have included a new CI job checking that we can build with all features activated, to prevent this kind of issue happening in the future. It is only checking the compilation, should it run tests as well instead?

closes #131

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,27 @@ on:
     types: [opened, repoened, synchronize]
 
 jobs:
+  check:
+    name: Check with all features
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        toolchain: [stable]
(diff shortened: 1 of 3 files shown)


irakliyk (approved):
This is great! Thank you for fixing it and adding an extra CI job. We should have probably done it a long time ago.

Regarding tests in concurrent mode: it would be great, but I also think it could be done in another PR.

PR #134, opened 2023-01-25
Add exp_vartime method for FieldElement trait

Nashtare:
Add `exp_vartime` method to the `FieldElement` trait to speed-up verifier exponentiations.
This is currently useful only for `f64` which implements by default constant-time operations (`exp_vartime` there is about 40% faster than its constant-time counterpart), but could be useful if we were to implement constant-time operations by default for all other supported fields.
For the fib-small example over f64, the verifier speed-up is measured at around 8% on my machine.

### DIFF
--- a/fri/src/verifier/mod.rs
+++ b/fri/src/verifier/mod.rs
@@ -235,7 +235,7 @@ where
         let folding_roots = (0..N)
             .map(|i| {
                 self.domain_generator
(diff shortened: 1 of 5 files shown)


irakliyk (approved):
Great improvement! I wanted to do something like this for a while myself - but never got to it. Thank you!

PR #135, opened 2023-01-26
`ProofOptions` Refactor

MavenRain:
ProofOptions is made pure (returns errors instead of panicking), and is refactored to make it much harder to initialize with invalid options.

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -5,7 +5,9 @@
 
 use fri::FriOptions;
 use math::StarkField;
-use utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable};
+use utils::{
+    ByteReader, ByteWriter, Deserializable, DeserializationError, ProofOptionError, Serializable,
+};
 
 // TYPES AND INTERFACES
 // ================================================================================================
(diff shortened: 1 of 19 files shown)


air/src/options.rs
```diff
@@ -34,6 +36,157 @@ pub enum FieldExtension {
     Cubic = 3,
 }
 
+/// Represents the FRI folding factor option, a power of two
+#[repr(u8)]
+pub enum FriFoldingFactor {
```
Nashtare, line 41: I am not really a fan of this approach, which I feel is too heavy for what it aims at achieving. For instance supporting a different folding factor (eg 2), would require multiple changes across the library, plus break the existing ordering (as First would become 2 here, and shift all the other values).

air/src/options.rs
```diff
@@ -34,6 +36,157 @@ pub enum FieldExtension {
...
+            .map_err(|_| {
+                DeserializationError::UnknownError(format!("Unexpected blowup factor of {value}"))
+            })
+            .and_then(TryInto::try_into)
+    }
+}
+
+/// Represents the max remainder size proof option, as a power of 2
+#[repr(u16)]
+pub enum FriMaximumRemainderSize {
```
Nashtare, line 136: Same comment as for FoldingFactor.

air/src/options.rs
```diff
@@ -63,7 +216,7 @@ pub struct ProofOptions {
     grinding_factor: u8,
     field_extension: FieldExtension,
     fri_folding_factor: u8,
-    fri_max_remainder_size: u8, // stored as power of 2
+    fri_max_remainder_size: u32,
```
Nashtare, line 219: any particular reason why this is increased? (I'm not against, but as it is orthogonal to the refactoring, I'm asking)

Nashtare (commented):
I'm a bit skeptical about the approach overall.
Typically, keeping only the Error management, and doing the checks seems enough, without the overhead of having all the custom enums and back and forth conversions. This is what you've done with the grinding factor (modulo the primitive type change) which is still an integer, with a check
`(_, factor) if factor > 32 => Err(ProofOptionError::GrindingFactor),`
when creating the ProofOption which is enough in this context.

air/src/options.rs
MavenRain: @Nashtare As it currently stands, `2` isn't a legal value, and the point of the enum is to ensure that you couldn't even try an illegal value if you wanted to do so.  From that standpoint, wanting to make `2` legal seems like it _should_ be a breaking change.

Another way of looking at it is that, if we wanted it to be _easy_ to assume a currently illegal value in the near future, then this proposal invites the question of the wisdom of _panicking_ on such a value in the first place, as opposed to something lighter, like a warning.

air/src/options.rs
MavenRain: @Nashtare In fact, it looks like we were previously getting lucky here.  This field is currently inhabited by the result of calling `trailing_zeros` on a primitive integer type.  That call always returns a `u32`, and that means that there are a lot of values that call could return that are outside the bounds of what is representable as a `u8`.

Nashtare: Ah indeed, good point.

air/src/options.rs
Nashtare: > As it currently stands, 2 isn't a legal value, and the point of the enum is to ensure that you couldn't even try an illegal value if you wanted to do so. From that standpoint, wanting to make 2 legal seems like it should be a breaking change.

I don't agree on this. If we were to support additional folding factors in the future, in no way this would be a breaking change. In the sense that code successfully compiling with winterfell as backend would keep on behaving correctly. Handling a previously erroneous case such that it is in the future a valid possibility is not a breaking change, but an extension of existing feature.

In addition (see my main review comment), we can still handle the panics by replacing the asserts with equivalent conditional early error returns (similarly to what you've done on `ProofOptions::new()`, without needing all those enums.

MavenRain: @Nashtare Truth be told, I wish I had not needed to resort to the conditional error returns for any parameter.  The problem with the grinding factor and the query number is that there are still too many legal values to restrain those to an enum.  Ideally, the fewer legal states can be accepted, the less states there are to consider for coverage.

Can I ask what your issue with enums is here?  For me, enums increase readability, and so I tend to prefer more to less, but it seems to be the reverse for you?

irakliyk:
Closing as stale.

PR #138, opened 2023-02-03
Optimized methods for squaring in extension fields of degree 2 & 3

Al-Kindi-0:
This PR implements more optimized methods for computing squares in the degrees 2 and 3 extension fields. This is in contrast to the old way of squaring which was using a generic multiplication.
Also included, is a method for multiplying field elements in 64-bit field with constants smaller than $2^{32}$. This will be useful for extension fields that are defined by irreducible polynomials of the form $X^d - \omega$ where $d$ is the degree of the extension and $\omega$ is a field element. In the case of such extension fields, reduction modulo $X^d - \omega$ becomes just a multiplication by $\omega$ and thus a method for multiplying with small constants will be useful.

### DIFF
--- a/math/src/field/extensions/cubic.rs
+++ b/math/src/field/extensions/cubic.rs
@@ -70,6 +70,12 @@ impl<B: ExtensibleField<3>> FieldElement for CubeExtension<B> {
     }
 
     #[inline]
+    fn square(self) -> Self {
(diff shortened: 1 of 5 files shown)


irakliyk (approved):
All looks good! Thank you!

PR #139, opened 2023-02-07
Send remainder polynomial instead of remainder code word

Al-Kindi-0:
This PR proposes another method for performing the final check in the FRI query stage. Specifically, in the current implementation the verifier receives a code word from the prover which the latter claims to be consistent with an honest execution of the protocol.
The verifier then checks that the commitment to the received code word matches the one received in the commit phase of the protocol. The verifier then goes on to interpolate the code word in order to find the corresponding remainder polynomial. The verifier concludes by checking whether the degree of this polynomial is less than the expected degree bound (which can be computed from the input polynomial degree bound and the protocol parameters).
The current proposal simplifies things by sending the remainder polynomial directly. Thus the verifier can perform the degree check almost trivially. Moreover, no commitments to the remainder polynomial are needed.

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -63,7 +63,7 @@ pub struct ProofOptions {
     grinding_factor: u8,
     field_extension: FieldExtension,
     fri_folding_factor: u8,
-    fri_max_remainder_size: u8, // stored as power of 2
+    fri_remainder_max_degree_plus_1: u8, // stored as power of 2
 }
 
 // PROOF OPTIONS IMPLEMENTATION
@@ -89,7 +89,7 @@ impl ProofOptions {
     /// * `blowup_factor` is smaller than 4, greater than 256, or is not a power of two.
     /// * `grinding_factor` is greater than 32.
     /// * `fri_folding_factor` is not 4, 8, or 16.
-    /// * `fri_max_remainder_size` is smaller than 32, greater than 1024, or is not a power of two.
+    /// * `fri_remainder_max_degree_plus_1` is 0, greater than 32.
     #[rustfmt::skip]
     pub fn new(
         num_queries: usize,
@@ -97,7 +97,7 @@ impl ProofOptions {
         grinding_factor: u32,
         field_extension: FieldExtension,
         fri_folding_factor: usize,
-        fri_max_remainder_size: usize,
+        fri_remainder_max_size_plus_1: usize,
     ) -> ProofOptions {
         // TODO: return errors instead of panicking
         assert!(num_queries > 0, "number of queries must be greater than 0");
@@ -114,9 +114,9 @@ impl ProofOptions {
         assert!(fri_folding_factor >= 4, "FRI folding factor cannot be smaller than 4");
         assert!(fri_folding_factor <= 16, "FRI folding factor cannot be greater than 16");
 
-        assert!(fri_max_remainder_size.is_power_of_two(), "FRI max remainder size must be a power of 2");
-        assert!(fri_max_remainder_size >= 32, "FRI max remainder size cannot be smaller than 32");
-        assert!(fri_max_remainder_size <= 1024, "FRI max remainder size cannot be greater than 1024");
+        assert!(fri_remainder_max_size_plus_1.is_power_of_two(), "FRI polynomial remainder degree plus 1 must be a power of 2");
+        assert!(fri_remainder_max_size_plus_1 >= 1, "FRI polynomial remainder degree plus 1 cannot be 0");
+        assert!(fri_remainder_max_size_plus_1 <= 32, "FRI polynomial remainder degree plus 1 cannot be greater than 32");
 
         ProofOptions {
             num_queries: num_queries as u8,
@@ -124,7 +124,7 @@ impl ProofOptions {
             grinding_factor: grinding_factor as u8,
             field_extension,
             fri_folding_factor: fri_folding_factor as u8,
-            fri_max_remainder_size: fri_max_remainder_size.trailing_zeros() as u8,
+            fri_remainder_max_degree_plus_1: fri_remainder_max_size_plus_1.trailing_zeros() as u8,
         }
     }
 
@@ -181,8 +181,12 @@ impl ProofOptions {
     /// Returns options for FRI protocol instantiated with parameters from this proof options.
     pub fn to_fri_options(&self) -> FriOptions {
         let folding_factor = self.fri_folding_factor as usize;
-        let max_remainder_size = 2usize.pow(self.fri_max_remainder_size as u32);
-        FriOptions::new(self.blowup_factor(), folding_factor, max_remainder_size)
+        let remainder_max_degree_plus_1 = 2usize.pow(self.fri_remainder_max_degree_plus_1 as u32);
+        FriOptions::new(
+            self.blowup_factor(),
+            folding_factor,
+            remainder_max_degree_plus_1,
+        )
     }
 }
 
@@ -194,7 +198,7 @@ impl Serializable for ProofOptions {
         target.write_u8(self.grinding_factor);
         target.write(self.field_extension);
         target.write_u8(self.fri_folding_factor);
-        target.write_u8(self.fri_max_remainder_size);
+        target.write_u8(self.fri_remainder_max_degree_plus_1);
     }
 }
 
--- a/fri/src/options.rs
+++ b/fri/src/options.rs
@@ -12,7 +12,7 @@ use math::StarkField;
 #[derive(Clone, PartialEq, Eq)]
 pub struct FriOptions {
     folding_factor: usize,
-    max_remainder_size: usize,
+    remainder_max_degree_plus_1: usize,
     blowup_factor: usize,
 }
 
@@ -23,8 +23,11 @@ impl FriOptions {
     /// Panics if:
     /// * `blowup_factor` is not a power of two.
     /// * `folding_factor` is not 2, 4, 8, or 16.
-    /// * `max_remainder_size` is not at least twice the size of the `folding_factor`.
-    pub fn new(blowup_factor: usize, folding_factor: usize, max_remainder_size: usize) -> Self {
+    pub fn new(
+        blowup_factor: usize,
+        folding_factor: usize,
+        remainder_max_degree_plus_1: usize,
+    ) -> Self {
         // TODO: change panics to errors
         assert!(
             blowup_factor.is_power_of_two(),
@@ -37,15 +40,9 @@ impl FriOptions {
                 || folding_factor == 16,
             "folding factor {folding_factor} is not supported"
         );
-        assert!(
-            max_remainder_size >= folding_factor * 2,
-            "expected max remainder size to be at least {}, but was {}",
-            folding_factor * 2,
-            max_remainder_size
-        );
         FriOptions {
             folding_factor,
-            max_remainder_size,
+            remainder_max_degree_plus_1,
             blowup_factor,
         }
     }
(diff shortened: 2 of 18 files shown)


fri/src/prover/mod.rs
```diff
@@ -222,12 +213,13 @@ where
         });
     }
 
-    /// Creates a FriRemainder from a vector of `evaluations` representing the remainder and
-    /// commits to the latter by hashing sequentially `evaluations`.
-    fn set_remainder(&mut self, channel: &mut C, evaluations: &mut [E]) {
-        let commitment = <H as ElementHasher>::hash_elements(evaluations);
-        channel.commit_fri_layer(commitment);
-        self.remainder = FriRemainder(evaluations.to_vec());
+    /// Creates remainder polynomial from a vector of `evaluations` representing the remainder
```
Nashtare, line 216: I'd perhaps reword this to emphasize that we create a poly in coeff form, from its evaluations. But not a big deal

fri/src/verifier/mod.rs
```diff
@@ -378,3 +346,11 @@ fn get_query_values<E: FieldElement, const N: usize>(
 
     result
 }
+
+// Evaluates a polynomial with coefficients in an extension field at a point in the base field.
+pub fn eval_horner<E>(p: &[E], x: E) -> E
```
Nashtare, line 351: The description of the method doesn't match it's actual behavior, as both `p` coefficients and the target point `x` are in the same field `E`. Plus when we're calling the horner evaluation in the verification, we're converting `x` from `E::BaseField` to `E`.

Nashtare (commented):
I took the liberty to add some light comments.
Also, I think you forgot to update in `examples/src/lib.rs` the default parameters of `to_proof_options()` (it's still using 256 for the now `fri_remainder_size` field.

air/src/options.rs
```diff
@@ -89,15 +89,15 @@ impl ProofOptions {
     /// * `blowup_factor` is smaller than 4, greater than 256, or is not a power of two.
     /// * `grinding_factor` is greater than 32.
     /// * `fri_folding_factor` is not 4, 8, or 16.
-    /// * `fri_max_remainder_size` is smaller than 32, greater than 1024, or is not a power of two.
+    /// * `fri_remainder_size` is smaller than 1, greater than 32, or is not a power of two.
```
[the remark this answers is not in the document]
Al-Kindi-0: Done!

fri/src/verifier/mod.rs
Al-Kindi-0: Thank you! I updated it to be more consistent with my original way of thinking. Let me know if it works now.

fri/src/verifier/mod.rs
```diff
@@ -378,3 +346,14 @@ fn get_query_values<E: FieldElement, const N: usize>(
 
     result
 }
+
+// Evaluates a polynomial with coefficients in an extension field at a point in the base field.
+pub fn eval_horner<E, B>(p: &[E], x: B) -> E
+where
+    E: FieldElement<BaseField = B>,
+    B: FieldElement,
```
Nashtare, line 354: Couldn't we abstract the second type and refer to `E::BaseField`? Something like:

```suggestion
pub fn eval_horner<E>(p: &[E], x: E::BaseField) -> E
where
    E: FieldElement,
```

fri/src/verifier/mod.rs
Nashtare: Thanks! I've added a comment, I think we can go down to a single generic parameter to make the function definition lighter.

Al-Kindi-0: Thanks @Nashtare ! I looks cleaner indeed.

Nashtare: Sorry for breaking the build, we need to update also when the function is called, to remove the extra `B` 😅

Al-Kindi-0: Sure thing, I updated it now :-)

Al-Kindi-0:
Thank you for the review!

fri/src/verifier/mod.rs
```diff
@@ -303,56 +303,24 @@ where
...
         for (&position, evaluation) in positions.iter().zip(evaluations) {
-            if remainder[position] != evaluation {
+            let comp_eval = eval_horner::<E>(
+                &remainder,
+                offset * domain_generator.exp((position as u64).into()),
+            );
+            if comp_eval != evaluation {
                 return Err(VerifierError::InvalidRemainderFolding);
             }
         }
```
irakliyk, line 321: A couple of small nits here:
- I'd probably rename `remainder` into `remainder_poly` to indicate that it is a polynomial in coefficient form.
- I'd use `exp_vartime()` instead of `exp()` here to make exponentiation potentially faster.

fri/src/verifier/channel.rs
```diff
@@ -90,26 +90,17 @@ pub trait VerifierChannel<E: FieldElement> {
...
-
-        // sequential hash of the remainder
-        let set_commitment =
-            <<Self as VerifierChannel<E>>::Hasher as ElementHasher>::hash_elements(&remainder);
-
-        if *commitment != set_commitment {
-            return Err(VerifierError::RemainderCommitmentMismatch);
+        if polynom::degree_of(&remainder) > max_degree {
+            return Err(VerifierError::RemainderDegreeMismatch(max_degree));
         }
```
irakliyk, line 104: I think it would be cleaner to perform this check in the verifier (i.e., in `verify_generic()` function).

Also, can we simplify this as:
```Rust
if remainder.len() > max_degree + 1 {
    return Err(VerifierError::RemainderDegreeMismatch(max_degree));
}
```

air/src/options.rs
```diff
@@ -63,7 +63,7 @@ pub struct ProofOptions {
     grinding_factor: u8,
     field_extension: FieldExtension,
     fri_folding_factor: u8,
-    fri_max_remainder_size: u8, // stored as power of 2
```
irakliyk, line 66: Should we name this `fri_max_remainder_degree`? I think this would better reflect what it actually is. We could still store it as a power of two, but now we'd be storing `log2(max_degree + 1)`.

air/src/options.rs
```diff
@@ -114,17 +115,17 @@ impl ProofOptions {
...
+        assert!(fri_remainder_size >= 1, "FRI polynomial remainder size cannot be 0");
+        assert!(fri_remainder_size <= 32, "FRI polynomial remainder size cannot be greater than 32");
 
         ProofOptions {
             num_queries: num_queries as u8,
             blowup_factor: blowup_factor as u8,
             grinding_factor: grinding_factor as u8,
             field_extension,
             fri_folding_factor: fri_folding_factor as u8,
-            fri_max_remainder_size: fri_max_remainder_size.trailing_zeros() as u8,
```
irakliyk, line 127: If we store remainder degree, this would be `log2(fri_max_remainder_degree + 1)`.

fri/src/options.rs
```diff
@@ -12,7 +12,7 @@ use math::StarkField;
 #[derive(Clone, PartialEq, Eq)]
 pub struct FriOptions {
     folding_factor: usize,
-    max_remainder_size: usize,
+    remainder_size: usize,
```
irakliyk, line 15: Similar comment to the one before: should we rename this into `max_remainder_degree`?

fri/src/prover/mod.rs
```diff
@@ -222,12 +213,12 @@ where
...
-        let commitment = <H as ElementHasher>::hash_elements(evaluations);
-        channel.commit_fri_layer(commitment);
-        self.remainder = FriRemainder(evaluations.to_vec());
+    /// Creates remainder polynomial in coefficient form from a vector of `evaluations` over a domain.
+    fn set_remainder(&mut self, evaluations: &mut [E]) {
+        let inv_twiddles = fft::get_inv_twiddles(evaluations.len());
+        fft::interpolate_poly_with_offset(evaluations, &inv_twiddles, E::BaseField::GENERATOR);
+        let remainder = evaluations[..evaluations.len() / self.options.blowup_factor()].to_vec();
+        self.remainder = FriRemainder(remainder);
     }
```
irakliyk, line 222: A few comments/questions here:
- Should we get the domain offset from options here rather than using `E::BaseField::GENERATOR`?
- One of the things that we did in `commit_fri_layer()` was reseeding of the random coin. I think we might still need to do this, but now we should reseed it with the entire remainder polynomial - right? If so, it might be more efficient to commit to the remainder polynomial and then use this commitment to reseed the public coin with this commitment.
- minor nit: I'd probably rename `remainder` variable into `remainder_poly`.

irakliyk (changes requested):
Looks great! Thank you! I left several comments inline.

fri/src/verifier/mod.rs
Al-Kindi-0: Done!

fri/src/options.rs
Al-Kindi-0: I have renamed all related variables as `max_remainder_degree_plus_1`. It is a bit verbose but accurate.

air/src/options.rs
Al-Kindi-0: Makes sense.

fri/src/verifier/channel.rs
Al-Kindi-0: Done!

fri/src/verifier/channel.rs
```diff
@@ -90,28 +90,16 @@ pub trait VerifierChannel<E: FieldElement> {
         Ok(group_vector_elements(layer_queries))
     }
 
-    /// Returns FRI remainder values (last FRI layer) read from this channel.
+    /// Returns FRI remainder polynomial read from this channel.
     ///
-    /// This also checks whether the remainder is valid against the provided commitment.
+    /// This also
     ///
```
irakliyk, line 96: nit: we should probably delete these lines.

fri/src/verifier/channel.rs
```diff
@@ -90,28 +90,16 @@ pub trait VerifierChannel<E: FieldElement> {
...
     /// # Errors
     /// Returns an error if:
-    /// - If the sequrntial hash constructed from the remainder values does not match the specified
-    ///   `commitment`.
-    fn read_remainder<const N: usize>(
-        &mut self,
-        commitment: &<<Self as VerifierChannel<E>>::Hasher as Hasher>::Digest,
-    ) -> Result<Vec<E>, VerifierError> {
+    /// - If the degree of the polynomial is greater than the maximum expected degree.
```
irakliyk, line 99: We should probably remove the "Errors" section.

fri/src/options.rs
```diff
@@ -87,28 +84,14 @@ impl FriOptions {
...
     ///
-    /// The remainder layer (the last FRI layer) is not included in the returned value.
-    ///
     /// The number of layers for a given domain size is defined by the `folding_factor` and
-    /// `max_remainder_size` settings.
+    /// `remainder_max_degree_plus_1` settings.
     pub fn num_fri_layers(&self, mut domain_size: usize) -> usize {
         let mut result = 0;
-        while domain_size > self.max_remainder_size {
+        while domain_size > (self.remainder_max_degree_plus_1 * self.blowup_factor) {
```
irakliyk, line 91: minor nit: for clarity, I'd probably write this as:
```Rust
let max_remainder_size = self.remainder_max_degree_plus_1 * self.blowup_factor;
while domain_size < max_remainder_size {
```

fri/src/prover/mod.rs
```diff
@@ -222,12 +214,15 @@ where
...
 
-    /// Creates a FriRemainder from a vector of `evaluations` representing the remainder and
-    /// commits to the latter by hashing sequentially `evaluations`.
+    /// Creates remainder polynomial in coefficient form from a vector of `evaluations` over a domain.
     fn set_remainder(&mut self, channel: &mut C, evaluations: &mut [E]) {
-        let commitment = <H as ElementHasher>::hash_elements(evaluations);
+        let inv_twiddles = fft::get_inv_twiddles(evaluations.len());
+        fft::interpolate_poly_with_offset(evaluations, &inv_twiddles, self.options.domain_offset());
+        let remainder_poly =
+            evaluations[..evaluations.len() / self.options.blowup_factor()].to_vec();
```
irakliyk, line 222: minor nit: for clarity, I'd do it like this:
```Rust
let remainder_size = evaluations.len() / self.options.blowup_factor();
let remainder_poly = evaluations[..remainder_size].to_vec();
```

irakliyk (changes requested):
Looks great! Thank you! Just a few more minor nits.

irakliyk (approved):
All looks good! Thank you!

PR #140, opened 2023-02-10
Rowmatrix for NTT

0xKanekiKen:
This PR is in continuation to #124 and addresses #93. It introduces a new row-major matrix structure in the `matrix` module` of the `prover` crate.


math/src/fft/fft_inputs.rs
```diff
@@ -69,17 +104,98 @@ pub trait FftInputs<B: StarkField> {
...
+        Self: Send,
+    {
+        let n = self.len();
+        let num_batches = rayon::current_num_threads().next_power_of_two();
+        let batch_size = n / num_batches.min(n);
+        rayon::scope(|s| {
+            for batch_idx in 0..num_batches {
+                // create another mutable reference to the slice of values to use in a new thread; this
+                // is OK because we never write the same positions in the slice from different threads
+                let values = unsafe { &mut *(&mut *self as *mut Self) };
```
vlopes11, line 131: We should do that only if the feature `concurrent` is enabled. Otherwise, it should `let values = self;` as there is no reason to do so if there is no parallelization.

Also, efore `unsafe` blocks, it's good to prefix comments with a `// Safety: ___` as in the convention

Fianlly, one alternative way to write that, that would be slightly more readable, is:

```rust
#[cfg(feature = "concurrent")]
// Safety: we bypass the lifetime check in order to allow parallelization as the
// algorithm is guaranteed to not have overlapping mutations.
let values = unsafe { (self as *mutSelf).as_mut().unwrap_unchecked() };

#[cfg(not(feature = "concurrent"))]
let values = self;
```

math/src/fft/fft_inputs.rs
```diff
@@ -69,17 +104,98 @@ pub trait FftInputs<B: StarkField> {
...
+            }
+        });
+    }
+
+    // SPLIT-RADIX FFT
+    // ================================================================================================
+
+    #[cfg(feature = "concurrent")]
+    /// In-place recursive FFT with permuted output.
+    /// Adapted from: https://github.com/0xProject/OpenZKP/tree/master/algebra/primefield/src/fft
```
vlopes11, line 151: We might want to use a permalink reference, instead. If we always point to `master`, this link can eventually be broken. Also, preferably, it will be good to have a permalink pointing to a source code lines range, instead of a directory. This way, the reference will be unambiguous.

math/src/fft/fft_inputs.rs
```diff
@@ -172,3 +335,65 @@ pub(super) fn fft_in_place<B, I>(
...
+// TRANSPOSING
+// ================================================================================================
+
+#[cfg(feature = "concurrent")]
+fn transpose_square_stretch<E, I>(matrix: &mut I, size: usize, stretch: usize)
+where
+    E: FieldElement,
+    I: FftInputs<E> + ?Sized,
+{
+    assert_eq!(matrix.len(), size * size * stretch);
```
vlopes11, line 348: Shouldn't this be `debug_assert_eq`? If we really need to assert that here, then we probably should add a comment on why this is a production panic, and extend this comment to the public API that invokes this function.

math/src/fft/fft_inputs.rs
```diff
@@ -172,3 +335,65 @@ pub(super) fn fft_in_place<B, I>(
...
+    match stretch {
+        1 => transpose_square_1(matrix, size),
+        2 => transpose_square_2(matrix, size),
+        _ => unimplemented!("only stretch sizes 1 and 2 are supported"),
```
vlopes11, line 352: Is this a tech-debt? If yes, probably should come with a tracking issue. If no, then probably in `unreachable` instead of `unimplemented`

math/src/fft/fft_inputs.rs
```diff
@@ -172,3 +335,65 @@ pub(super) fn fft_in_place<B, I>(
...
+
+#[cfg(feature = "concurrent")]
+fn transpose_square_1<E, I>(matrix: &mut I, size: usize)
+where
+    E: FieldElement,
+    I: FftInputs<E> + ?Sized,
+{
+    debug_assert_eq!(matrix.len(), size * size);
+    if size % 2 != 0 {
+        unimplemented!("odd sizes are not supported");
```
vlopes11, line 364: Ditto; tech-debt or `unreachable`?

0xKanekiKen:
Thanks @vlopes11 for the review! I've sent another PR which only supports serial based execution as it might be too big to review at one go. Please see #142. Sorry for the trouble.

PR #142, opened 2023-02-12
Rowmatrix for NTT

0xKanekiKen:
This Pull Request is a continuation of #124 and partially addresses #93. It introduces a new matrix structure in the `matrix` module of the `prover` crate, which uses a row-major layout.

The following improvements have been made in this PR:

-  Introduction of a [Segment](https://github.com/0xKanekiKen/winterfell/blob/new-rowmatrix/prover/src/matrix/row_matrix.rs#L121-L127) struct that is utilized in processing a `column-major` matrix containing polynomial coefficients. to be evaluated on a larger domain. See [from_polys](https://github.com/0xKanekiKen/winterfell/blob/new-rowmatrix/prover/src/matrix/row_matrix.rs#L380-L428).
-  The new structure provides improved efficiency in NTT computation. The [RowMatrix](https://github.com/0xKanekiKen/winterfell/blob/new-rowmatrix/prover/src/matrix/row_matrix.rs#L29-L33) implementation had decreased performance due to the increased cache miss rate during butterfly operations. After evaluation, the [Segments](https://github.com/0xKanekiKen/winterfell/blob/new-rowmatrix/prover/src/matrix/row_matrix.rs#L362-L367) are [transposed](https://github.com/0xKanekiKen/winterfell/blob/new-rowmatrix/prover/src/matrix/row_matrix.rs#L431-L451) back to a `RowMatrix`.
- Currently, this PR only supports `serial-based execution` (support for `concurrent based execution` will be add later in subsequent PRs) and provides a performance improvement of approximately `33%` compared to the column-based approach. The benchmark was performed on a `M1 Macbook Pro`.

The benchmark results, as depicted in the accompanying snapshot, compare the performance of `matrix_evaluate_columns` (representing the `column-major` matrix) against that of `matrix_evaluate_matrix` (representing the `row-major` matrix). 
Screenshot 2023-02-12 at 4 13 51 PM

### DIFF
--- a/prover/benches/row_matrix.rs
+++ b/prover/benches/row_matrix.rs
@@ -13,8 +13,8 @@ use winter_prover::{Matrix, RowMatrix, StarkDomain};
 // ================================================================================================
 
 const SIZE: usize = 524_288;
-const BLOWUP_FACTOR: usize = 8;
-const NUM_POLYS: [usize; 4] = [16, 32, 64, 96];
+const BLOWUP_FACTOR: [usize; 3] = [2, 4, 8];
+const NUM_POLYS: [usize; 3] = [32, 64, 96];
 
 fn evaluate_columns(c: &mut Criterion) {
     let mut group = c.benchmark_group("matrix_evaluate_columns");
@@ -24,17 +24,20 @@ fn evaluate_columns(c: &mut Criterion) {
     for &num_poly in NUM_POLYS.iter() {
         let columns: Vec<Vec<BaseElement>> = (0..num_poly).map(|_| rand_vector(SIZE)).collect();
         let column_matrix = Matrix::new(columns);
-        group.bench_function(BenchmarkId::new("with_offset", num_poly), |bench| {
-            bench.iter_with_large_drop(|| {
-                let twiddles = fft::get_twiddles::<BaseElement>(SIZE);
-                let stark_domain = StarkDomain::from_custom_inputs(
-                    twiddles,
-                    BLOWUP_FACTOR,
-                    BaseElement::GENERATOR,
-                );
-                column_matrix.evaluate_columns_over(&stark_domain)
+        for &blowup_factor in BLOWUP_FACTOR.iter() {
+            let params = BenchmarkParams {
+                num_poly,
+                blowup_factor,
+            };
+            group.bench_function(BenchmarkId::new(SIZE.to_string(), params), |bench| {
+                bench.iter_with_large_drop(|| {
+                    let twiddles = fft::get_twiddles::<BaseElement>(SIZE);
+                    let stark_domain =
+                        StarkDomain::from_twiddles(twiddles, blowup_factor, BaseElement::GENERATOR);
+                    column_matrix.evaluate_columns_over(&stark_domain)
+                });
             });
-        });
+        }
     }
     group.finish();
 }
@@ -47,14 +50,36 @@ fn evaluate_matrix(c: &mut Criterion) {
     for &num_poly in NUM_POLYS.iter() {
         let columns: Vec<Vec<BaseElement>> = (0..num_poly).map(|_| rand_vector(SIZE)).collect();
         let column_matrix = Matrix::new(columns);
-        group.bench_function(BenchmarkId::new("with_offset", num_poly), |bench| {
-            bench.iter_with_large_drop(|| {
-                RowMatrix::from_polys(&column_matrix, BLOWUP_FACTOR);
+        for &blowup_factor in BLOWUP_FACTOR.iter() {
+            let params = BenchmarkParams {
+                num_poly,
+                blowup_factor,
+            };
+            group.bench_function(BenchmarkId::new(SIZE.to_string(), params), |bench| {
+                bench.iter_with_large_drop(|| {
+                    RowMatrix::transpose_and_extend(&column_matrix, blowup_factor);
+                });
             });
-        });
+        }
     }
     group.finish();
 }
 
-criterion_group!(matrix_group, evaluate_matrix, evaluate_columns,);
+/// Benchmark parameters.
+struct BenchmarkParams {
+    num_poly: usize,
+    blowup_factor: usize,
+}
+
+impl std::fmt::Display for BenchmarkParams {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "num_poly: {}, blowup_factor: {}",
+            self.num_poly, self.blowup_factor
+        )
+    }
+}
+
+criterion_group!(matrix_group, evaluate_columns, evaluate_matrix);
 criterion_main!(matrix_group);
--- a/prover/benches/row_matrix.rs
+++ b/prover/benches/row_matrix.rs
@@ -1,0 +1,60 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use math::{fft, fields::f64::BaseElement, StarkField};
+use rand_utils::rand_vector;
+use std::time::Duration;
+use winter_prover::{Matrix, RowMatrix, StarkDomain};
+
+// CONSTANTS
+// ================================================================================================
+
+const SIZE: usize = 524_288;
+const BLOWUP_FACTOR: usize = 8;
+const NUM_POLYS: [usize; 4] = [16, 32, 64, 96];
+
+fn evaluate_columns(c: &mut Criterion) {
+    let mut group = c.benchmark_group("matrix_evaluate_columns");
+    group.sample_size(10);
+    group.measurement_time(Duration::from_secs(10));
+
+    for &num_poly in NUM_POLYS.iter() {
+        let columns: Vec<Vec<BaseElement>> = (0..num_poly).map(|_| rand_vector(SIZE)).collect();
+        let column_matrix = Matrix::new(columns);
+        group.bench_function(BenchmarkId::new("with_offset", num_poly), |bench| {
+            bench.iter_with_large_drop(|| {
+                let twiddles = fft::get_twiddles::<BaseElement>(SIZE);
+                let stark_domain = StarkDomain::from_custom_inputs(
+                    twiddles,
+                    BLOWUP_FACTOR,
+                    BaseElement::GENERATOR,
+                );
+                column_matrix.evaluate_columns_over(&stark_domain)
+            });
+        });
+    }
+    group.finish();
+}
+
+fn evaluate_matrix(c: &mut Criterion) {
+    let mut group = c.benchmark_group("matrix_evaluate_matrix");
+    group.sample_size(10);
+    group.measurement_time(Duration::from_secs(10));
+
+    for &num_poly in NUM_POLYS.iter() {
+        let columns: Vec<Vec<BaseElement>> = (0..num_poly).map(|_| rand_vector(SIZE)).collect();
+        let column_matrix = Matrix::new(columns);
+        group.bench_function(BenchmarkId::new("with_offset", num_poly), |bench| {
+            bench.iter_with_large_drop(|| {
+                RowMatrix::from_polys(&column_matrix, BLOWUP_FACTOR);
+            });
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(matrix_group, evaluate_matrix, evaluate_columns,);
+criterion_main!(matrix_group);
--- a/prover/src/domain.rs
+++ b/prover/src/domain.rs
@@ -52,19 +52,25 @@ impl<B: StarkField> StarkDomain<B> {
     }
 
     /// Returns a new STARK domain initialized with the provided custom inputs.
-    pub fn from_custom_inputs(
-        trace_twiddles: Vec<B>,
-        blowup_factor: usize,
-        domain_offset: B,
-    ) -> Self {
-        let ce_domain_size = trace_twiddles.len() * 2;
+    pub fn from_twiddles(trace_twiddles: Vec<B>, blowup_factor: usize, domain_offset: B) -> Self {
+        // both `trace_twiddles` length and `blowup_factor` must be a power of two.
+        assert!(
+            trace_twiddles.len().is_power_of_two(),
+            "the length of trace twiddles must be a power of 2"
+        );
+        assert!(
+            blowup_factor.is_power_of_two(),
+            "blowup factor must be a power of 2"
+        );
+
+        let ce_domain_size = trace_twiddles.len() * blowup_factor * 2;
         let domain_gen = B::get_root_of_unity(log2(ce_domain_size));
         let ce_domain = get_power_series(domain_gen, ce_domain_size);
 
         StarkDomain {
             trace_twiddles,
             ce_domain,
-            ce_to_lde_blowup: blowup_factor,
+            ce_to_lde_blowup: 1,
             ce_domain_mod_mask: ce_domain_size - 1,
             domain_offset,
         }
--- a/prover/src/domain.rs
+++ b/prover/src/domain.rs
@@ -51,6 +51,25 @@ impl<B: StarkField> StarkDomain<B> {
         }
     }
 
+    /// Returns a new STARK domain initialized with the provided custom inputs.
+    pub fn from_custom_inputs(
+        trace_twiddles: Vec<B>,
+        blowup_factor: usize,
+        domain_offset: B,
+    ) -> Self {
+        let ce_domain_size = trace_twiddles.len() * 2;
+        let domain_gen = B::get_root_of_unity(log2(ce_domain_size));
+        let ce_domain = get_power_series(domain_gen, ce_domain_size);
+
+        StarkDomain {
+            trace_twiddles,
+            ce_domain,
+            ce_to_lde_blowup: blowup_factor,
+            ce_domain_mod_mask: ce_domain_size - 1,
+            domain_offset,
+        }
+    }
+
     // EXECUTION TRACE
     // --------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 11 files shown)


irakliyk (changes requested):
Very interesting! Thank you! This is not a full review, just a few initial thoughts about the overall structure.

1. I would put the `Matrix` struct into its own module - e.g., `/src/matrix/col_matrix.rs`.
2. I would also move `Segment` and `Segments` structs into their own module - e.g., `/src/matrix/segment.rs`.
3. I would add a `from_polys()` constructor to the `RowMatrix` struct and make segment structs internal to the `matrix` module (i.e., they shouldn't be exposed publicly from the crate).

The last point also means that we should probably benchmark this `from_polys()` method, and I'd compare it against `Matrix::evaluate_columns_over()` method (for simplicity, we could create another constructor for `StarkDomain` struct).

0xKanekiKen:
Thanks @irakliyk for the review. I followed your suggestions and reorganized the matrix module.

I conducted a benchmark on a `M1 mini` and obtained the following results:

Screenshot from 2023-02-13 13-46-13

prover/src/matrix/segments.rs
```diff
@@ -0,0 +1,495 @@
...
+// ================================================================================================
+
+/// Represents a collection of Segment objects.
+#[derive(Debug, Clone)]
+pub struct Segments<E>
+where
+    E: FieldElement,
+{
+    matrix: Vec<Segment<E>>,
+}
```
irakliyk, line 289: Why do we need this `Segments` struct? From what I can tell, it is used only for implementing the iterators, but then the iterators don't seem to do anything special (unless I missed something). Could we just use `Vec<Segment>` instead of creating a special struct?

prover/src/matrix/segments.rs
```diff
@@ -0,0 +1,495 @@
...
+            self.data[row_idx][4] *= offset;
+
+            // apply on index 5.
+            self.data[row_idx][5] *= offset;
+
+            // apply on index 6.
+            self.data[row_idx][6] *= offset;
+
+            // apply on index 7.
+            self.data[row_idx][7] *= offset;
```
irakliyk, line 242: nit: I'd probably get rid of the comments and blank lines here: I think it is fairly obvious what's going on.

prover/src/matrix/segments.rs
```diff
@@ -0,0 +1,495 @@
...
```
irakliyk, line 274: nit: I'd probably get rid of the comments and blank lines here: I think it is fairly obvious what's going on.

prover/src/matrix/segments.rs
```diff
@@ -0,0 +1,495 @@
...
+    // CONSTRUCTOR
+    // --------------------------------------------------------------------------------------------
+
+    /// Creates a new segment of a row-major matrix from the specified data.
+    pub fn new(data: Vec<[E; ARR_SIZE]>) -> Self {
+        Self { data }
+    }
+
+    // PUBLIC ACCESSORS
+    // --------------------------------------------------------------------------------------------
```
irakliyk, line 63: Do we actually need all these accessors here? It seems like the only ones we really need are `as_data()` and `evaluate_poly()`.

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,155 @@
...
+        offsets.push(E::BaseField::ONE);
+        for i in 1..num_rows {
+            offsets.push(offsets[i - 1] * domain_offset);
+        }
+
+        // create segments.
+        (0..num_of_segments).for_each(|i| {
+            // create a vector of arrays to hold the result.
+            let mut result_vec_of_arrays =
+                unsafe { uninit_vector::<[E; ARR_SIZE]>(num_rows * blowup_factor) };
```
irakliyk, line 77: I wonder if it would be more efficient to create a single vector of size `num_rows * row_width * blowup_factor` and then break it into individual vectors via [from_raw_parts()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts).

Either way, I'd probably encapsulate this memory allocation logic into a stand-alone function which is run before we enter this loop. The function could have the following signature:
```Rust
fn allocate_segments(num_rows: usize, row_width: usize, blowup_factor: usize) -> Vec<Vec<[E; ARR_SIZE]>>
```

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,155 @@
...
+            (i * ARR_SIZE..(i + 1) * ARR_SIZE).for_each(|col_idx| {
+                let row = polys.get_column(col_idx);
+                row.iter().enumerate().for_each(|(row_idx, elem)| {
+                    result_vec_of_arrays[row_idx][col_idx - i * ARR_SIZE] =
+                        elem.mul_base(offsets[row_idx]);
+                });
+            });
+
+            // create a `Segment` object from the result.
+            let row_matrix = Segment::new(result_vec_of_arrays);
```
irakliyk, line 90: I think part of the comment is wrong here (i.e., no polynomial are evaluated). Also, I'd probably encapsulate this logic into a stand-alone function. Something like this:
```Rust
fn prepare_segment(segment: &mut Vec<[E; ARR_SIZE]>, offsets: Vec<E::BaseElement>) -> Segment
```

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,155 @@
...
+            (segment.as_data()).iter().enumerate().for_each(|(j, row)| {
+                result[j * row_width / ARR_SIZE + i] = *row;
+            })
+        });
+
+        // create a `RowMatrix` object from the result.
+        RowMatrix {
+            data: flatten_vector_elements(result),
+            row_len: row_width,
+        }
```
irakliyk, line 115: I would encapsulate this into a stand-alone function.

Also, a comment for the future: I wonder if we could improve it by using some kind of in-place transposition algorithm to avoid additional memory allocation.

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,155 @@
...
+        // get the number of segments in the matrix.
+        let num_of_segments = row_width / ARR_SIZE;
+        let mut segments = Segments::new(Vec::new());
+
+        // precompute offsets for each row.
+        let mut offsets = Vec::with_capacity(num_rows);
+        offsets.push(E::BaseField::ONE);
+        for i in 1..num_rows {
+            offsets.push(offsets[i - 1] * domain_offset);
+        }
```
irakliyk, line 71: I would encapsulate this into a stand alone function.

prover/src/domain.rs
```diff
@@ -51,6 +51,25 @@ impl<B: StarkField> StarkDomain<B> {
         }
     }
 
+    /// Returns a new STARK domain initialized with the provided custom inputs.
+    pub fn from_custom_inputs(
```
irakliyk, line 55: nit: I'd rename this to `from_twiddles()`.

prover/src/matrix/tests.rs
```diff
@@ -0,0 +1,67 @@
...
+    // evaluate columns using the using the polynomial evaluation implementation.
+    let offset = BaseElement::GENERATOR;
+    let domain = build_domain(n * blowup_factor);
+    let shifted_domain = domain.iter().map(|&x| x * offset).collect::<Vec<_>>();
+    for p in columns.iter_mut() {
+        *p = polynom::eval_many(p, &shifted_domain);
+    }
+
+    // transpose the columns back to a row major format.
+    let eval_col = transpose(columns);
```
irakliyk, line 35: Instead of doing a transpose, could we compare `row_matrix` and `columns` row-by-row? We can just implement a simple `get_row()` function and use it in a loop. The function could look like this:
```Rust
fn get_row(columns: &Vec<Vec<BaseElement>>, row_id: usize) -> Vec<BaseElement>
```

irakliyk (changes requested):
Thank you! Looks good! I left a few comments inline - most of them are about removing potentially un-needed code and also encapsulating some code in stand-alone functions (I think this will help when we need to provide concurrent implementations of these functions).

prover/src/matrix/row_matrix.rs
0xKanekiKen: After considering the problem, I believe that transposing the data in place is not a straightforward task. Consider the desired final layout of the data to be in the format(location of array in the final vector) `vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]`, but we receive `vec![1, 5, 9, 2, 6, 10, 3, 7, 11, 4, 8, 12]` from the current Segment's approach.

I don't at the moment have a solution that allows us to directly achieve the final transposed state in just one pass without allocating new memory. I'll think about it but for the time being I'm sending an update using the current approach.

prover/src/matrix/segments.rs
0xKanekiKen: Yeah- we don't need it in the current implementation. I've removed it in the latest update.

prover/src/matrix/segments.rs
0xKanekiKen: done. I've removed all the unnecessary methods from the struct implementation.

prover/src/matrix/row_matrix.rs
0xKanekiKen: I think it get bit easier if we just return a Vec of `Segment` instead of `unallocated_vector`. We are using the output in `prepare_segment` which should return a `Segment`. Given this vector is not made up of  `Segment`, we might not be able to update the input vector.

prover/src/matrix/row_matrix.rs
0xKanekiKen: I've added a [TODO](https://github.com/facebook/winterfell/blob/934f0db10a959f08490f9d8cb09281a5fbe64425/prover/src/matrix/row_matrix.rs#L90).

prover/src/matrix/tests.rs
```diff
@@ -0,0 +1,53 @@
...
+    // evaluate columns using the using the polynomial evaluation implementation.
+    let offset = BaseElement::GENERATOR;
+    let domain = build_domain(n * blowup_factor);
+    let shifted_domain = domain.iter().map(|&x| x * offset).collect::<Vec<_>>();
+    for p in columns.iter_mut() {
+        *p = polynom::eval_many(p, &shifted_domain);
+    }
+
+    // compare the results of the two implementations row by row.
+    for row in 0..n {
```
irakliyk, line 34: This should probably be in `0..n * blowup_factor`.

prover/src/domain.rs
```diff
@@ -51,6 +51,21 @@ impl<B: StarkField> StarkDomain<B> {
         }
     }
 
+    /// Returns a new STARK domain initialized with the provided custom inputs.
+    pub fn from_twiddles(trace_twiddles: Vec<B>, blowup_factor: usize, domain_offset: B) -> Self {
```
irakliyk, line 55: I know we use this only for tests, but I think it might be a good idea to add a couple of assertions to check that `trace_twiddles.len()` and `blowup_factor` are both powers of 2.

prover/benches/row_matrix.rs
```diff
@@ -0,0 +1,57 @@
...
+use rand_utils::rand_vector;
+use std::time::Duration;
+use winter_prover::{Matrix, RowMatrix, StarkDomain};
+
+// CONSTANTS
+// ================================================================================================
+
+const SIZE: usize = 524_288;
+const BLOWUP_FACTOR: usize = 8;
+const NUM_POLYS: [usize; 4] = [16, 32, 64, 96];
```
irakliyk, line 17: We can probably get rid of the case for 16 polys (just keep 32, 64, and 96) - but I think it would be cool benchmark a different blowup factor too. Specifically, blowup factor of 2.

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,200 @@
...
+    segment: &mut Segment<E>,
+    polys: &Matrix<E>,
+    seg_idx: usize,
+    offsets: &[E::BaseField],
+) where
+    E: FieldElement,
+{
+    (seg_idx * ARR_SIZE..(seg_idx + 1) * ARR_SIZE).for_each(|col_idx| {
+        // get the column from the polys matrix.
+        let row = polys.get_column(col_idx);
```
irakliyk, line 194: Is the variable naming here correct? Should it be `column` rather than `row`?

irakliyk (changes requested):
Looks good! Thank you! Just a couple more minor comments inline.

prover/src/matrix/tests.rs
0xKanekiKen: thanks for pointing

prover/src/matrix/row_matrix.rs
0xKanekiKen: yes, thanks

prover/src/domain.rs
```diff
@@ -51,6 +51,31 @@ impl<B: StarkField> StarkDomain<B> {
...
+        let domain_gen = B::get_root_of_unity(log2(ce_domain_size));
+        let ce_domain = get_power_series(domain_gen, ce_domain_size);
+
+        StarkDomain {
+            trace_twiddles,
+            ce_domain,
+            ce_to_lde_blowup: blowup_factor,
+            ce_domain_mod_mask: ce_domain_size - 1,
+            domain_offset,
+        }
```
irakliyk, line 76: Sorry, one more thing: I think `ce_domain_size` should be `trace_twiddles.len() * blowup_factor * 2` and then `ce_to_lde_blowup` should be just `1`.

0xKanekiKen:
After running the benchmark for different blowup factors, the results are as follows:

Screenshot from 2023-02-16 05-55-40
Screenshot from 2023-02-16 06-11-51

prover/src/domain.rs
0xKanekiKen: Thanks, I've updated it [here](https://github.com/facebook/winterfell/blob/4ed1c7ed9d6c3e52e78a40e96fc2ec99ab71d996/prover/src/domain.rs#L66-L73).

prover/benches/row_matrix.rs
```diff
@@ -0,0 +1,86 @@
...
+    group.finish();
+}
+
+/// Benchmark parameters.
+struct BenchmarkParams {
+    num_poly: usize,
+    blowup_factor: usize,
+}
+
+/// Implements Display trait for BenchmarkParams.
```
Nashtare, line 74: nit: not sure this description is useful.

prover/src/domain.rs
```diff
@@ -51,6 +51,31 @@ impl<B: StarkField> StarkDomain<B> {
         }
     }
 
+    /// Returns a new STARK domain initialized with the provided custom inputs.
+    pub fn from_twiddles(trace_twiddles: Vec<B>, blowup_factor: usize, domain_offset: B) -> Self {
+        // both trace twiddles and blowup factor must be a power of two.
```
Nashtare, line 56: nit: 
```suggestion
        // both `trace_twiddles` length and `blowup_factor` must be a power of two.
```

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,200 @@
...
+// ROWMAJOR MATRIX
+// ================================================================================================
+
+/// A row-major matrix of field elements. The matrix is represented as a single vector of field
+/// elements, where the first `row_len` elements represent the first row of the matrix, the next
+/// `row_len` elements represent the second row, and so on.
+///
+/// # Note
+/// - The number of rows in the matrix is always a multiple of the ARR_SIZE.
+/// - The number of columns in the matrix is always a multiple of the ARR_SIZE.
```
Nashtare, line 21: nit, I'd remove "the"
```suggestion
/// - The number of rows in the matrix is always a multiple of ARR_SIZE.
/// - The number of columns in the matrix is always a multiple of ARR_SIZE.
```

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,200 @@
...
+    // CONSTRUCTORS
+    // --------------------------------------------------------------------------------------------
+
+    /// Creates a new row-major matrix from the specified data and row length. The data must be
+    /// arranged in row-major order, i.e. the first `row_len` elements of the data represent the first
+    /// row of the matrix, the next `row_len` elements represent the second row, and so on.
+    ///
+    /// # Panics
+    /// - if the number of elements in the data is not a multiple of the specified row length;
+    /// - if the specified row length is not a multiple of the ARR_SIZE;
```
Nashtare, line 41: nit
```suggestion
    /// - if the specified row length is not a multiple of ARR_SIZE;
```

prover/src/matrix/row_matrix.rs
```diff
@@ -0,0 +1,200 @@
...
+        assert!(data.len() % row_len == 0);
+        assert!(row_len % ARR_SIZE == 0);
+        assert!(!data.is_empty());
+
+        Self { data, row_len }
+    }
+
+    /// Converts a column-major matrix of polynomials `Matrix<E>` into a RowMatrix evaluated at
+    /// a shifted domain offset.
+    pub fn from_polys(polys: &Matrix<E>, blowup_factor: usize) -> Self {
```
Nashtare, line 53: Although I get the idea, I find the naming / description slightly off point / confusing. But I'm not sure what would be really best, so maybe we can discard my comment

prover/benches/row_matrix.rs
0xKanekiKen: I've removed it in the latest commit.

prover/src/matrix/row_matrix.rs
0xKanekiKen: I was thinking of using `from_col_matrix` but find the current one better.

Nashtare: This basically boils down to a matrix transposition followed by an LDE extension, right? Maybe we could try something reflecting this, like `pub fn transpose_and_extend()`?

0xKanekiKen: done! I've updated it here. Please see [this](https://github.com/facebook/winterfell/blob/19be088ab4849cc34de3db0a1821db5364d30164/prover/src/matrix/row_matrix.rs#L53)

Nashtare: Thanks!

irakliyk (approved):
All looks good! Thank you!

PR #146, opened 2023-02-17
Fix as_int() non-canonical output

### LINKED ISSUE #145: Bug in f64 deserialization
Unless I should be using another method for serialization, deserialization of zero values in this field does not work due to their canonical representation. The code below will panic with the error message: `InvalidValue("invalid field element: value 18446744069414584321 is greater than or equal to the field modulus")'`

```rust
use winter_math::{fields::f64::BaseElement, StarkField};
use winter_utils::Deserializable;
use winter_utils::SliceReader;

let bytes = BaseElement::from(0u8).as_int().to_le_bytes();
let mut reader = SliceReader::new(&bytes);
BaseElement::read_batch_from(&mut reader, 1).unwrap();
```

The issue is the following line, which should be changed to a greater than inequality:
https://github.com/facebook/winterfell/blob/cd76df242fcec8bab45efcdabf366fb0d8c76fbd/math/src/field/f64/mod.rs#L584

Nashtare:
This change fixes the non-canonical conversion of 0 from its Montgomery form equivalent in `f64.`

closes #145

### DIFF
--- a/math/src/field/f64/mod.rs
+++ b/math/src/field/f64/mod.rs
@@ -245,9 +245,11 @@ impl StarkField for BaseElement {
     #[inline]
     fn as_int(&self) -> Self::PositiveInteger {
         let x = self.0;
-        let (r, c) = x.overflowing_add(x << 32);
-        let res = r.wrapping_sub(r >> 32).wrapping_sub(c as u64);
-        M - res
+        let (a, e) = x.overflowing_add(x << 32);
+        let b = a.wrapping_sub(a >> 32).wrapping_sub(e as u64);
+
+        let (r, c) = 0u64.overflowing_sub(b);
+        r.wrapping_sub(0u32.wrapping_sub(c as u32) as u64)
     }
 }
 
--- a/math/src/field/f64/tests.rs
+++ b/math/src/field/f64/tests.rs
@@ -125,6 +125,11 @@ fn element_as_int() {
     let v = u64::MAX;
     let e = BaseElement::new(v);
     assert_eq!(v % super::M, e.as_int());
+
+    let e1 = BaseElement::new(0);
+    let e2 = BaseElement::new(M);
+    assert_eq!(e1.as_int(), e2.as_int());
+    assert_eq!(e1.as_int(), 0);
 }
 
 #[test]


irakliyk:
Thank you! This a very good fix! A couple of questions:

1. Should we add a test which would have failed before but is not failing now?
2. We used to have this function simply call `mont_red_cst()` (see below) - did that have a similar problem?

```Rust
fn as_int(&self) -> Self::PositiveInteger {
    mont_red_cst(self.0 as u128)
}
```
(I can also check myself - but probably tomorrow).

Nashtare:
I'll add a test later for the edge case, and make sure `mont_red_cst()` doesn't have the same problem!

Nashtare:
@irakliy I've added an edge-case test in the existing as_int() test. Calling `mont_red_cst()` on either `0u128, (M as u128)` or other multiples does return the correct value. There is actually a slight difference in the way reduction is handled now, which caused this edge case.

irakliyk:
@Nashtare - thank you! I'm wondering if we should just go back to the way we had it before:
```Rust
fn as_int(&self) -> Self::PositiveInteger {
    mont_red_cst(self.0 as u128)
}
```
Though, seems like even with the additions in this PR, the new way should still be slightly more efficient:
```Rust
fn as_int(&self) -> Self::PositiveInteger {
     let x = self.0;
     let (r, c) = x.overflowing_add(x << 32);
     let res = r.wrapping_sub(r >> 32).wrapping_sub(c as u64);
     let res = M - res;

     // Constant-time branching, handling 0 case.
     let mask = (self.0 == 0) as u64;
     res ^ (mask & res)
}
```
What do you think?

Nashtare:
I think we could stick with it still, as you pointed out, it's slightly faster than the ct version of Montgomery reduction, and doesn't really complicate the implementation.

math/src/field/f64/tests.rs
```diff
@@ -125,6 +125,10 @@ fn element_as_int() {
     let v = u64::MAX;
     let e = BaseElement::new(v);
     assert_eq!(v % super::M, e.as_int());
+
+    let e1 = BaseElement::new(0);
+    let e2 = BaseElement::new(M);
+    assert_eq!(e1.as_int(), e2.as_int());
```
irakliyk, line 131: Should we also add the following assertion?:
```Rust
assert_eq!(e1.as_int(), 0);
```
I think with the old code both `0` and `M` were getting reduced to `18446744069414584321` - so, I think we may need an extra check here.

Nashtare: Very good point. It actually made me realize I had messed up the constant time branching, as I wasn't setting the binary mask to all 1s when the equality was satisfied. I've fixed it now, should be all good.

math/src/field/f64/mod.rs
```diff
@@ -247,7 +247,11 @@ impl StarkField for BaseElement {
         let x = self.0;
         let (r, c) = x.overflowing_add(x << 32);
         let res = r.wrapping_sub(r >> 32).wrapping_sub(c as u64);
-        M - res
+        let res = M - res;
+
+        // Constant-time branching, handling 0 case.
+        let mask = -((self.0 == 0) as i64) as u64;
+        res ^ (mask & res)
```
irakliyk, line 254: This works! But I'm wondering if this is better than:
```Rust
let x = self.0;
let (a, e) = x.overflowing_add(x << 32);
let b = a.wrapping_sub(a >> 32).wrapping_sub(e as u64);

let (r, c) = 0u64.overflowing_sub(b);
r.wrapping_sub(0u32.wrapping_sub(c as u32) as u64)
```

On [godbolt](https://godbolt.org/) the above seems to give a very marginally better result:

Screenshot 2023-02-18 at 12 43 06 PM

vs.

Screenshot 2023-02-18 at 12 44 14 PM

irakliyk: Yeah - let's do it. Thank you!

Nashtare: Fixed, sorry about that!

irakliyk (approved):
All looks good! Thank you again for fixing this!

PR #148, opened 2023-02-19
Prepare v0.5 release

irakliyk:
This PR prepares the crates for v0.5 release. Specifically:

- Updates versions of all crates to v0.5.
- Updates versions of outdated dependences (there were a couple).
- Updates minimum required Rust version to v1.67.
- Updates the CHANGELOG.

### DIFF
--- a/air/Cargo.toml
+++ b/air/Cargo.toml
@@ -1,16 +1,16 @@
 [package]
 name = "winter-air"
-version = "0.4.2"
+version = "0.5.0"
 description = "AIR components for the Winterfell STARK prover/verifier"
 authors = ["winterfell contributors"]
 readme = "README.md"
 license = "MIT"
 repository = "https://github.com/novifinancial/winterfell"
-documentation = "https://docs.rs/winter-air/0.4.2"
+documentation = "https://docs.rs/winter-air/0.5.0"
 categories = ["cryptography", "no-std"]
 keywords = ["crypto", "arithmetization", "air"]
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.67"
 
 [lib]
 bench = false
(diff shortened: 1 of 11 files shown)


irakliyk:
@Nashtare, @Al-Kindi-0 - would appreciate a review. Thank you!

air/Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
 readme = "README.md"
 license = "MIT"
 repository = "https://github.com/novifinancial/winterfell"
-documentation = "https://docs.rs/winter-air/0.4.2"
+documentation = "https://docs.rs/winter-air/0.5.0"
 categories = ["cryptography", "no-std"]
 keywords = ["crypto", "arithmetization", "air"]
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.67"
```
Nashtare, line 13: Curious about the toolchain change directly up to 1.67. I assumed there was a breaking change but I'm on 1.66 locally and everything runs smoothly.

Nashtare (commented):
LGTM, modulo waiting for #146 and #150 as mentioned in the CHANGELOG.
Also have a tiny question about the need for raising MSRV to 1.67.

air/Cargo.toml
irakliyk: There was no real need to push it that far (and I think 1.60 would have also worked fine). I do notice that there are minor performance improvements with each new version of Rust - so, didn't want it to fall too far behind. I can change it to 1.66 if that works better though.

Nashtare: No need to downgrade it to 1.66 🙂 I was just curious about the rationale behind the bump.

PR #149, opened 2023-02-19
`TraceTable::fill()`: use least restrictive closure type for arguments

plafer:
Currently, `TraceTable::fill()` uses `Fn` as the closure type for both its closure arguments. This needlessly restricts the possibilities for the caller. This PR uses the least restrictive closure type for each argument. 

For example, the following is now made possible:

```rust
pub fn build_do_work_trace(start: BaseElement, n: usize) -> TraceTable<BaseElement> {
    let trace_width = 1;
    let mut trace = TraceTable::new(trace_width, n);

    let mut gen = Generator{};

    trace.fill(
        |state| {
            state[0] = start;
        },
        |_, state| {
            // here, `gen` is borrowed mutably by the closure,
            // which is now allowed.
            state[0] = state[0].exp(3u32.into()) + gen.next();
        },
    );

    trace
}

struct Generator;

impl Generator {
    fn next(&mut self) -> BaseElement { ... }
}
```

### DIFF
--- a/prover/src/trace/trace_table.rs
+++ b/prover/src/trace/trace_table.rs
@@ -227,10 +227,10 @@ impl<B: StarkField> TraceTable<B> {
     ///   - index of the last updated row (starting with 0).
     ///   - a mutable reference to the last updated state; the contents of the state are copied
     ///     into the next row of the trace after the closure returns.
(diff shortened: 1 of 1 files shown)


irakliyk (approved):
Looks good! Thank you!

PR #150, opened 2023-02-19
Serialization trait improvements

irakliyk:
This PR refactors serialization/deserialization traits in `core-utils` crate. Specifically:

- Moves the related code from `lib.rs` to separate models.
- Adds a few convenience methods and reduces code duplication in `ByteReader`.
- Renames a few methods to be more consistent with the overall naming convention.


irakliyk:
@Nashtare, @Al-Kindi-0 - would appreciate a review. Thank you!

Nashtare (commented):
LGTM

utils/core/src/serde/byte_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+    ///
+    /// # Errors
+    /// Returns a [DeserializationError] error the reader is at EOF.
+    fn read_u8(&mut self) -> Result<u8, DeserializationError>;
+
+    /// Returns the next byte to be read from `self` without advancing the reader to the next byte.
+    ///
+    /// # Errors
+    /// Returns a [DeserializationError] error the reader is at EOF.
+    fn peek_u8(&self) -> Result<u8, DeserializationError>;
```
vlopes11, line 21: optional nit: peek could instead return an `Option` as the user might just want to check if there is another byte to be read. The signature would look like:

```rust
fn peek_u8(&self) -> Result<Option<u8>, DeserializationError>;
```

Also, another method that can be helpful is
```rust
fn peek_into(&self, buf: &mut [u8]) -> Result<usize, DeserializationError>;
```

Without this method, we cannot peek more than one byte, and this might be a too restrictive

utils/core/src/serde/byte_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn read_u128(&mut self) -> Result<u128, DeserializationError> {
+        let bytes = self.read_array::<16>()?;
+        Ok(u128::from_le_bytes(bytes))
+    }
+
+    /// Reads a deserializable value from `self`.
+    ///
+    /// # Panics
+    /// Panics if the value could not be read from `self`.
+    fn read<D: Deserializable>(&mut self) -> Result<D, DeserializationError>
```
vlopes11, line 105: optional nit: it is better to keep a single style to declare generics. Either we put all the trait requirements inside the `where`, or we declare them with the generics.

Here we have both. The community currently favors `where` as a more readable way to quickly check the trait requirements, so I'd recommend this.

utils/core/src/serde/byte_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct SliceReader<'a> {
+    source: &'a [u8],
+    pos: usize,
+}
+
+impl<'a> SliceReader<'a> {
+    /// Creates a new slice reader from the specified slice.
+    pub fn new(source: &'a [u8]) -> Self {
+        SliceReader { source, pos: 0 }
+    }
```
vlopes11, line 126: One alternative is to hold only the slice that we effectively need. That way, we could get rid of `pos`, and [slice::split_at](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at) whenever we are reading new content. Example:

Suppose we need to read 8 bytes
```rust
self.check_eor(8)?;
let (read, remainder) = self.source.split_at(8);
self.source = remainder;
(... do something with read ...)
```

Or, to peek:

```rust
(n < self.source.len()).then(|| &self.source[..n])
```

Also, as a constructor, it might be better to have more flexibility as we could set as argument a vec, slice, string, etc

```rust
pub fn new<B>(source: B) -> Self where B: AsRef<[u8]>
```

utils/core/src/serde/byte_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+        let result = self.source[self.pos..self.pos + len].to_vec();
+        self.pos += len;
+        Ok(result)
+    }
+
+    fn read_array<const N: usize>(&mut self) -> Result<[u8; N], DeserializationError> {
+        self.check_eor(N)?;
+        let bytes = &self.source[self.pos..self.pos + N];
+        self.pos += N;
+        Ok(bytes.try_into().expect("slice to array conversion failed"))
```
vlopes11, line 153: Using a `try_into` might be slightly more expensive than it needs to be because the compiler will have to worry to display the expected error message: `slice to array conversion failed"`

This is in fact infallible, so we probably shouldn't add more cycles for something that will never happen (as the length of the bytes is already checked)

We could, instead

```rust
self.check_eor(N)?;
let mut buf = [0_u8; N];
buf.copy_from_slice(&self.source[self.pos..self.pos + N]);
self.pos += N;
Ok(buf)
```

utils/core/src/serde/byte_writer.rs
```diff
@@ -0,0 +1,80 @@
...
+    ///
+    /// # Panics
+    /// Panics if the byte could not be written into `self`.
+    fn write_u8(&mut self, value: u8);
+
+    /// Writes a sequence of bytes into `self`.
+    ///
+    /// # Panics
+    /// Panics if the sequence of bytes could not be written into `self`.
+    fn write_bytes(&mut self, values: &[u8]);
```
vlopes11, line 21: This is debatable, but we might want it to be fallible instead of potentially throwing a panic.

If we plan to have only `Vec` as implementor, then we don't really need to worry about that as the only possible failure is when we run out of memory. However, we might want in the future to send serializable data to I/O (either network, disk, some big table, etc). If the trait is infallible, we won't allow such options.

utils/core/src/serde/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// Returns an estimate of how many bytes are needed to represent self.
+    ///
+    /// The default implementation returns zero.
+    fn get_size_hint(&self) -> usize {
+        0
+    }
+}
+
+impl Serializable for () {
```
vlopes11, line 47: We might want to implement it for `String` and/or `&str` as well.

Also, it might be good to implement `Serializable` for `n-tuples`. Example:

```rust
impl<A> Serializable for (A) where A: Serializable
impl<A, B> Serializable for (A, B) where A: Serializable, B: Serializable
```

Usually, libraries will make that available up to 32-tuple.

If we use batch as generic iterators, this will become very powerful as we can serialize anything - from vectors to hash maps

utils/core/src/serde/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+        let mut result = Vec::with_capacity(self.get_size_hint());
+        self.write_into(&mut result);
+        result
+    }
+
+    /// Serializes all elements of the `source` and writes these bytes into the `target`.
+    ///
+    /// This method does not write any metadata (e.g. number of serialized elements) into the
+    /// `target`.
+    fn write_batch_into<W: ByteWriter>(source: &[Self], target: &mut W) {
```
vlopes11, line 33: We could make this more flexible by taking an iterator as argument instead of a slice.

```rust
fn write_batch_into<T, I, W>(source: I, target: &mut W) where
T: core::borrow::Borrow<Self>,
I: IntoIterator<Item = T>,
W: ByteWriter
```

This way, we will allow anything to be used as argument - including slices.

utils/core/src/serde/byte_reader.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn peek_u8(&self) -> Result<u8, DeserializationError>;
+
+    /// Returns a byte vector of the specified length read from `self`.
+    ///
+    /// # Errors
+    /// Returns a [DeserializationError] if a vector of the specified length could not be read
+    /// from `self`.
+    fn read_vec(&mut self, len: usize) -> Result<Vec<u8>, DeserializationError>;
+
+    /// Returns a byte array of length `N` reade from `self`.
```
Al-Kindi-0, line 30: nit: reade -> read

utils/core/src/serde/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+        for item in source {
```
vlopes11, line 34: We probably want to serialize the length of the collection before the data as it could be impossible to reconstruct the collection from a stream of bytes if we don't do that.

Example
```rust
(source.len() as u64).write_into(target);
```

It is important we cast to a specific type as `usize` might differ depending on architecture, and we want to avoid architecture-dependant serialization.

utils/core/src/serde/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// * The `source` does not contain enough bytes to deserialize the specified number of
+    ///   elements.
+    /// * Bytes read from the `source` do not represent a valid value for `Self` for any of the
+    ///   elements.
+    ///
+    /// Note: if the error occurs, the reader is not rolled back to the state prior to calling
+    /// this function.
+    fn read_batch_from<R: ByteReader>(
+        source: &mut R,
+        num_elements: usize,
```
vlopes11, line 137: Ditto; instead of asking the user for the expected elements count, we could embed it into the vector serialization, by default.

Al-Kindi-0 (commented):
Looks great! 
Just one nit from my side.

utils/core/src/serde/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> Result<Vec<Self>, DeserializationError> {
```
vlopes11, line 138: Instead of returning a vec, we might want ot return anything that implements `FromIterator`. This will give us great flexibility and allow us to transparently reconstruct complex structures such as maps, files, databases, etc

```rust
fn read_batch_from<R, T>(
    source: &mut R,
) -> Result<T, DeserializationError> where
R: ByteReader,
T: FromIterator<Self>
```

vlopes11 (commented):
Looks good overall! A couple of suggestions to further increase the flexibility

utils/core/src/serde/byte_reader.rs
irakliyk: Fixed!

utils/core/src/serde/byte_reader.rs
irakliyk: Agreed! Fixed!

utils/core/src/serde/byte_reader.rs
irakliyk: Good idea! Updated.

utils/core/src/serde/byte_reader.rs
irakliyk: This is a cool way to do it! However, I'm not sure how the compiler will handle this (i.e., if it will actually result in less efficient code) - so, without further investigation, I'll leave it as is for now. We can always improve this later as it will be non-breaking change.

Regarding a more general constructor: I tried to make it work but ran into lifetime bound issues - so, left it for later as well.

utils/core/src/serde/mod.rs
irakliyk: I am leaving this for the future as adding implementation for more types will be a non-breaking change.

utils/core/src/serde/mod.rs
irakliyk: I like this, and tried to make it work, but ran into lifetime bounds issues - so, leaving this for the future.

utils/core/src/serde/mod.rs
irakliyk: Here, we don't want to serialize the length as we don't know what are potential maximum bounds on the length or if the length can be determined in some other way.

So, the user of this method is responsible for either serialization of length or convening length info to the deserializer in some other way.

utils/core/src/serde/mod.rs
irakliyk: Similar to the above, this is a performance optimization as length could be bounded by some small number (i.e., max of 255 items) and may be stored one time for multiple collections).

utils/core/src/serde/mod.rs
irakliyk: I'm not sure how to provide a generic implementation for this - so, leaving it for the future.

utils/core/src/serde/byte_reader.rs
irakliyk: For the first one, we already have `has_more_bytes()` method. The second one would be cool - but I'm also leaving it for the future.

utils/core/src/serde/byte_writer.rs
irakliyk: Agreed - though this will introduce more changes than I'd like to handle in this PR. So, leaving it for the future.

kevinlewi:
@irakliyk In a follow-up, can you add the license headers to the new files introduced in this PR?

irakliyk:
@kevinlewi - ah sorry! Done in https://github.com/facebook/winterfell/commit/c743557f21b64ef684402211cc8621eea47b3745.

PR #151, opened 2023-02-20
Security estimate

Nashtare:
This PR adds proven security level estimate.

Examples now show both conjectured and proven security levels.

The $m$ parameter in the proven security is updated dynamically to optimize the balance between field security and query security. The approach taken is based on the two following assumptions:

- For the "pre-FRI-query err" as mentioned in [ethSTARK paper](https://eprint.iacr.org/2021/582.pdf), only the third (dominant) component, i.e. $\dfrac{(m+\frac{1}{2})^7 \cdot n^2}{2\rho^{3/2}q}$, is considered.
- For the "FRI-query err", the value $\sqrt{\rho} \cdot (1 + \frac{1}{2m})$ is approximated by $\sqrt{\rho}$. The reason for this is that the term $\frac{1}{2m} \leq \frac{1}{6}$ is fairly small, and doubling $m$ removes about 7 bits of security for the pre-FRI-query term.

Based on those two assumptions, we can solve (using the paper notations):
$$-\log_2{\dfrac{(m+\frac{1}{2})^7 \cdot n^2}{2\rho^{3/2}q}} = \zeta - s \cdot \log_2{\sqrt{\rho}}$$ which yields

$$\log_2{(m + \frac{1}{2})} = \dfrac{1 + \log_2{q} + \frac{s+3}{2} \log_2{\rho} - \zeta - 2\log_2{n}}{7} = T$$

$$m = 2^T - \frac{1}{2}$$

with
- $\zeta$ the grinding factor
- $\rho$ the inverse of the blowup factor $b$ (i.e. $\log_2{\rho} = -\log_2{b}$)
- $s$ the number of queries
- $q$ the extended field size
- $n$ the extended trace length

### DIFF
--- a/air/src/proof/mod.rs
+++ b/air/src/proof/mod.rs
@@ -108,7 +108,7 @@ impl StarkProof {
             get_conjectured_security(
                 self.context.options(),
                 self.context.num_modulus_bits(),
-                self.lde_domain_size() as u64,
+                self.trace_length() as u64,
                 H::COLLISION_RESISTANCE,
             )
         } else {
@@ -181,12 +181,12 @@ impl StarkProof {
 fn get_conjectured_security(
     options: &ProofOptions,
     base_field_bits: u32,
-    lde_domain_size: u64,
+    trace_domain_size: u64,
     collision_resistance: u32,
 ) -> u32 {
     // compute max security we can get for a given field size
     let field_size = base_field_bits * options.field_extension().degree();
-    let field_security = field_size - lde_domain_size.trailing_zeros();
+    let field_security = field_size - trace_domain_size.trailing_zeros();
 
     // compute security we get by executing multiple query rounds
     let security_per_query = log2(options.blowup_factor());
--- a/air/src/proof/mod.rs
+++ b/air/src/proof/mod.rs
@@ -112,8 +112,12 @@ impl StarkProof {
                 H::COLLISION_RESISTANCE,
             )
         } else {
-            // TODO: implement provable security estimation
-            unimplemented!("proven security estimation has not been implement yet")
+            get_proven_security(
+                self.context.options(),
+                self.context.num_modulus_bits(),
+                self.lde_domain_size() as u64,
+                H::COLLISION_RESISTANCE,
+            )
         }
     }
 
@@ -195,6 +199,54 @@ fn get_conjectured_security(
 
     cmp::min(
         cmp::min(field_security, query_security) - 1,
+        collision_resistance,
+    )
+}
+
+/// Estimates proven security level for the specified proof parameters.
+fn get_proven_security(
+    options: &ProofOptions,
+    base_field_bits: u32,
+    lde_domain_size: u64,
+    collision_resistance: u32,
+) -> u32 {
+    let extension_field_bits = (base_field_bits * options.field_extension().degree()) as f64;
+    let blowup_bits = log2(options.blowup_factor()) as f64;
+    let num_fri_queries = options.num_queries() as f64;
+    let lde_size_bits = lde_domain_size.trailing_zeros() as f64;
+
+    // m is a parameter greater or equal to 3.
+    // A larger m gives a worse field security bound but a better query security bound.
+    // An optimal value of m is then a value that would balance field and query security
+    // but there is no simple closed form solution.
+    // This sets m so that field security is equal to the best query security for any value
+    // of m, unless the calculated value is less than 3 in which case it gets rounded up to 3.
+    let mut m = extension_field_bits + 1.0;
+    m -= options.grinding_factor() as f64;
+    m -= (num_fri_queries + 3.0) / 2.0 * blowup_bits;
+    m -= 2.0 * lde_size_bits;
+    m /= 7.0;
+    m = 2.0_f64.powf(m);
+    m -= 0.5;
+    m = m.max(3.0);
+
+    // compute pre-FRI query security
+    // this considers only the third component given in the corresponding part of eq. 20
+    // in https://eprint.iacr.org/2021/582, i.e. (m+1/2)^7.n^2 / (2\rho^1.5.q) as all
+    // other terms are negligible in comparison.
+    let pre_query_security = (extension_field_bits + 1.0
+        - 3.0 / 2.0 * blowup_bits
+        - 2.0 * lde_size_bits
+        - 7.0 * (m + 0.5).log2()) as u32;
+
+    // compute security we get by executing multiple query rounds
+    let security_per_query = 0.5 * blowup_bits - (1.0 + 1.0 / (2.0 * m)).log2();
+    let mut query_security = (security_per_query * num_fri_queries) as u32;
+
+    query_security += options.grinding_factor();
+
+    cmp::min(
+        cmp::min(pre_query_security, query_security) - 1,
         collision_resistance,
     )
 }
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -106,14 +106,14 @@ impl ExampleOptions {
     }
 
     /// Returns security level of the input proof in bits.
-    pub fn get_proof_security_level(&self, proof: &StarkProof) -> usize {
+    pub fn get_proof_security_level(&self, proof: &StarkProof, conjectured: bool) -> usize {
         let security_level = match self.hash_fn.as_str() {
-            "blake3_192" => proof.security_level::<Blake3_192>(true),
-            "blake3_256" => proof.security_level::<Blake3_256>(true),
-            "sha3_256" => proof.security_level::<Sha3_256>(true),
-            "rp64_256" => proof.security_level::<Rp64_256>(true),
-            "rp_jive64_256" => proof.security_level::<RpJive64_256>(true),
-            "griffin_jive64_256" => proof.security_level::<GriffinJive64_256>(true),
+            "blake3_192" => proof.security_level::<Blake3_192>(conjectured),
+            "blake3_256" => proof.security_level::<Blake3_256>(conjectured),
+            "sha3_256" => proof.security_level::<Sha3_256>(conjectured),
+            "rp64_256" => proof.security_level::<Rp64_256>(conjectured),
+            "rp_jive64_256" => proof.security_level::<RpJive64_256>(conjectured),
+            "griffin_jive64_256" => proof.security_level::<GriffinJive64_256>(conjectured),
             val => panic!("'{val}' is not a valid hash function option"),
         };
 
--- a/examples/src/main.rs
+++ b/examples/src/main.rs
@@ -76,8 +76,12 @@ fn main() {
 
     let proof_bytes = proof.to_bytes();
     debug!("Proof size: {:.1} KB", proof_bytes.len() as f64 / 1024f64);
-    let security_level = options.get_proof_security_level(&proof);
-    debug!("Proof security: {} bits", security_level);
+    let conjectured_security_level = options.get_proof_security_level(&proof, true);
+    let proven_security_level = options.get_proof_security_level(&proof, false);
+    debug!(
+        "Proof security: {} bits ({} proven)",
+        conjectured_security_level, proven_security_level,
+    );
     #[cfg(feature = "std")]
     debug!(
         "Proof hash: {}",


irakliyk (changes requested):
This looks great! Thank you so much! I left a couple of small comments inline.

One other thing: I haven't cross checked myself yet, but I'm curious how the provable soundness computations here match up with some of the results from [this paper](https://github.com/ulrich-haboeck/FRI-Survey/blob/main/FRI.pdf) (i.e., tables in section 3.5).

examples/src/lib.rs
```diff
@@ -106,14 +110,14 @@ impl ExampleOptions {
...
-            "rp_jive64_256" => proof.security_level::<RpJive64_256>(true),
-            "griffin_jive64_256" => proof.security_level::<GriffinJive64_256>(true),
+            "blake3_192" => proof.security_level::<Blake3_192>(conjectured),
+            "blake3_256" => proof.security_level::<Blake3_256>(conjectured),
+            "sha3_256" => proof.security_level::<Sha3_256>(conjectured),
+            "rp64_256" => proof.security_level::<Rp64_256>(conjectured),
+            "rp_jive64_256" => proof.security_level::<RpJive64_256>(conjectured),
+            "griffin_jive64_256" => proof.security_level::<GriffinJive64_256>(conjectured),
             val => panic!("'{val}' is not a valid hash function option"),
         };
```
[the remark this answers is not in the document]
Nashtare: Ah yes good point, I overlooked that.

air/src/proof/mod.rs
```diff
@@ -177,12 +180,10 @@ impl StarkProof {
 fn get_conjectured_security(
     options: &ProofOptions,
     base_field_bits: u32,
-    lde_domain_size: u64,
     collision_resistance: u32,
 ) -> u32 {
     // compute max security we can get for a given field size
-    let field_size = base_field_bits * options.field_extension().degree();
-    let field_security = field_size - lde_domain_size.trailing_zeros();
+    let field_security = base_field_bits * options.field_extension().degree();
```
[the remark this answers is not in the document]
Nashtare: I guess this depends on what we are expecting / considering from the conjecture. The one in the paper states that for the Toy Problem (which is conjectured as hard as the IOP), we do not consider the trace size, but only the underlying field size. This is what's been adopted in plonky2 (Goldilocks with degree 2), Risc0 (Goldilocks with degree 2 or BabyBear with degree 4).

I'm not against reverting the changes on this section to keep the consideration on the trace length (extended or not), but it does feel somewhat weird as it the conjectured estimate explicitly removes the need for this extra care.

examples/src/main.rs
```diff
@@ -77,7 +77,15 @@ fn main() {
...
-    debug!("Proof security: {} bits", security_level);
+    debug!(
+        "Proof security: {} bits ({})",
+        security_level,
+        if options.proven_security {
+            "proven security"
+        } else {
+            "conjectured security"
+        }
+    );
```
irakliyk, line 88: One other thought I had: what if instead of selecting proof security type via CLI argument, we just print out both levels. Maybe the output line could look something like this:
```
Proof security: 96 bits (40 bits proven)
```

Nashtare: I had actually went for it originally, but thought we wouldn't want to display the proven security by default as it may look quite scary. But if we think it's not a problem, then I can revert it!

irakliyk: Yeah - I thought about it too. But I guess it is better to show it and then people can decide for themselves which security assumptions they are comfortable with.

air/src/proof/mod.rs
```diff
@@ -198,3 +199,54 @@ fn get_conjectured_security(
...
+    m = 2.0_f64.powf(m);
+    m -= 0.5;
+    m = m.max(3.0);
+
+    // compute field security
+    // this considers only the third component given in the corresponding part of eq. 20
+    // in https://eprint.iacr.org/2021/582, i.e. (m+1/2)^7.n^2 / (2\rho^1.5.q) as all
+    // other terms are negligible in comparison.
+    let field_security = (extension_field_bits + 1.0
+        - 3.0 / 2.0 * blowup_bits
```
Nashtare, line 235:
```suggestion
        + 3.0 / 2.0 * blowup_bits
```
There's actually a mistake in the sign here. I'll fix it.

examples/src/main.rs
Nashtare: I'm calling the function twice (for both conjectured and proven), would you rather have the utility method returning some tuple (conjectured, proven) directly instead?

air/src/proof/mod.rs
Nashtare: Nevermind, double-checking this it is correct.

Nashtare:
@irakliyk I've done a few things since your review:
- I removed the commit updating the conjectured security. I feel like we'd need more discussion around whether we want to alter it (through the non-extended trace length, or removing this component completely) before doing any changes.
- I fixed the proven security computation (I'll update the README shortly to detail the approach for computing the closeness parameter $m$ and ease review. Plugging the values from the paper you mentioned, I get ($m$ used in the paper between parenthesis):

For $\lambda$ = 67:
| $b$        | $m$           | $\lambda$  |
| ------------- |:-------------:| -----:|
| 3      | 6.4 (6) | 69 |
| 4      | 3.9 (4)     |   69 |
| 5     | 3.0 (3)     |   68 |


For $\lambda$ = 112:
(_note here I had to alter the security estimate function manually to test for larger blowup factors_)
| $b$        | $m$           | $\lambda$  |
| ------------- |:-------------:| -----:|
| 6      | 21 (16) | 115 |
| 8      | 14 (7)     |   114 |
| 10     | 4.4 (3)    |   116 |


For $\lambda$ = 128:
| $b$        | $m$           | $\lambda$  |
| ------------- |:-------------:| -----:|
| 3      | 8.3 (9) | 128 |
| 4      | 4.9 (6)     |   128 |
| 5     | 3.0 (4)    |   127 |

air/src/proof/mod.rs
irakliyk: Hmm - I think there is still something odd here. It seems like increasing grinding factor lowers proven security. For example:

Running `lamport-a` example with `-q 27 -b 16 -g 0 `, I get:
```
Proof security: 107 bits (52 proven)
```

But if I run it with `q 27 -b 16 -g 20`, I get:
```
Proof security: 108 bits (46 proven)
```

Nashtare: Oh interesting, I was running tests without grinding to match Ulrich's paper values. Let me have a deeper look.

Nashtare: Ah, this arises because the dynamic update of $m$ _assumes_ we're considering the grinding factor, while the contribution threshold is 80 bits, which we very rarely reach with parameters targeted for the conjectured security. The easiest solution seems to fully remove the conditional addition (only for the proven security estimate), as otherwise we may run into circular dependencies for calculating $m$. What do you think?

examples/src/main.rs
irakliyk: I think it is fine to call it twice as you have now.

air/src/proof/mod.rs
irakliyk: > The easiest solution seems to fully remove the conditional addition

This would basically remove the floor, right? I think this is fine but we can probably just compute security level twice  - once assuming there is grinding factor and once assuming there isn't - and then if the value without grinding factor is over some threshold, we could pick the the result which includes the grinding factor.

Having said this, there might be two things to consider:
- Maybe the floor for proven security should be lower (i.g., 60 bits).
- Maybe we don't need the floor at all?

I can see arguments for both. Probably more for the former and less for the latter.

air/src/proof/mod.rs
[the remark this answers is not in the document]
irakliyk: I'm actually thinking it is OK if we change this to something like:
```Rust
let field_security = field_size - trace_domain_size.trailing_zeros();
```

air/src/proof/mod.rs
Nashtare: Although, we may be have sometimes the case where the pre_query security level is lower than the query security level, in which case the suggested display above may look weird (with something like `k bits proven, including n bits of grinding`, where `n > k`).

irakliyk: > The threshold is here to ensure that we don't rely solely on a huge PoW to reach sufficient security, and keeping one would mean we would almost never consider the impact on grinding.

Agreed. Let's just remove the floor for proven security. I guess it will be rather unlikely (or maybe even impossible?) for proven security (without floor) to be higher than conjectured security (with the floor).

air/src/proof/mod.rs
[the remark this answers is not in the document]
Nashtare: I had missed your comment! Updated it in latest commit. Also took care of the grinding floor contribution removal for proven security. Let me know if I have missed anything.

irakliyk:
@Nashtare - thank you! I think we are almost there. The results are pretty close to the expected ones but there are some sets of parameters that give counterintuitive results. For example:

Running `fib-small` example with `-q 38 -b 16 -g 20 -e 2` gives:
```
Proof security: 118 bits (83 proven)
```
But if we increase blowup factor to 32 (I had to make a small update to be able to run it) and run with `-q 38 -b 16 -g 20 -e 2`, we get:
```
Proof security: 118 bits (79 proven)
```
Basically, it seems like increasing blowup factor lowers security. I think this happens because for `-q 38 -b 32` it doesn't really matter what `-g` is as we can never get above 79 bits of proven security. But I didn't look deeply enough to understand yet why this has to be the case.

Nashtare:
Hmm interesting, I did went up in my experiments with blowup factor to meet Ulrich results and didn't notice anything suspicious.. I'll dive more into it

Nashtare:
This is actually to be expected. To replicate your values, I used `fib-small -n 1024` (let me know if you were in a different setting).

For the first case, `-q 38 -b 16 -g 20 -e 2 fib-small -n 1024`, we obtain the following internal values:

- `pre_query_security`: 84
- `query_security`: 67
- `query_security` + `grinding`: 87
- `m` = 3.0

As we can see, both estimates (after grinding adjustement are fairly close). We can't get better than that, as `m` is minimal, and increasing it would decrease `pre_query_security`, decreasing the overall security level.

For the second case, `-q 38 -b 32 -g 20 -e 2 fib-small -n 1024`, we obtain the following values:

- `pre_query_security`: 80
- `query_security`: 86
- `query_security` + `grinding`: 106
- `m` = 3.0

There is now an imbalance between the two levels, which is to be _expected_. Indeed, the increased LDE domain reduces the estimated security on the first component, while consequently increasing the query security level. When the `pre_query_security` is the limiting factor, the only thing we can do is reduce the closeness parameter `m`, down to its minimal value which is already reached here. What could be done, in this scenario, is to decrease the `query_security` level, without decreasing the overall security level, by going down from 38 to 27 queries. This would yield  `query_security` + `grinding`: 81.

Note: This situation can also arise with the conjectured security estimate, when `field_security` becomes the bottleneck with very large traces, and decreasing the overall security level after some threshold.

Nashtare:
One thing that I could think of to alleviate this possibly confusing behavior with larger traces/blowup factors, but that would imply quite some refactoring, would be to specify instead a target security level (similarly to what Miden does), along with possibly extra arguments like `security_model` (conjectured or proven), `optimization_level` (prover time, proof size), and then compute the required parameters (blowup, nb queries, grinding, ...) based on those. But that's probably not necessary for the `examples` crate.

irakliyk:
> would be to specify instead a target security level

This would be cool - but I think it could be built on top of the current structure rather than instead of it, and should probably go into a different PR.

> the increased LDE domain reduces the estimated security on the first component

Ah - this makes sense now! It is not that the increase in blowup factor by itself leads to lower security, it is that increase in blowup factor leads to increase in LDE domain, and for extension 2 field, this leads to decreased security (and if we are to use extension 3 field, then the the impact would be more "intuitive").

One last comment I have: in the doc comments for `get_proven_security` - should we say something like "Estimates proven security ..." rather than "Calculates"? If I'm understanding correctly, this code provides a pretty close estimate (i.e., usually with accuracy within 3 bits) but does not always give the exact result - right?

Nashtare:
> If I'm understanding correctly, this code provides a pretty close estimate (i.e., usually with accuracy within 3 bits) but does not always give the exact result - right?

Actually it's slightly different, the code does yield the bit of security claimed by the proven security analysis of the paper.

- for the query security part, we do calculate $\zeta - s \cdot \log_2\left(\sqrt{\rho} \cdot (1 + \frac{1}{2m})\right)$
- for the pre-query security, we only consider the third term inside the log, i.e. $\dfrac{\left(m + \frac{1}{2}\right)^7 \cdot n^2}{2\rho^{3/2}q}$. However, some analysis shows that considering the extra terms will not affect the estimate, by even 1 bit.

The difference with Ulrich's paper comes from the selection of the closeness parameter $m$, which, due to the dynamic update, is in most cases helping to gain 1 to 3 extra bits of security (it was always fixed to 3 for simplicity in the ethSTARK paper).

But it's true that given the way we compute it (with the two light assumptions to derive $m$, "estimate" sounds more accurate). I'll update it.

irakliyk (approved):

PR #152, opened 2023-02-20
nit: Typos and doc

Nashtare:
Fixes a few typos and adds the first row describing the used MDS Matrices in their doc for completeness.


irakliyk (approved):
All looks good! Thank you for catching these!

PR #155, opened 2023-02-20
Fix no-std build for winter-utils

### LINKED ISSUE #153: Fix `--no-default-features` build
`winter-utils` is failing to build with `--no-default-features` for all targets.

`Vec` should be pulled from `super`, and not from prelude, as they will not be available by default under `no-std`.

vlopes11:
closes #153


irakliyk (approved):
All looks good! Thank you!

PR #157, opened 2023-02-23
Clarify use of gamma in doc

### LINKED ISSUE #88: Does `Air::get_deep_composition_coefficients` assume that the proof is generated using an extension field?
This is not a real bug because I guess is at most generating an extra unused random coefficient. 
https://github.com/novifinancial/winterfell/blob/10a5f5ad90592889fe977c505012cd9f2159fb23/air/src/air/mod.rs#L512
In this line 3 coefficients are sampled for each trace polynomial: `\alpha_i` for the trace opening to `z`, `\beta_i` for the opening to `g*z`, and `\gamma_i` for showing that the trace poly is defined over the base field. If we are not working over an extension field then we don't need to prove the third statement. I guess this is not a big problem because, at most, the extra coefficient never gets read.

Nashtare:
Tiny PR to clear out potential confusion of generating triples of random coefficients, mentioned in #88.
I assumed that it was probably preferable to leave the triple as is, rather than handling the two cases separately.
@irakliyk let me know what you'd prefer.

closes #88.

### DIFF
--- a/air/src/air/coefficients.rs
+++ b/air/src/air/coefficients.rs
@@ -100,6 +100,9 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
 ///   to the highest constraint degree rounded to the next power of two. For example, if the
 ///   highest constraint degree is 6, $m$ will be equal to 8.
 /// * $\alpha_i, \beta_i, \gamma_i$ are composition coefficients for the $i$th trace polynomial.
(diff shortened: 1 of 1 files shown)


irakliyk (approved):
Looks good! Thank you!

There is a question though whether the conjugate check is needed at all (see [here](https://twitter.com/EliBenSasson/status/1625140730736828416)). I am hoping to remove some time in the future as it simplifies things a bit.

PR #159, opened 2023-03-01
Update Matrix Transpose Logic

itzmeanjan:
During transposing a column-major matrix to row-major one, prefer writing to contiguous memory locations. 

> **Note** This change brings ~2-3% performance improvement.


irakliyk:
Interesting! I'm seeing a similar improvement on an M1. Will check on an AMD CPU a bit later.

irakliyk:
I've benchmarked on AMD Ryzen 9 5950X and didn't see any noticeable improvement (but there was no degradation either).

irakliyk (approved):

PR #164, opened 2023-03-12
refactor: f64 fft inline hotspots

vlopes11:
This commit refactors the inline directive distribution, loosing the requirements on cold spots, and introducing new directives into hot spots.

It aims to decrease the code size so we reduce the L1 cache miss for core threads. The target use case is Miden VM proof generation, and we dropped from 7.64% L1 cache miss to ~2.3% on Intel x64 for a 2^21 proof generation. We achieved a significant performance gain for such proof generation of approx. 20% for both Intel x64 and Apple M1. The performance gain will present only for large matrices that will be more likely to suffer cache miss, so there will be no significant gain for small operations.

This commit also introduces an improvement on branch miss. It is much harder for rayon to properly optimize branches if we have `skip`, and for the fft implementation, it was trivial to just refactor the starting indexes as we deal mostly with slices of matrices. The branch miss dropped from ~8.2% to 0.89% on Intel x64.

The focus of this commit is restricted to f64 as it is the one consumed by Miden VM. The f62 and f128 implementations aren't covered by this commit, and can probably benefit from a similar refactor in the future.

A particularly critical implementation of `transpose_square_stretch` was refactored to use pointers, as `slice::swap` will perform checks that are redundant, and will keep expensive call traces so they can panic if the slice bounds are violated. This unsafe implementation will extend the safety of the previous implementation as the bounds of the slice are always checked and protected by the FftInputs type safety.

### DIFF
--- a/air/src/air/transition/frame.rs
+++ b/air/src/air/transition/frame.rs
@@ -57,25 +57,21 @@ impl<E: FieldElement> EvaluationFrame<E> {
     // --------------------------------------------------------------------------------------------
 
     /// Returns a reference to the current row.
-    #[inline(always)]
     pub fn current(&self) -> &[E] {
         &self.current
     }
 
     /// Returns a mutable reference to the current row.
-    #[inline(always)]
     pub fn current_mut(&mut self) -> &mut [E] {
         &mut self.current
     }
 
     /// Returns a reference to the next row.
-    #[inline(always)]
     pub fn next(&self) -> &[E] {
         &self.next
     }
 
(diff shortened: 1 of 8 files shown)


vlopes11:
I updated the changeset to be minimal for the Miden VM proof & rebased to current main

vlopes11:
@irakliyk let's close this for now as it will pile up conflicts. We should revisit the topic in the future and reorganize the inlined routines as that would cause significant cache miss drop

PR #166, opened 2023-03-12
Remove conjugate term in deep composition polynomial

Al-Kindi-0:
Following the remark [here](https://twitter.com/EliBenSasson/status/1625140730736828416) the present PR proposes a simplification in the way the deep composition polynomial is computed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.6.0 
+* Removed term involving conjugate OOD challenge z from deep composition polynomial (#166).
+
 ## 0.5.1 (2023-02-20)
 * Fix no-std build for winter-utils (#153)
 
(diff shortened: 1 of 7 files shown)


irakliyk (changes requested):
Looks good! Thank you! Two small comments:

1. Do we still need `RandomCoin::draw_triple()` method? If not, we could probably remove it.
2. Could you add a bullet point to the CHANGELOG describing the change? (this could probably go under v0.6 heading).

Al-Kindi-0:
Thank you!

1. You are right, it was useful only in the context of conjugate term. I have removed it.
2. Sure will do that.

irakliyk (approved):
Looks good! Thank you!

@Nashtare - curious if you have any comments.

Nashtare:
I had a look yesterday and it seemed good to me, modulo the same comment you had (if nothing changed in the meantime!).

PR #168, opened 2023-03-14
Integrate RowMatrix into the prover

irakliyk:
This PR integrates `RowMatrix` struct into the prover. Specifically:

- Changes the return type of `build_trace_commitment()` function in the prover.
- Updates `TraceLde` struct to hold `RowMatrix`'s instead of `ColMatrix`'s.
- Propagated the above changes to other components to make sure everything compiles.

For small proofs, the effect of these changes is usually not noticeable. However, for large proofs (e.g., $2^{20}$ rows, dozens of columns) the speed up in LDE and Merkle Tree construction steps could be significant.


irakliyk:
@Nashtare, @Al-Kindi-0 - would appreciate a review, if you get a chance.

Nashtare:
> However, for large proofs (e.g., $2^{20}$ rows, dozens of columns) the speed up in LDE and Merkle Tree construction steps could be significant.

Just to complement on this point, benchmarking fib-small (2 columns) and lamport-threshold (28 columns), the two extremities of the spectrum among the examples, I am getting on my machine:

- a slowdown of 140% for the LDE extension step on the fib-small example (2^20 rows, blowup = 16), from 1406ms to 3375ms.
- a speedup of 48% for the LDE extension step on the lamport-t example (2^19 rows, blowup = 16), from 16672ms to 8803ms.

I don't think the slowdown for narrow traces is really an issue as practically speaking we will most likely always optimize proving time (and amortize verifier if needed through recursive composition in a near future), but thought it was worth mentioning in case some people notice a performance downgrade on the existing examples (most of which having very few columns).

prover/src/lib.rs
```diff
@@ -456,17 +456,17 @@ pub trait Prover {
...
+    ) -> (RowMatrix<E>, MerkleTree<Self::HashFn>, ColMatrix<E>)
     where
         E: FieldElement<BaseField = Self::BaseField>,
     {
         // extend the execution trace
         #[cfg(feature = "std")]
         let now = Instant::now();
         let trace_polys = trace.interpolate_columns();
-        let trace_lde = trace_polys.evaluate_columns_over(domain);
+        let trace_lde = RowMatrix::evaluate_polys_over::<8>(&trace_polys, domain);
```
Nashtare, line 469: I'm curious how the value 8 has been chosen, was it experimentally?

prover/src/matrix/row_matrix.rs
```diff
@@ -50,25 +52,40 @@ impl<E: FieldElement> RowMatrix<E> {
     ///
     /// To improve performance, polynomials are evaluated in batches specified by the `N` type
     /// parameter. Minimum batch size is 0.
-    pub fn evaluate_polys<const N: usize>(polys: &Matrix<E>, blowup_factor: usize) -> Self {
-        assert!(N > 0, "batch size N must be greater than zero");
```
Nashtare, line 54: Even though it is triggered in `build_segments()`, we may want to keep the check at the higher level? Not really blocking though.

prover/src/matrix/row_matrix.rs
```diff
@@ -126,6 +156,39 @@ impl<E: FieldElement> RowMatrix<E> {
...
     }
+
+    // COMMITMENTS
+    // --------------------------------------------------------------------------------------------
+
+    /// Returns a commitment to this matrix.
+    ///
+    /// The commitment is built as follows:
+    /// * Each row of the matrix is hashed into a single digest of the specified hash function.
+    /// * The resulting values are used to built a binary Merkle tree such that each row digest
```
Nashtare, line 167:
```suggestion
    /// * The resulting values are used to build a binary Merkle tree such that each row digest
```

prover/src/matrix/row_matrix.rs
```diff
@@ -126,6 +156,39 @@ impl<E: FieldElement> RowMatrix<E> {
...
+    ///   becomes a leaf in the tree. Thus, the number of leaves in the tree is equal to the
+    ///   number of rows in the matrix.
+    /// * The resulting Merkle tree is return as the commitment to the entire matrix.
```
Nashtare, line 170:
```suggestion
    /// * The resulting Merkle tree is returned as the commitment to the entire matrix.
```

prover/src/trace/trace_lde.rs
```diff
@@ -3,25 +3,31 @@
...
 use math::FieldElement;
 use utils::collections::Vec;
 
 // TRACE LOW DEGREE EXTENSION
 // ================================================================================================
-/// TODO: add docs
+/// Contains all segments of the extended execution trace.
+///
+/// Segments are stored in two groups:
+/// - Main segment: this it the first trace segment generated by the prover. Values in this segment
```
Nashtare, line 16:
```suggestion
/// - Main segment: this is the first trace segment generated by the prover. Values in this segment
```

prover/src/trace/trace_lde.rs
```diff
@@ -3,25 +3,31 @@
...
+///   will always be elements in the base field (even when extension field is used).
```
Nashtare, line 17:
```suggestion
///   will always be elements in the base field (even when an extension field is used).
```

prover/src/matrix/row_matrix.rs
irakliyk: Agreed! I've added this check to all constructors.

prover/src/lib.rs
irakliyk: Yes, this seems to work the best for 64-bit fields (at least on Apple M1). I'll create a constant for this and will add some comments.

irakliyk:
Thank you for the review!

> a slowdown of 140% for the LDE extension step on the fib-small example (2^20 rows, blowup = 16), from 1406ms to 3375ms.

Ah - interesting! I have't tested it on such narrow traces - but I think with the current settings it is to be expected. Basically, since segment width is set to 8, we actually run NTTs on 8 columns instead of 2. I wonder if this degradation can be eliminated by setting segment width to 2.

> a speedup of 48% for the LDE extension step on the lamport-t example (2^19 rows, blowup = 16), from 16672ms to 8803ms.

What machine are you running this on? I've tested on an M1 and on AMD Ryzen 9 5950X and there was no improvement in the former and about 3% degradation in the latter (though, changing segment width to 4 resulted in a slight improvement).

> I don't think the slowdown for narrow traces is really an issue

Yes, that's what I think too. In most practical situations we'll have more than just a few columns and if needed, we might be able to get back to the original performance by adjusting segment width.

Al-Kindi-0 (commented):
LGTM, thank you @irakliyk !
I left a small comment only.

Nashtare:
> What machine are you running this on? I've tested on an M1 and on AMD Ryzen 9 5950X and there was no improvement in the former and about 3% degradation in the latter (though, changing segment width to 4 resulted in a slight improvement).

Same as before when I'm commenting about benchmarks, Ubuntu 22.04 Intel® Core™ i7-9750H CPU @ 2.60GHz × 12 CPUs. (although as we've discussed in the past, there may be weird results sometimes with Linux and/or x86_64, as with the RP64 benchmarks for instance)..

prover/src/matrix/row_matrix.rs
```diff
@@ -49,26 +51,45 @@ impl<E: FieldElement> RowMatrix<E> {
     /// in the `polys` matrix) and the `blowup_factor`.
     ///
     /// To improve performance, polynomials are evaluated in batches specified by the `N` type
-    /// parameter. Minimum batch size is 1.
-    pub fn evaluate_polys<const N: usize>(polys: &Matrix<E>, blowup_factor: usize) -> Self {
+    /// parameter. Minimum batch size is 0.
```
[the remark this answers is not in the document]
irakliyk: Thank you! Fixed!

PR #169, opened 2023-03-19
New root of unity for f64 field

irakliyk:
This PR changes the "two-adic" root of unity for the `f64` field. The new root of unity has a nice property that for domains of size 64 the root of unity is 8, and thus many multiplications can be replaced with shifts for NTT computations (though, we are not taking advantage of this yet).


Nashtare:
Not sure if it's a big deal but technically the SAGE code in the comment of the root is not really meaningful, as any n-root $r$ of unity with $n < 2^{32}$ will yield $r^2{32}=1$.

irakliyk:
> the SAGE code in the comment of the root is not really meaningful

Ah, yes! I did something similar to what you've suggested.

PR #171, opened 2023-03-22
FRI options cleanup

irakliyk:
This PR does minor refactoring of FRI and proof options structs to store max remainder degree and increase limits for some parameters.


Al-Kindi-0 (commented):
Looks great, especially the simplifications to `max_remainder_degree`. Thank you!

PR #172, opened 2023-03-22
To element reduction of public inputs and proof context

irakliyk:
This PR implements reduction to elements of public inputs and proof context. This will allow `RandomCoin` to be initialized from a list of field elements (rather than bytes). The advantage of this approach is that it simplifies handling of random coin in the context of recursive proofs.

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -4,8 +4,10 @@
 // LICENSE file in the root directory of this source tree.
 
 use fri::FriOptions;
-use math::StarkField;
-use utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable};
+use math::{StarkField, ToElements};
+use utils::{
+    collections::Vec, ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
+};
 
 // CONSTANTS
 // ================================================================================================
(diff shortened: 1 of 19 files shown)


irakliyk:
@Nashtare, @Al-Kindi-0 - would appreciate a review on this.

air/src/options.rs
```diff
@@ -206,6 +208,22 @@ impl ProofOptions {
     }
 }
 
+impl<E: StarkField> ToElements<E> for ProofOptions {
+    fn to_elements(&self) -> Vec<E> {
+        // encode filed extension and FRI parameters into a single field element
```
Al-Kindi-0, line 213: nit: field

air/src/proof/context.rs
```diff
@@ -93,6 +93,45 @@ impl Context {
...
+        let (m1, m2) = self.field_modulus_bytes.split_at(num_modulus_bytes / 2);
+        result.push(bytes_to_element(m1));
+        result.push(bytes_to_element(m2));
+
+        // convert proof options and trace length to elements
+        result.append(&mut self.options.to_elements());
+        result.push(E::from(self.trace_length as u64));
+
+        // convert trace metadata to elements; this is done by breaking trace metadata into chunks
+        // of bytes which are slightly smaller than field element requires, and then converting
```
Al-Kindi-0, line 123: nit: sentence containing "requires" should probably be reformulated

Al-Kindi-0: This sentence from the same PR should work: "..is slightly smaller than the number of bytes needed to encode a field element"

Al-Kindi-0 (commented):
Looks great, thank you!
I left only some small nits.

Nashtare:
Nothing else to add 🙂

air/src/proof/context.rs
irakliyk: Thank you! Fixed!

Issue #162, opened 2023-03-08
`RandomCoin` as a trait


Al-Kindi-0:
This issue aims to start a discussion on the merits of changing `RandomCoin<B: StarkField, H: Hasher>` from a struct to a trait. The reasons for this being that for some choices of `H` (specifically the ones based on the sponge construction) as well as of `B` (e.g. when it is the base field of the algebraic hasher `H`) seeding (which can be done with absorbing) and drawing random values (which can be done by squeezing) can be achieved in more efficient ways.

irakliyk:
This would be great! I've been thinking about something like this as well. Maybe the trait could look like this:
```Rust
pub trait RandomCoin {
    type BaseField: StarkField;
    type Hasher: Hasher;

    // REQUIRED METHODS
    // --------------------------------------------------------------------------------------------

    fn new(seed: &[Self::BaseField]) -> Self;

    fn reseed(&mut self, data: <Self::Hasher as Hasher>::Digest);

    fn reseed_with_int(&mut self, value: u64);

    fn check_leading_zeros(&self, value: u64) -> u32;

    fn leading_zeros(&self) -> u32;

    fn draw<E: FieldElement<BaseField = Self::BaseField>>(&mut self) -> Result<E, RandomCoinError>;

    // PROVIDED METHODS
    // --------------------------------------------------------------------------------------------

    fn draw_pair<E>(&mut self) -> Result<(E, E), RandomCoinError>
    where
        E: FieldElement<BaseField = Self::BaseField>,
    {
        Ok((self.draw()?, self.draw()?))
    }

    fn draw_triple<E>(&mut self) -> Result<(E, E, E), RandomCoinError>
    where
        E: FieldElement<BaseField = Self::BaseField>,
    {
        Ok((self.draw()?, self.draw()?, self.draw()?))
    }

    fn draw_integers(
        &mut self,
        num_values: usize,
        domain_size: usize,
    ) -> Result<Vec<usize>, RandomCoinError> {
        todo!()
    }
}
```

irakliyk:
Closed by #176.

PR #173, opened 2023-03-22
change RandomCoin to a trait with a default struct impl

grjte:
This PR addresses #162. It adds a `RandomCoin` trait and migrates the previous `RandomCoin` struct to a new default implementation of the `RandomCoin` trait.


air/src/air/mod.rs
```diff
@@ -504,7 +504,7 @@ pub trait Air: Send + Sync {
     /// composition polynomial.
     fn get_constraint_composition_coefficients<E, H>(
         &self,
-        public_coin: &mut RandomCoin<Self::BaseField, H>,
+        public_coin: &mut RandomCoinImpl<Self::BaseField, H>,
     ) -> Result<ConstraintCompositionCoefficients<E>, RandomCoinError>
     where
```
irakliyk, line 509: `RandomCoin` should be a generic type parameter. This probably also means that `H` would need to be `ElementHasher` rather than just `Hasher`.

air/src/air/mod.rs
```diff
@@ -530,7 +530,7 @@ pub trait Air: Send + Sync {
     /// composition polynomial.
     fn get_deep_composition_coefficients<E, H>(
         &self,
-        public_coin: &mut RandomCoin<Self::BaseField, H>,
+        public_coin: &mut RandomCoinImpl<Self::BaseField, H>,
     ) -> Result<DeepCompositionCoefficients<E>, RandomCoinError>
```
irakliyk, line 534: `RandomCoin` should be a generic type parameter. This probably also means that `H` would need to be `ElementHasher` rather than just `Hasher`.

crypto/src/random/mod.rs
```diff
@@ -43,20 +50,21 @@ use utils::collections::Vec;
...
-/// let mut coin2 = RandomCoin::<BaseElement, Blake3_256<BaseElement>>::new(&[2, 3, 4, 5]);
+/// let mut coin1 = RandomCoinImpl::<BaseElement, Blake3_256<BaseElement>>::new(seed);
+/// let seed = Blake3_256::<BaseElement>::hash(&[2, 3, 4, 5]);
+/// let mut coin2 = RandomCoinImpl::<BaseElement, Blake3_256<BaseElement>>::new(seed);
 /// let e1 = coin1.draw::<BaseElement>().unwrap();;
 /// let e2 = coin2.draw::<BaseElement>().unwrap();;
 /// assert_ne!(e1, e2);
 /// ```
-pub struct RandomCoin<B, H>
+pub struct RandomCoinImpl<B, H>
```
irakliyk, line 67: I would probably move `RandomCoin` trait here and the implementation into a sub-module. Also, maybe we should use something like `DefaultRandomCoin` instead of `RandomCoinImpl` to indicate that it just a default implementation?

crypto/src/random/random_coin.rs
```diff
@@ -0,0 +1,101 @@
...
+use math::{FieldElement, StarkField};
+use utils::collections::Vec;
+
+// RANDOM COIN TRAIT
+// ================================================================================================
+
+/// Pseudo-random element generator for finite fields.
+///
+/// A random coin can be used to draws elements uniformly at random from the specified base field
+// (which is specified via the `B` type parameter) or from any extension of the base field.
```
irakliyk, line 16: typo: missing one `/`.

crypto/src/random/random_coin.rs
```diff
@@ -0,0 +1,101 @@
...
+///
+/// Internally we use a cryptographic hash function (which is specified via the `H` type parameter),
+/// to draw elements from the field. The coin works roughly as follows:
+/// - The internal state of the coin consists of a `seed` and a `counter`. At instantiation
+///   time, the `seed` is set to a hash of the provided bytes, and the `counter` is set to 0.
+/// - To draw the next element, we increment the `counter` and compute hash(`seed` || `counter`).
+///   If the resulting value is a valid field element, we return the result; otherwise we try
+///   again until a valid element is found or the number of allowed tries is exceeded.
+/// - We can also re-seed the coin with a new value. During the reseeding procedure, the
+///   seed is set to hash(`old_seed` || `new_seed`), and the counter is reset to 0.
```
irakliyk, line 26: I would probably remove it as concrete implementations may chose to do this differently.

crypto/src/random/random_coin.rs
```diff
@@ -0,0 +1,101 @@
...
+    ///
+    /// # Errors
+    /// Returns an error if any of the field elements could not be generated after 100 calls to
+    /// the PRNG;
+    fn draw_triple<E>(&mut self) -> Result<(E, E, E), RandomCoinError>
+    where
+        E: FieldElement<BaseField = Self::BaseField>,
+    {
+        Ok((self.draw()?, self.draw()?, self.draw()?))
+    }
```
irakliyk, line 100: I don't think `draw_triple()` is needed any more.

irakliyk (changes requested):
Looks good! Thank you! I left some comments inline. The biggest thing is that `RandomCoin` should be handled generically pretty much everywhere (except for tests and examples). That is, we should be able to specify a random coin implementation for a given prover and verifier.

fri/src/verifier/mod.rs
```diff
@@ -100,7 +100,7 @@ where
     /// * An error was encountered while drawing a random α value from the coin.
     pub fn new(
         channel: &mut C,
-        public_coin: &mut RandomCoin<B, H>,
+        public_coin: &mut RandomCoinImpl<B, H>,
         options: FriOptions,
         max_poly_degree: usize,
     ) -> Result<Self, VerifierError> {
```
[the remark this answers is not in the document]
grjte: > Separately, we can probably get rid of B type parameter as it is equivalent to E::BaseField.

I'm not sure if I interpreted this comment correctly, so I made this change as a separate commit.

grjte:
Thanks for your review @irakliyk. I've made the suggested changes, and since `H` is now `ElementHasher`, I also changed the `new` method to take a slice of elements rather than a digest.

irakliyk (changes requested):
Looks great! Thank you! I left just a few small comments inline.

Al-Kindi-0 (commented):
Just a small nit :)

irakliyk (approved):
All looks good! Thank you!

PR #176, opened 2023-03-24
change RandomCoin to a trait with a default struct impl 

grjte:
This PR addresses https://github.com/facebook/winterfell/issues/162. It adds a RandomCoin trait and migrates the previous RandomCoin struct to a new default implementation of the RandomCoin trait.


irakliyk (approved):
All looks good! Thank you!

PR #178, opened 2023-03-26
Minor trait and derive fixes

irakliyk:
This PR makes minor fixes to trait bounds and also adds `Clone` derive to error enums.


grjte (commented):
Looks good, thank you!

PR #182, opened 2023-03-29
Fix `no_std` support

frisitano:
The `get_proven_security` function uses both `log2` and `powf` which are both incompatible with a `no_std` environment.  To fix support for `no_std` this PR moves the `get_proven_security` function behind a `std` feature flag and panics if it is called in a `no_std` context.

### DIFF
--- a/air/src/proof/mod.rs
+++ b/air/src/proof/mod.rs
@@ -112,6 +112,10 @@ impl StarkProof {
                 H::COLLISION_RESISTANCE,
             )
         } else {
+            #[cfg(not(feature = "std"))]
+            panic!("proven security level is not available in no_std mode");
+
+            #[cfg(feature = "std")]
             get_proven_security(
                 self.context.options(),
                 self.context.num_modulus_bits(),
@@ -203,6 +207,7 @@ fn get_conjectured_security(
     )
 }
 
+#[cfg(feature = "std")]
 /// Estimates proven security level for the specified proof parameters.
 fn get_proven_security(
     options: &ProofOptions,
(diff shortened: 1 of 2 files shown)


air/src/proof/mod.rs
```diff
@@ -203,6 +207,7 @@ fn get_conjectured_security(
     )
 }
 
+#[cfg(feature = "std")]
```
vlopes11, line 210: It's not immediately obvious why we need to restrict that to `std` as (from a quick reading) it should depend only on `core`. Maybe add some docs on why that is needed?

Nashtare: How are you obtaining this? Running `cargo build --no-default-features` and then `./target/debug/winterfell fib-small -n 32` runs fine on my machine, see below:

```Computed Fibonacci sequence up to 32th term in 0 ms
Generating proof for computing Fibonacci sequence (2 terms per step) up to 32th term
---------------------
Generated execution trace of 2 registers and 2^4 steps in 0 ms
---------------------
Proof generated in 66 ms
Proof size: 3.9 KB
Proof security: 59 bits (32 proven)
---------------------
Proof verified in 1.0 ms
```

bitwalker: Those functions are not available in no-std environments, at least as methods on `f64` values - however, you can call the underlying intrinsics directly, which are defined in `core` [here](https://github.com/rust-lang/rust/blob/cf32b9de1e8f66526c36ad2927458558d2e81093/library/core/src/intrinsics.rs#L1473). The functions defined in `std` just delegate to these intrinsics, so it is safe to call them directly on WebAssembly at least (most targets in fact, but there are a few edge cases handled by the function in libstd for a couple of specific target configurations, e.g. android). That should be a sufficient work around for this specific error anyway.

PR #185, opened 2023-04-14
MerkleTree, Segment, ColMatrix changes for GPU RPO support

andrewmilson:
Changes:
* Make MerkleTree fields public so the merkle nodes can be generated in a custom way (on the GPU in my case)
* Add merge_column to ColMatrix to append columns
* Make get_offsets public
* Make Segment public
* remove .data() from Segment and instead implement Deref (gives access to the data Vec instead of the slice which I need to call .capacity() on)

### DIFF
--- a/crypto/src/merkle/mod.rs
+++ b/crypto/src/merkle/mod.rs
@@ -94,8 +94,9 @@ pub struct MerkleTree<H: Hasher> {
 // ================================================================================================
 
 impl<H: Hasher> MerkleTree<H> {
-    // CONSTRUCTOR
+    // CONSTRUCTORS
     // --------------------------------------------------------------------------------------------
+
     /// Returns new Merkle tree built from the provide leaves using hash function specified by the
     /// `H` generic parameter.
     ///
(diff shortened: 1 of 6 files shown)


prover/src/matrix/segments.rs
```diff
@@ -54,31 +56,63 @@ impl<B: StarkField, const N: usize> Segment<B, N> {
...
+    /// Evaluation is performed over the domain specified by the provided twiddles and offsets.
+    ///
+    /// # Panics
+    /// Panics if:
+    /// - `poly_offset` greater than or equal to the number of base field columns in `polys`.
+    /// - Number of offsets is not a power of two.
+    /// - Number of offsets is smaller than or equal to the polynomial size.
+    /// - The number of twiddles is not half the size of the polynomial size.
+    /// - Number of offsets is smaller than the length of the data buffer
+    pub fn new_with_buffer<E>(
```
andrewmilson, line 85: Feel free to take it from here @irakliyk

prover/src/matrix/segments.rs
```diff
@@ -54,31 +56,63 @@ impl<B: StarkField, const N: usize> Segment<B, N> {
...
         assert_eq!(poly_size, twiddles.len() * 2);
         assert!(poly_offset < polys.num_base_cols());
+        assert_eq!(data.len(), domain_size);
 
         // determine the number of polynomials to add to this segment; this number can be either N,
         // or smaller than N when there are fewer than N polynomials remaining to be processed
         let num_polys_remaining = polys.num_base_cols() - poly_offset;
         let num_polys = if num_polys_remaining < N {
+            // the segment will remain unfilled so we pad it with zeros
+            data.fill([B::ZERO; N]);
```
irakliyk, line 110: I would probably move this initialization into `new()` constructor. My thinking is that if we are using `new_with_buffer()`, the caller is responsible for initializing the buffer as they need.

crypto/src/merkle/mod.rs
```diff
@@ -86,8 +86,8 @@ mod tests;
 /// ```
 #[derive(Debug)]
 pub struct MerkleTree<H: Hasher> {
-    nodes: Vec<H::Digest>,
-    leaves: Vec<H::Digest>,
+    pub nodes: Vec<H::Digest>,
+    pub leaves: Vec<H::Digest>,
 }
```
irakliyk, line 91: I don't mind this change, but I wonder if instead we could meet the requirements by adding another constructor. Maybe something like:
```Rust
pub fn from_raw_parts(nodes: Vec<H::Digest>, leaves: Vec<H::Digest>) -> Self
```

irakliyk (changes requested):
Looks great! Thank you! I left a few small comments inline - mostly related to API ergonomics.

prover/src/lib.rs
```diff
@@ -78,7 +78,7 @@ mod domain;
 pub use domain::StarkDomain;
 
 mod matrix;
-pub use matrix::{ColMatrix, RowMatrix};
+pub use matrix::{get_offsets, ColMatrix, RowMatrix, Segment};
```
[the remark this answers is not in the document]
irakliyk: Actually, let's do this as
```rust
pub mod matrix;
pub use matrix::{ColMatrix, RowMatrix};
```
So that we don't introduce a breaking change.

irakliyk (approved):
All looks good! Thank you!

PR #186, opened 2023-04-15
Remove log2() and use .ilog2() instead

andrewmilson:
Removed log2() since .ilog2() has been stabilized https://blog.rust-lang.org/2023/01/26/Rust-1.67.0.html#stabilized-apis

### DIFF
--- a/air/src/air/boundary/tests.rs
+++ b/air/src/air/boundary/tests.rs
@@ -8,7 +8,7 @@ use super::{
     Assertion, BoundaryConstraint,
 };
 use crypto::{hashers::Blake3_256, DefaultRandomCoin, RandomCoin};
-use math::{fields::f128::BaseElement, log2, polynom, FieldElement, StarkField};
+use math::{fields::f128::BaseElement, polynom, FieldElement, StarkField};
 use rand_utils::{rand_value, rand_vector, shuffle};
 use utils::collections::{BTreeMap, Vec};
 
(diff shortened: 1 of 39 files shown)


air/src/options.rs
```diff
@@ -68,7 +68,7 @@ pub enum FieldExtension {
 ///    time decrease the number of queries in such a way that the proofs become smaller.
 /// 4. Grinding factor - higher values increase proof soundness, but also may increase proof
 ///    generation time. More precisely, conjectured proof soundness is bounded by
-///    `num_queries * log2(blowup_factor) + grinding_factor`.
+///    `num_queries * blowup_factor.ilog2() + grinding_factor`.
```
irakliyk, line 71: Here, and in other places in this file, I'd probably keep the comments as they were as they more for explaining concepts rather than for demonstrating code.

math/src/utils/mod.rs
```diff
@@ -187,24 +187,6 @@ where
...
-/// assert_eq!(log2(1), 0);
-/// assert_eq!(log2(16), 4);
-/// assert_eq!(log2(1 << 20), 20);
-/// assert_eq!(log2(2usize.pow(20)), 20);
-/// ```
-pub fn log2(n: usize) -> u32 {
-    assert!(n.is_power_of_two(), "n must be a power of two");
-    n.trailing_zeros()
-}
-
```
irakliyk, line 207: I would probably keep this here for now (not to introduce a breaking change). We can get rid of it right before the v0.7 release.

irakliyk (changes requested):
Looks great! Thank you! I've actually wanted to do this myself for some time now.

I left two very minor comments inline.

irakliyk (approved):
All looks good! Thank you!

Issue #187, opened 2023-04-15
Support for Tip3


ameya-deshmukh:
### @irakliyk hi, I am doing research on recursive STARKs and was wondering if support for arithmetization-oriented hash functions which are suitable for recursive verification is possible within Winterfell. I am especially looking at [Tip3](https://eprint.iacr.org/2023/107.pdf) and was curious to know the feasibility of adding support in Winterfell for the same. Would be more than happy to take this up as an independent task under your guidance. Thanks!

jan-ferdinand:
@ameya-deshmukh, I like your idea and have taken the liberty to start hacking on it. The PR (#188) is not complete yet, as described in greater detail there. I'm not sure how much additional time I'll spend coding on the PR – feel free to take it as a draft for your own hacking. :blush:

ameya-deshmukh:
Wow, one of the authors of the paper himself! Glad you liked the idea. I'll be hacking on it independently and would love to compare code across implementations. Are there any specific optimizations etc that you'd like to see added which you may not have the time to implement? Would love to do it

jan-ferdinand:
Most of the code I wrote for that PR is adapter code, _i.e._, getting the reference implementation of Tip5 to fit the requirements of the winterfell codebase. As far as we can tell, the reference implementation of Tip5 is in a pretty good spot right now.[^0] This does not mean you shouldn't go hunting for speedups if you want to – we welcome contributions to the code base!

Additionally, it's quite possible that the adapter stuff I hacked together for the PR is not optimal. For example, I chose to go with chaining iterators instead of updating mutable slices, without knowing which of the two approaches is faster. Gut feel says that this won't be the bottleneck, but gut feeling is not a very good profiler. :grin:

[^0]: There is one speed improvement that we will incorporate (hopefully) within the upcoming week, but other than that, I currently know of no optimizations that we haven't implemented yet.

ameya-deshmukh:
Got it! I actually want to discuss the project I'm working on with you, what's the best channel of communication for that?

jan-ferdinand:
I messaged you on an instant messenger. :+1:

PR #188, opened 2023-04-15
add hash function Tip5

jan-ferdinand:
This PR is trying to address #187.

Unfortunately, the tests are not passing. I suspect this has to do with me raising the upper bound for the number of bytes in a `Digest` (see [here](https://github.com/facebook/winterfell/blob/main/crypto/src/hash/mod.rs#L69)). However, the feedback provided by the failing tests doesn't let me pinpoint the root cause very easily. Since I don't know the codebase very well, I'd be happy if someone in the know could take a look.

Additionally, I'm not sure if my handling of bytes is correct in all places. For example (but not limited to): my current implementation of the `Hasher`'s trait method `fn hash(bytes: &[u8])` interprets 8 bytes as a field element of the field with $2^{64} - 2^{32} + 1$ elements[^0]. This can lead to hash collisions, depending on the type of objects being fed into the method. I'd appreciate some guidance here.

[^0]: This is the only field the hash function Tip5 is defined over, so there is no need to be generic with respect to the field.

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -33,6 +33,8 @@ std = ["blake3/std", "math/std", "sha3/std", "utils/std"]
 blake3 = { version = "1.3", default-features = false }
 math = { version = "0.6", path = "../math", package = "winter-math", default-features = false }
 sha3 = { version = "0.10", default-features = false }
+twenty-first = "0.19.3"
+rand = "0.8.5"
 utils = { version = "0.6", path = "../utils/core", package = "winter-utils", default-features = false }
 
 [dev-dependencies]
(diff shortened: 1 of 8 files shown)


ameya-deshmukh:
Hi! I'll try hacking on this and will let you know if tests pass for me. In addition to this, happy to reproduce and review this PR

Nashtare:
I don't think it makes sense to increase the size of Digest bytes for all existing instantiations, as this need comes solely from the specific version of Tip5 you want to implement. Instead I'd suggest having the output length defined by an associated constant.

jan-ferdinand:
@Nashtare, I agree. However, taken straight from the description of the `Digest` trait:

> Ideally, the length of the returned array should be defined by an associated constant, but using associated constants in const generics is not supported by Rust yet. Thus, we put an upper limit on the possible digest size. For digests which are smaller than 32 [now 40] bytes, the unused bytes should be set to 0.

Refactoring the trait was beyond what I was willing to invest into this PR for the time being.

PR #191, opened 2023-04-25
Make changes to enable custom build_constraint_commitment

andrewmilson:
Changes:
* Made CompositionPoly and ConstraintCommitment public to allow overriding build_constraint_commitment
* Removed `.evaluate` from composition poly and use RowMatrix::evaluate_polys_over instead (this is now consistent with build_trace_commitment)
* ConstraintCommitment takes a RowMatrix of evaluations instead of a ColMatrix

### DIFF
--- a/prover/src/lib.rs
+++ b/prover/src/lib.rs
@@ -79,7 +79,8 @@ pub mod matrix;
 pub use matrix::{ColMatrix, RowMatrix};
 
 mod constraints;
-use constraints::{CompositionPoly, ConstraintCommitment, ConstraintEvaluator};
+use constraints::ConstraintEvaluator;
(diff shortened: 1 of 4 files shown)


prover/src/lib.rs
```diff
@@ -517,7 +518,10 @@ pub trait Prover {
         // evaluate composition polynomial columns over the LDE domain
         #[cfg(feature = "std")]
         let now = Instant::now();
```
andrewmilson, line 520: Hey @irakliyk. What do you think of removing build_constraint_commitment and just using build_trace_commitment for it instead?

irakliyk (approved):
Looks good! Thank you!

PR #192, opened 2023-04-25
Remove build_constraint_commitment and use build_trace_commitment instead

andrewmilson:


irakliyk:
Closing this as no longer compatible with the current structure.

Issue #195, opened 2023-05-06
Simplifying constraint composition polynomial


Al-Kindi-0:
In Winterfell, the constraint composition polynomial is constructed using the following formula

$$
\sum_{i=0}^{k-1}  \left(\alpha_i + \beta_i \cdot X^{d_i}\right) \cdot \frac{C_i(X)}{Z_{H_i}(X)}
$$

where:

1. $(\alpha_i, \beta_i)$ is a per-constraint random tuple.
2. $d_i$ is what is called the adjustment factor. This is needed in order to make each of the terms in the above expression have the same degree.
3. $C_i$ is the i-th constraint.
4. $Z_{H_i}$ is the vanishing polynomial on the sub-group $H_i$ where the constraint $C_i$ should hold.

In fact, in Winterfell the constraints are grouped into groups that share the same $d_i$ in order to optimize things.
The use of degree adjustments made sense in earlier versions of the protocol i.e. the original ALI protocol. Moreover, and as noticed [here](https://eprint.iacr.org/2022/1216) in page 16,  we can compute the constraint composition polynomial as 

$$
\mathbb{C}(X) := \sum_{i=0}^{k-1}  \delta_i \cdot \frac{C_i(X)}{Z_{H_i}(X)}
$$

For all sensible choices of security parameters, choosing $\delta_i := \delta^i$ does not degrade soundness as this is usually not the most dominiant part.

Related to this is how $\mathbb{C}(X)$ is divided into `d:= max_degree - 1` many polynomials. [In](https://eprint.iacr.org/2022/1216) page 15, the following decomposition is proposed

$$
\mathbb{C}(X) = h_0(X) + X^{|H|}\cdot h_1(X) + \cdots + X^{(d-1)\cdot|H|}\cdot h_{d-1}(X)
$$

where $|H|$ is the length of the trace.
This decomposition has the advantage of giving a slightly better soundness error as well as avoiding having quotients with denominator  $(X - z^d)$.
The following issue aims at discussing the merits of introducing such changes in the Winterfell setting as well as the best approach to implement them if we decide to go this route.

irakliyk:
This is very interesting! If we implement this, this would simplify the code base quite a bit.

[...]

PR #198, opened 2023-05-09
Simplify constraint composition by removing adjustment factors

Al-Kindi-0:
Addressing #195

### DIFF
--- a/air/src/air/coefficients.rs
+++ b/air/src/air/coefficients.rs
@@ -51,22 +51,21 @@ impl<E: FieldElement> Default for AuxTraceRandElements<E> {
 /// function. In the interactive version of the protocol, the verifier draws these coefficients
 /// uniformly at random from the extension field of the protocol.
 ///
-/// There are two coefficients for each constraint so that we can compute a random linear
-/// combination of constraints like so:
+/// There is one coefficient for each constraint so that we can compute a random linear
+/// combination of constraints as:
 /// $$
-/// \sum_{i = 0}^k{C_i(x) \cdot (\alpha_i + \beta_i \cdot x^{d_i})}
+/// \sum_{i = 0}^k{\alpha_i \cdot C_i(x)}
 /// $$
 /// where:
-/// * $\alpha_i$ and $\beta_i$ are the coefficients for the $i$th constraint.
+/// * $\alpha_i$ is the coefficient for the $i$th constraint.
 /// * $C_i(x)$ is an evaluation of the $i$th constraint at $x$.
-/// * $d_i$ is the degree adjustment factor needed to normalize all constraints to the same degree.
 ///
 /// The coefficients are separated into two lists: one for transition constraints and another one
 /// for boundary constraints. This separation is done for convenience only.
 #[derive(Debug, Clone)]
 pub struct ConstraintCompositionCoefficients<E: FieldElement> {
-    pub transition: Vec<(E, E)>,
-    pub boundary: Vec<(E, E)>,
+    pub transition: Vec<E>,
+    pub boundary: Vec<E>,
 }
 
 // DEEP COMPOSITION COEFFICIENTS
@@ -84,7 +83,7 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
 /// Y(x) = \sum_{i=0}^k{(
 ///     \alpha_i \cdot \frac{T_i(x) - T_i(z)}{x - z} +
 ///     \beta_i \cdot \frac{T_i(x) - T_i(z \cdot g)}{x - z \cdot g}
-/// )} + \sum_{j=0}^m{\delta \cdot \frac{H_j(x) - H_j(z^m)}{x - z^m}}
+/// )} + \sum_{j=0}^m{\delta_j \cdot \frac{H_j(x) - H_j(z)}{x - z}}
 /// $$
 /// where:
 /// * $z$ is an out-of-domain point drawn randomly from the entire field. In the interactive
@@ -94,9 +93,7 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
 /// * $T_i(x)$ is an evaluation of the $i$th trace polynomial at $x$, and $k$ is the total
 ///   number of trace polynomials (which is equal to the width of the execution trace).
 /// * $H_i(x)$ is an evaluation of the $j$th constraint composition column polynomial at $x$,
-///   and $m$ is the total number of column polynomials. The number of column polynomials is equal
-///   to the highest constraint degree rounded to the next power of two. For example, if the
-///   highest constraint degree is 6, $m$ will be equal to 8.
+///   and $m$ is the total number of column polynomials.
 /// * $\alpha_i, \beta_i$ are composition coefficients for the $i$th trace polynomial.
 /// * $\delta_j$ is a composition coefficient for $j$th constraint column polynomial.
 ///
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -4,6 +4,7 @@
 // LICENSE file in the root directory of this source tree.
 
 use crate::{air::TransitionConstraintDegree, ProofOptions, TraceInfo};
+use core::cmp;
 use math::StarkField;
 use utils::collections::Vec;
 
@@ -184,14 +185,6 @@ impl<B: StarkField> AirContext<B> {
         self.trace_info.length() * self.ce_blowup_factor
     }
 
-    /// Returns the degree to which all constraint polynomials are normalized before they are
-    /// composed together.
-    ///
-    /// This degree is always `ce_domain_size` - 1.
-    pub fn composition_degree(&self) -> usize {
-        self.ce_domain_size() - 1
-    }
-
     /// Returns the size of the low-degree extension domain.
     ///
     /// This is guaranteed to be a power of two, and is equal to `trace_length * lde_blowup_factor`.
@@ -237,6 +230,50 @@ impl<B: StarkField> AirContext<B> {
         self.num_transition_exemptions
     }
 
+    /// Returns the number of columns needed to store the constraint composition polynomial.
+    ///
+    /// This is the maximum of:
+    /// 1. The maximum evaluation degree over all transition constraints minus the degree
+    /// of the transition constraint divisor divided by trace length.
+    /// 2. `1`, because the constraint composition polynomial requires at least one column.
+    ///
+    /// Since the degree of a constraint `C(x)` can be well approximated by
+    /// `[constraint.base + constraint.cycles.len()] * [trace_length - 1]` the degree of the
+    /// constraint composition polynomial can be approximated by:
+    /// `([constraint.base + constraint.cycles.len()] * [trace_length - 1] - [trace_length - n])`
+    /// where `constraint` is the constraint attaining the maximum and `n` is the number of
+    /// exemption points.
+    /// In the case `n = 1`, the expression simplifies to:
+    /// `[constraint.base + constraint.cycles.len() - 1] * [trace_length - 1]`
+    /// Thus, if each column is of length `trace_length`, we would need
+    /// `[constraint.base + constraint.cycles.len() - 1]` columns to store the coefficients of
+    /// the constraint composition polynomial.
+    /// This means that if the highest constraint degree is equal to `5`, the constraint
+    /// composition polynomial will require four columns and if the highest constraint degree is
+    /// equal to `7`, it will require six columns to store.
+    pub fn num_constraint_composition_columns(&self) -> usize {
+        let mut highest_constraint_degree = 0_usize;
+        for degree in self
+            .main_transition_constraint_degrees
+            .iter()
+            .chain(self.aux_transition_constraint_degrees.iter())
+        {
+            let eval_degree = degree.get_evaluation_degree(self.trace_len());
+            if eval_degree > highest_constraint_degree {
+                highest_constraint_degree = eval_degree
+            }
+        }
+        let trace_length = self.trace_len();
+        let transition_divisior_degree = trace_length - self.num_transition_exemptions();
+
+        // we use the identity: ceil(a/b) = (a + b - 1)/b
+        let num_constraint_col =
+            (highest_constraint_degree - transition_divisior_degree + trace_length - 1)
+                / trace_length;
+
+        cmp::max(num_constraint_col, 1)
+    }
+
     // DATA MUTATORS
     // --------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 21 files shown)


Al-Kindi-0:
One thing that remains to be addressed is removing `AirContext::composition_degree` as there is no need for it with the new modifications. The reason this is not included in the latest commit is due to the dependence of `AirContext::set_num_transition_exemptions` on it and the fact that I am not 100% sure how to modify this in our current context.

prover/src/constraints/evaluator.rs
```diff
@@ -273,13 +265,11 @@ impl<'a, A: Air, E: FieldElement<BaseField = A::BaseField>> ConstraintEvaluator<
...
-        self.transition_constraints.main_constraints().iter().fold(E::ZERO, |result, group| {
-            let (power, offset_exp) = (group.degree_adjustment(), group.domain_offset_exp());
-            let xp = domain.get_ce_x_power_at(step, power, offset_exp);
-            result + group.merge_evaluations(evaluations, xp)
-        })
+       // merge constraint evaluations for the main trace segment
+        let result = evaluations.iter().zip(self.transition_constraints.main_constraint_coef().iter()).fold(E::ZERO, |acc, (&const_eval, &coef)| {
+            acc + coef.mul_base(const_eval)
+        });
+        result
```
irakliyk, line 272: minor nit: I'd improve formatting here - maybe run `rustfmt` manually as it doesn't run for this function by default.

Also:
- I'd keep the previous comment.
- I don't think we need to introduce the `result` variable. We can probably return the result of folding directly.

prover/src/constraints/evaluator.rs
```diff
@@ -314,11 +303,10 @@ impl<'a, A: Air, E: FieldElement<BaseField = A::BaseField>> ConstraintEvaluator<
...
         // we can do this here because all transition constraints have the same divisor.
-        self.transition_constraints.aux_constraints().iter().fold(E::ZERO, |result, group| {
-            let (power, offset_exp) = (group.degree_adjustment(), group.domain_offset_exp());
-            let xp = domain.get_ce_x_power_at(step, power, offset_exp);
-            result + group.merge_evaluations::<E::BaseField, E>(evaluations, xp)
-        })
+        let result = evaluations.iter().zip(self.transition_constraints.aux_constraint_coef().iter()).fold(E::ZERO, |acc, (&const_eval, &coef)| {
+            acc + coef * const_eval
+        });
+        result
```
irakliyk, line 309: minor nit: similar comment re formatting and `result` variable as above.

prover/src/constraints/evaluation_table.rs
```diff
@@ -164,11 +164,6 @@ impl<'a, E: FieldElement> ConstraintEvaluationTable<'a, E> {
         // by the evaluations of its corresponding divisor, and add all resulting evaluations
         // together into a single vector
         for (column, divisor) in self.evaluations.into_iter().zip(self.divisors.iter()) {
-            // in debug mode, make sure post-division degree of each column matches the expected
-            // degree
-            #[cfg(debug_assertions)]
-            validate_column_degree(&column, divisor, self.domain, column.len() - 1)?;
```
irakliyk, line 170: is this check now meaningless?

prover/src/constraints/composition_poly.rs
```diff
@@ -90,22 +101,27 @@ impl<E: FieldElement> CompositionPoly<E> {
...
+/// the two columns will be: (c * x + d) and (a * x + b).
+fn segment<E: FieldElement>(coefficients: Vec<E>, trace_len: usize) -> Vec<Vec<E>> {
+    let degree_plus_1 = degree_of(&coefficients) + 1;
+    let remainder = degree_plus_1 % trace_len;
+
+    let num_columns = if remainder == 0 {
+        degree_plus_1 / trace_len
+    } else {
+        (degree_plus_1 / trace_len) + 1
     };
```
irakliyk, line 114: Is there an easy way to compute the number of columns without using `degree_of()`? This function will scan memory from the end of the vector until a non-zero element is found. Not sure if it'll have significant impact on performance, but if there is an easy way to avoid it (i.e., buy reading number of columns from AIR context), we should do it).

prover/src/constraints/composition_poly.rs
```diff
@@ -90,22 +101,27 @@ impl<E: FieldElement> CompositionPoly<E> {
...
     // TODO: implement multi-threaded version
-    for (i, coeff) in coefficients.into_iter().enumerate() {
-        let row_idx = i / num_columns;
-        let col_idx = i % num_columns;
-        result[col_idx][row_idx] = coeff;
+    for (i, coeff) in coefficients.iter().take(degree_plus_1).enumerate() {
+        let col_idx = i / trace_len;
+        let row_idx = i % trace_len;
+        result[col_idx][row_idx] = *coeff;
     }
```
irakliyk, line 125: Ideally, we should be able to do this without re-allocating, but that may require unsafe code. For now, I wonder if we can use [split_off()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off) method and leave a TODO to come back and figure-out zero-allocation approach later.

prover/src/constraints/composition_poly.rs
```diff
@@ -67,11 +60,29 @@ impl<E: FieldElement> CompositionPoly<E> {
...
+        assert_eq!(
+            self.column_len(),
+            domain.trace_length(),
+            "inconsistent trace domain size; expected {}, but received {}",
+            self.column_len(),
+            domain.trace_length()
+        );
+
+        self.data.evaluate_columns_over(domain)
+    }
```
irakliyk, line 81: Where is this method being used? (I couldn't find upon a quick search).

irakliyk (changes requested):
This looks great (and simplifies things quite a bit)! Thank you! I left a few comments inline.

air/src/air/context.rs
```diff
@@ -237,6 +239,34 @@ impl<B: StarkField> AirContext<B> {
         self.num_transition_exemptions
     }
 
+    /// Returns the number of of columns needed to store the constraint composition polynomial.
+    /// This is the maximum of:
+    /// 1- The maximum evaluation degrees over all transition constraints minus the degree
+    /// of the transition constraint divisor.
+    /// 2- 1, because the constraint composition polynomial requires at least one column.
+    pub fn num_constraint_col(&self) -> usize {
```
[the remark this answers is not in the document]
Al-Kindi-0: Makes sense. I have tried to explain the logic for the way things are done (although it is a bit more difficult compared to how it was done previously).

air/src/air/context.rs
```diff
@@ -3,6 +3,8 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use core::cmp;
+
 use crate::{air::TransitionConstraintDegree, ProofOptions, TraceInfo};
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

verifier/src/lib.rs
```diff
@@ -202,16 +202,17 @@ where
     public_coin.reseed(H::hash_elements(ood_trace_frame.values()));
 
     // read evaluations of composition polynomial columns sent by the prover, and reduce them into
-    // a single value by computing sum(z^i * value_i), where value_i is the evaluation of the ith
-    // column polynomial at z^m, where m is the total number of column polynomials; also, reseed
+    // a single value by computing sum(z^(i * l) * value_i), where value_i is the evaluation of
+    // the ith column polynomial at z, where l is length of the trace; also, reseed
     // the public coin with the OOD constraint evaluations received from the prover.
```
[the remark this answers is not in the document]
Al-Kindi-0: Makes sense, changed!

air/src/air/coefficients.rs
```diff
@@ -54,19 +54,18 @@ impl<E: FieldElement> Default for AuxTraceRandElements<E> {
 /// There are two coefficients for each constraint so that we can compute a random linear
 /// combination of constraints like so:
 /// $$
-/// \sum_{i = 0}^k{C_i(x) \cdot (\alpha_i + \beta_i \cdot x^{d_i})}
+/// \sum_{i = 0}^k{\alpha_i \cdot C_i(x)}
```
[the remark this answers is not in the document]
Al-Kindi-0: Thanks

air/src/air/mod.rs
```diff
@@ -512,12 +510,12 @@ pub trait Air: Send + Sync {
     {
         let mut t_coefficients = Vec::new();
         for _ in 0..self.context().num_transition_constraints() {
-            t_coefficients.push(public_coin.draw_pair()?);
+            t_coefficients.push(public_coin.draw()?);
```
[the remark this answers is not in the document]
Al-Kindi-0: Once we remove the final FRI-related degree adjustment, we can remove it.

air/src/air/transition/mod.rs
```diff
@@ -23,13 +23,15 @@ const MIN_CYCLE_LENGTH: usize = 2;
...
 pub struct TransitionConstraints<E: FieldElement> {
-    main_constraints: Vec<TransitionConstraintGroup<E>>,
+    //main_constraints: Vec<TransitionConstraintGroup<E>>,
+    main_constraint_coef: Vec<E>,
     main_constraint_degrees: Vec<TransitionConstraintDegree>,
-    aux_constraints: Vec<TransitionConstraintGroup<E>>,
+    //aux_constraints: Vec<TransitionConstraintGroup<E>>,
+    aux_constraint_coef: Vec<E>,
     aux_constraint_degrees: Vec<TransitionConstraintDegree>,
     divisor: ConstraintDivisor<E::BaseField>,
```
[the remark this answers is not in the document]
Al-Kindi-0: Yes, thank you!

prover/src/constraints/evaluator.rs
Al-Kindi-0: Done

prover/src/constraints/evaluation_table.rs
Al-Kindi-0: When we had the degree adjustments we were always assured that the expected degree is `column.len() -1`. Now that we have removed those we need to computed the expected degree using the highest evaluation degree of the constraints in a group.

prover/src/constraints/composition_poly.rs
Al-Kindi-0: I agree, removed it now

prover/src/constraints/composition_poly.rs
Al-Kindi-0: ~The problem with using number of columns is that `degree_plus_1` doesn't divide evenly by `trace_len` and hence we cant replace `for (i, coeff) in coefficients.iter().take(degree_plus_1).enumerate()` with `for (i, coeff) in coefficients.iter().take(num_cols * trace_len).enumerate()`.~
Actually, it will work, provided we initialize the result matrix with `E::ZERO`.
I have made an attempt at using the AIR context, let me know if it does the job.

Al-Kindi-0: Actually it will work even without intialization using your suggestion with `split_off`

prover/src/constraints/composition_poly.rs
Al-Kindi-0: What about using slices like
```rust
let mut result = Vec::new();
    for i in (0..coefficients.len())
        .take(num_cols * trace_len)
        .step_by(trace_len)
    {
        let end = i + trace_len;
        let slice = &coefficients[i..end];
        result.push(slice.to_vec());
    }
```

irakliyk: The above works! I wonder if we can simplify it like this:
```Rust
coefficients
    .chunks(trace_len)
    .take(num_cols)
    .map(|slice| slice.to_vec())
    .collect()
```

irakliyk: Also, should we do something like:
```Rust
debug_assert!(degree_of(coefficients) < trace_len * nu_cols - 1);
```

air/src/air/context.rs
```diff
@@ -237,6 +238,48 @@ impl<B: StarkField> AirContext<B> {
...
+    /// where `constraint` is the constraint attaining the maximum and `n` is the number of
+    /// exemption points.
+    /// In the case `n = 1`, the expression simplifies to:
+    /// `[constraint.base + constraint.cycles.len() - 1] * [trace_length - 1]`
+    /// Thus, if each column is of length `trace_length`, we would need
+    /// `[constraint.base + constraint.cycles.len() - 1]` columns to store the coefficients of
+    /// the constraint composition polynomial.
+    /// This means that if the highest constraint degree is equal to `5`, the constraint
+    /// composition polynomial will require four columns and if the highest constraint degree is
+    /// equal to `7`, it will require six columns to store.
```
irakliyk, line 258: minor nit: Should we move this explanation below the second point below? For example, something like this:
```
/// 1. The maximum evaluation degree over all transition constraints minus the degree
/// of the transition constraint divisor divided by trace length.
/// 2. `1`, because the constraint composition polynomial requires at least one column.
/// 
/// Since the degree of a constraint `C(x)` can be well approximated by
/// ...
```

prover/src/constraints/composition_poly.rs
Al-Kindi-0: Makes sense!

air/src/air/context.rs
Al-Kindi-0: Done

prover/src/constraints/composition_poly.rs
```diff
@@ -67,11 +60,12 @@ impl<E: FieldElement> CompositionPoly<E> {
         self.column_len() - 1
     }
 
-    /// Returns evaluations of all composition polynomial columns at point z^m, where m is
-    /// the number of column polynomials.
+    // LOW-DEGREE EXTENSION
+    // --------------------------------------------------------------------------------------------
```
irakliyk, line 64: nit: we can probably delete this.

air/src/air/context.rs
```diff
@@ -237,6 +230,50 @@ impl<B: StarkField> AirContext<B> {
         self.num_transition_exemptions
     }
 
+    /// Returns the number of columns needed to store the constraint composition polynomial.
+    /// This is the maximum of:
+    ///
+    /// 1. The maximum evaluation degree over all transition constraints minus the degree
```
irakliyk, line 236: minor nit: I'd put the blank line between the first two lines, like so:
```
/// Returns the number of columns needed to store the constraint composition polynomial.
///
/// This is the maximum of:
/// 1. The maximum evaluation degree over all transition constraints minus the degree
```
The reasons is that the first line will be used as the method summary in the docs and if we don't have this blank line "This is the maximum of:" would be included in the method summary and it will look incomplete.

irakliyk (approved):
All looks good! Thank you! I left a couple of doc-related nits inline.

air/src/air/context.rs
Al-Kindi-0: Thanks!
I have fixed it now

PR #199, opened 2023-05-12
Compactify serialization of ood frame

Al-Kindi-0:
This PR introduces a small optimization that compactifies the OOD frame when passed through the channels. This leads to a reduction in the number of required hashes needed for reseeding the random coin which should make a difference in the setting of recursive proofs.

### DIFF
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -3,7 +3,6 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-use crate::EvaluationFrame;
 use math::FieldElement;
 use utils::{
     collections::Vec, ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
@@ -13,7 +12,7 @@ use utils::{
 // TYPE ALIASES
 // ================================================================================================
 
-type ParsedOodFrame<E> = (EvaluationFrame<E>, Option<EvaluationFrame<E>>, Vec<E>);
+type ParsedOodFrame<E> = (Vec<E>, Vec<E>);
 
 // OUT-OF-DOMAIN FRAME
 // ================================================================================================
@@ -38,18 +37,33 @@ impl OodFrame {
     // UPDATERS
     // --------------------------------------------------------------------------------------------
 
-    /// Updates the trace state portion of this out-of-domain frame.
+    /// Updates the trace state portion of this out-of-domain frame. This also returns a compactified
+    /// version of the out-of-domain frame with the rows interleaved. This is done so that reseeding
+    /// of the random coin needs to be done only once as opposed to once per each row.
     ///
     /// # Panics
     /// Panics if evaluation frame has already been set.
-    pub fn set_trace_states<E: FieldElement>(&mut self, trace_states: &[Vec<E>]) {
+    pub fn set_trace_states<E: FieldElement>(&mut self, trace_states: &[Vec<E>]) -> Vec<E> {
         assert!(
             self.trace_states.is_empty(),
             "trace sates have already been set"
         );
-        for trace_state in trace_states {
-            trace_state.write_into(&mut self.trace_states);
+
+        // save the evaluations with the current and next evaluations interleaved for each polynomial
+        let frame_size = trace_states.len();
+        let width = trace_states[0].len();
+
+        let mut result = vec![];
+        for i in 0..width {
+            for row in trace_states.iter() {
+                result.push(row[i]);
+            }
         }
+        debug_assert!(frame_size <= u8::MAX as usize);
+        self.trace_states.write_u8(frame_size as u8);
+        result.write_into(&mut self.trace_states);
+
+        result
     }
 
     /// Updates constraint evaluation portion of this out-of-domain frame.
@@ -94,23 +108,16 @@ impl OodFrame {
         assert!(main_trace_width > 0, "trace width cannot be zero");
         assert!(num_evaluations > 0, "number of evaluations cannot be zero");
 
-        // parse current and next trace states for main and auxiliary trace evaluation frames
+        // parse main and auxiliary trace evaluation frames
         let mut reader = SliceReader::new(&self.trace_states);
-        let current = E::read_batch_from(&mut reader, main_trace_width)?;
-        let current_aux = E::read_batch_from(&mut reader, aux_trace_width)?;
-        let next = E::read_batch_from(&mut reader, main_trace_width)?;
-        let next_aux = E::read_batch_from(&mut reader, aux_trace_width)?;
+        let frame_size = reader.read_u8()? as usize;
+        let trace = E::read_batch_from(
+            &mut reader,
+            (main_trace_width + aux_trace_width) * frame_size,
+        )?;
         if reader.has_more_bytes() {
             return Err(DeserializationError::UnconsumedBytes);
         }
-
-        // instantiate the frames from the parsed rows
-        let main_frame = EvaluationFrame::from_rows(current, next);
-        let aux_frame = if aux_trace_width > 0 {
-            Some(EvaluationFrame::from_rows(current_aux, next_aux))
-        } else {
-            None
-        };
 
         // parse the constraint evaluations
         let mut reader = SliceReader::new(&self.evaluations);
@@ -119,7 +126,7 @@ impl OodFrame {
             return Err(DeserializationError::UnconsumedBytes);
         }
 
-        Ok((main_frame, aux_frame, evaluations))
+        Ok((trace, evaluations))
     }
 }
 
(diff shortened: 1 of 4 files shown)


air/src/proof/ood_frame.rs
```diff
@@ -42,14 +47,12 @@ impl OodFrame {
...
-    pub fn set_trace_states<E: FieldElement>(&mut self, trace_states: &[Vec<E>]) {
+    pub fn set_trace_states<E: FieldElement>(&mut self, trace_states: Vec<E>) {
         assert!(
             self.trace_states.is_empty(),
             "trace sates have already been set"
         );
-        for trace_state in trace_states {
-            trace_state.write_into(&mut self.trace_states);
-        }
+        trace_states.write_into(&mut self.trace_states);
```
irakliyk, line 55: I wonder if we should keep the old interface and handle interleaved serialization here instead. We could also return the interleaved result from here so that in the prover channel we could just call `reseed()` with it. So, the signature could look something like this:
```Rust
pub fn set_trace_states<E: FieldElement>(&mut self, trace_states: &[Vec<E>]) -> Vec<E>
```

Also, maybe we should serialize the frame size as the first value - this way, we won't need to rely on `FRAME_SIZE` constant.

prover/src/channel.rs
```diff
@@ -86,10 +86,20 @@ where
...
+        debug_assert_eq!(trace_states.len(), FRAME_SIZE);
+
+        // save the evaluations with the current and next evaluations interleaved for each polynomial
+        let width = trace_states[0].len();
+        let mut result = vec![E::ZERO; width * FRAME_SIZE];
+        for i in 0..width {
+            for j in 0..FRAME_SIZE {
+                result[i * FRAME_SIZE + j] = trace_states[j][i];
+            }
         }
```
irakliyk, line 98: I'd probably do this as something like:
```Rust
let size = trace_states.len();
let width = trace_states[0].len();
let mut result = Vec::new();
for i in 0..width {
    for j in 0..size {
        result.push(trace_states[j][i]);
    }
}
```
This way we don't need to tie the code to `FRAME_SIZE` constant.

verifier/src/lib.rs
```diff
@@ -202,17 +203,35 @@ where
...
+        ];
+        let width = trace_states[0].len();
+        let mut result = vec![E::ZERO; width * FRAME_SIZE];
+        for i in 0..width {
+            for j in 0..FRAME_SIZE {
+                result[i * FRAME_SIZE + j] = trace_states[j][i];
+            }
+        }
+        public_coin.reseed(H::hash_elements(&result));
     }
```
irakliyk, line 235: I wonder if this could be somehow done inside `OodFrame` struct. I don't have a great suggestion how to do this yet, but I'll try to come up with something.

irakliyk (commented):
Thank you! Looks good! This is not a finished review yet - but I left a few comments inline.

air/src/proof/ood_frame.rs
Al-Kindi-0: Makes sense. I have changed it accordingly.

prover/src/channel.rs
Al-Kindi-0: Done!

verifier/src/lib.rs
Al-Kindi-0: Agree, all options I could come up with are either a bit "hacky" or require significant changes to some of the structs.

irakliyk: One potential solution (though still not ideal) is to put this logic into [TraceOodFrame](https://github.com/facebook/winterfell/blob/main/verifier/src/channel.rs#L346) struct. Basically, we could change this strict to look something like:

```Rust
pub struct TraceOodFrame<E: FieldElement> {
    values: Vec<E>,
    main_trace_width: usize,
    aux_trace_width: usize,
}

impl<E: FieldElement> TraceOodFrame<E> {
    pub fn new(values: Vec<E>, main_trace_width: usize, aux_trace_width: usize) -> Self {
        Self {
            values,
            main_trace_width,
            aux_trace_width,
        }
    }

    pub fn values(&self) -> &[E] {
        &self.values
    }

    pub fn main_frame(&self) -> EvaluationFrame<E> {
        todo!("build the main frame here from relevant parts of self.values")
    }

    pub fn aux_frame(&self) -> Option<EvaluationFrame<E>> {
        if self.aux_trace_width == 0 {
            None
        } else {
            todo!("build the aux frame here from relevant parts of self.values")        
        }
    }
}
```

Then, we'd update [read_ood_trace_frame()](https://github.com/facebook/winterfell/blob/main/verifier/src/channel.rs#L143) method to look as follows:
```Rust
pub fn read_ood_trace_frame(&mut self) -> TraceOodFrame {
    let frame = self.ood_trace_frame.take().expect("already read")
}
```

And finally, we'd update [parse()](https://github.com/facebook/winterfell/blob/main/air/src/proof/ood_frame.rs#L88) method of `OodFrame` to return just two vectors: one with all trace values and the other one with constraint evaluations. The first vector would then be passed to `TraceOodFrame` as a construction parameters.

We still have the logic split between two places - but I feel it would be a bit better encapsulated.

verifier/src/lib.rs
```diff
@@ -188,7 +188,9 @@ where
     // read the out-of-domain trace frames (the main trace frame and auxiliary trace frame, if
     // provided) sent by the prover and evaluate constraints over them; also, reseed the public
     // coin with the OOD frames received from the prover.
-    let (ood_main_trace_frame, ood_aux_trace_frame) = channel.read_ood_trace_frame();
+    let ood_trace_frame = channel.read_ood_trace_frame();
+    let ood_main_trace_frame = ood_trace_frame.main_frame();
+    let ood_aux_trace_frame = ood_trace_frame.aux_frame();
```
irakliyk, line 193: minor nit: we probably don't need to introduce variables for these and make calls directly in the lines below like so:
```Rust
let ood_constraint_evaluation_1 = evaluate_constraints(
    &air,
    constraint_coeffs,
    &ood_trace_frame.main_frame(),
    &ood_trace_frame.aux_frame(),
    aux_trace_rand_elements,
    z,
);
```

air/src/proof/ood_frame.rs
```diff
@@ -38,18 +37,32 @@ impl OodFrame {
...
+        let frame_size = trace_states.len();
+        let width = trace_states[0].len();
+
+        let mut result = vec![];
+        for i in 0..width {
+            for row in trace_states.iter() {
+                result.push(row[i]);
+            }
         }
+        self.trace_states.write_u8(frame_size as u8);
```
irakliyk, line 62: nit: should we put a `debug_assert` here to check that the frame size can fit into u8? Maybe something like:
```Rust
debug_assert!(frame_size <= u8::MAX as usize);
```

verifier/src/channel.rs
```diff
@@ -343,16 +342,59 @@ impl<E: FieldElement, H: ElementHasher<BaseField = E::BaseField>> ConstraintQuer
...
+            main_trace_width,
+            aux_trace_width,
+        }
+    }
+
+    pub fn values(&self) -> &[E] {
+        &self.values
+    }
+
+    pub fn main_frame(&self) -> EvaluationFrame<E> {
```
irakliyk, line 364: minor nit: maybe add a brief comment explaining why we do it this way (and maybe mentioning `OodFrame` so that people can easily find where the "other side" of this code is)?

irakliyk (approved):
Looks good! Thank you! I left a couple of minor comments inline. We can merge after these are addressed.

verifier/src/lib.rs
irakliyk: Ah! Then let's keep it the way you have it.

air/src/proof/ood_frame.rs
Al-Kindi-0: Done!

verifier/src/channel.rs
Al-Kindi-0: Let me know if the comments makes sense or if they are too much verbose.

air/src/proof/ood_frame.rs
Al-Kindi-0: Just a curiosity: What is the best way, in your opinion, to use newlines to make the following code more readable?

```rust

        let frame_size = trace_states.len();
        let width = trace_states[0].len();

        let mut result = vec![];
        for i in 0..width {
            for row in trace_states.iter() {
                result.push(row[i]);
            }
        }
        debug_assert!(frame_size <= u8::MAX as usize);
        self.trace_states.write_u8(frame_size as u8);
        result.write_into(&mut self.trace_states);
```

irakliyk: I think the above is fine. Maybe I would have added an extra newline like so:
```Rust
let frame_size = trace_states.len();
let width = trace_states[0].len();

let mut result = vec![];
for i in 0..width {
    for row in trace_states.iter() {
        result.push(row[i]);
    }
}

debug_assert!(frame_size <= u8::MAX as usize);
self.trace_states.write_u8(frame_size as u8);
result.write_into(&mut self.trace_states);
```
But, in my opinion, the difference is marginal at best.

verifier/src/channel.rs
irakliyk: This looks awesome! Thank you!

PR #203, opened 2023-05-24
Simplify DEEP composition polynomial

Al-Kindi-0:
Removes the degree adjustment performed on the DEEP composition polynomial. This leads to a negligible increase in the soundness error of the protocol i.e. the $+2$ in the formula for $k^+$. The new soundness error for the current protocol is given by:

$$
L^{+}\cdot\left(\frac{1}{|\mathbb{F}|} + \frac{d\cdot(k^+ - 1) + (k-1)}{|\mathbb{F}| - |D\cup H|}\right) + \epsilon_{FRI}
$$

where:

1. $H$ is the trace domain of size $k$
2. $D$ is the LDE domain of size $n$
3. $k^+ := k + 2$
4. $L^+ := \frac{m+0.5}{\sqrt{\rho^+}}$ where $m\geq 3$ is the Johnson proximity parameter which for simiplicity can be taken to be equal to $3$ and $\rho^+ := \frac{k^+}{n}$
5. $\mathbb{F}$ is the extension field.
6. $\epsilon_{FRI}$ is the soundness error bound for FRI run with proximity parameter $\theta^+ := 1 - \alpha^+ := 1 - (1 + \frac{1}{2m})\cdot \sqrt{\rho^+}$
7. $d$ is the number of constraint composition column polynomials.

### DIFF
--- a/air/src/air/coefficients.rs
+++ b/air/src/air/coefficients.rs
@@ -77,39 +77,40 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
 /// function. In the interactive version of the protocol, the verifier draws these coefficients
 /// uniformly at random from the extension field of the protocol.
 ///
-/// The coefficients are used in computing the DEEP composition polynomial in two steps. First,
-/// we compute a random linear combination of trace and constraint composition polynomials as:
+/// The coefficients are used in computing the DEEP composition polynomial as:
 /// $$
 /// Y(x) = \sum_{i=0}^k{(
-///     \alpha_i \cdot \frac{T_i(x) - T_i(z)}{x - z} +
-///     \beta_i \cdot \frac{T_i(x) - T_i(z \cdot g)}{x - z \cdot g}
-/// )} + \sum_{j=0}^m{\delta_j \cdot \frac{H_j(x) - H_j(z)}{x - z}}
+///     \alpha_i \cdot (\frac{T_i(x) - T_i(z)}{x - z} +
+///     \frac{T_i(x) - T_i(z \cdot g)}{x - z \cdot g})
+/// )} + \sum_{j=0}^m{\beta_j \cdot \frac{H_j(x) - H_j(z)}{x - z}}
 /// $$
 /// where:
 /// * $z$ is an out-of-domain point drawn randomly from the entire field. In the interactive
 ///   version of the protocol, $z$ is provided by the verifier.
-/// * $g$ is the generator of the trace domain. This is the same as $n$th root of unity where
+/// * $g$ is the generator of the trace domain. This is the $n$th root of unity where
 ///   $n$ is the length of the execution trace.
 /// * $T_i(x)$ is an evaluation of the $i$th trace polynomial at $x$, and $k$ is the total
 ///   number of trace polynomials (which is equal to the width of the execution trace).
 /// * $H_i(x)$ is an evaluation of the $j$th constraint composition column polynomial at $x$,
 ///   and $m$ is the total number of column polynomials.
-/// * $\alpha_i, \beta_i$ are composition coefficients for the $i$th trace polynomial.
-/// * $\delta_j$ is a composition coefficient for $j$th constraint column polynomial.
+/// * $\alpha_i$ is a composition coefficient for the $i$th trace polynomial.
+/// * $\beta_j$ is a composition coefficient for the $j$th constraint column polynomial.
 ///
-/// $T(x)$ and $H(x)$ are polynomials of degree $n - 1$, where $n$ is the length of the execution
-/// trace. Thus, the degree of $Y(x)$ polynomial is $n - 2$. To bring the degree back up to
-/// $n - 1$, we compute the DEEP composition polynomial as:
-/// $$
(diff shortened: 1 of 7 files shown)


prover/src/composer/mod.rs
```diff
@@ -88,7 +88,7 @@ impl<E: FieldElement> DeepCompositionPoly<E> {
                 &mut t1_composition,
                 poly,
                 ood_trace_states[0][i],
-                self.cc.trace[i].0,
```
irakliyk, line 91: I think we should also update comments on lines 57 - 58 above.

verifier/src/composer.rs
```diff
@@ -79,11 +79,11 @@ impl<E: FieldElement> DeepComposer<E> {
                 let value = E::from(value);
                 // compute the numerator of T'_i(x) as (T_i(x) - T_i(z)), multiply it by a
                 // composition coefficient, and add the result to the numerator aggregator
-                t1_num += (value - ood_main_trace_states[0][i]) * self.cc.trace[i].0;
+                t1_num += (value - ood_main_trace_states[0][i]) * self.cc.trace[i];
```
irakliyk, line 82: I think we should also update comments on lines 49 - 51 above.

air/src/air/coefficients.rs
```diff
@@ -77,39 +77,40 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
...
+/// https://eprint.iacr.org/2022/1216 and it relies on two points:
+///
+/// 1. The evaluation proofs for each trace polynomial at $z$ and $g \cdot z$ can be batched using
+/// the non-normalized Lagrange kernel over the set $\{z, g \cdot z\}$. This, however, requires
+/// that the FRI protocol is run with rate $\rho^{+} := \frac{\kappa + 2}{\nu}$ where $\kappa$ and
+/// $\nu$ are the length of the execution trace and the LDE domain size, respectively.
+/// 2. The resulting $Y(x)$ do not need to be degree adjusted but the soundness error of the
+/// protocol needs to be updated. For most combinations of batching parameters, this leads to a
+/// negligible increase in soundness error. The formula for the updated error can be found in
+/// Theorem 8 of https://eprint.iacr.org/2022/1216.
```
irakliyk, line 109: Should we update our security level computations [here](https://github.com/facebook/winterfell/blob/main/air/src/proof/mod.rs#L105) because of this?

prover/src/composer/mod.rs
Al-Kindi-0: Thanks

verifier/src/composer.rs
Al-Kindi-0: Thanks

air/src/proof/mod.rs
```diff
@@ -212,13 +213,22 @@ fn get_proven_security(
...
 ) -> u32 {
     let extension_field_bits = (base_field_bits * options.field_extension().degree()) as f64;
+
     let blowup_bits = options.blowup_factor().ilog2() as f64;
     let num_fri_queries = options.num_queries() as f64;
     let lde_size_bits = lde_domain_size.trailing_zeros() as f64;
 
+    // blowup_plus_bits is the number of bits in the blowup factor which is the inverse of
+    // `\rho^+ := trace_domain_size / lde_domain_size`. `\rho^+` is used in order to define a larger
+    // agreement parameter `\alpha^+ := (1 + 1/2m)\sqrt{rho^+} := 1 - \theta^+`. The reason for
```
irakliyk, line 227: - Should it be `\rho^+ := (trace_domain_size + 2) / lde_domain_size`
- Rather than `\rho^+ := trace_domain_size / lde_domain_size`?

At least that's how it seems to be computed in the code.

irakliyk (approved):
Looks good! I left a couple of minor nits inline.

Also, my understanding is that these changes do not affect conjectured security. Or should we make adjustments there too?

air/src/proof/mod.rs
Al-Kindi-0: Yes, thank you. That's the whole point of the change.

air/src/proof/mod.rs
```diff
@@ -212,13 +213,22 @@ fn get_proven_security(
...
+    // running FRI with a larger agreement parameter is to account for to account for the simplifed
+    // DEEP composition polynomial. See Protocol 3 in https://eprint.iacr.org/2022/1216.
```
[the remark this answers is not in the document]
Al-Kindi-0: Thanks!

Al-Kindi-0:
> Looks good! I left a couple of minor nits inline.
> 
> Also, my understanding is that these changes do not affect conjectured security. Or should we make adjustments there too?

Yes, we can make the same changes for the sake of consistency although it is questionable marrying the rigorous argument for changing $\rho$ to $\rho^+$ with something that is just a conjecture. In any case, the expression for the conjectured security has to be revisited as discussed in #151 .

One more thing, I had to use `log2` instead of `ilog2`. Is that ok?

One last thing, the proven security as currently implemented only takes into account the FRI part. This should be updated to include the terms coming from the IOP.

irakliyk (approved):
All looks good! Thank you!

Issue #180, opened 2023-03-26
Support custom implementations of Trace LDE


irakliyk:
Currently, Trace LDE is implemented as a concrete struct (see [here](https://github.com/facebook/winterfell/blob/main/prover/src/trace/trace_lde.rs)). It might be good to instead define as a trait and provide a default implementation (similar to what we did with `RandomCoin` recently). Making it a trait would make the structure more flexible and allow for custom implementations of Trace LDE which might be more suitable for various use cases (i.e., long-running provers where memory may not need to be allocated/de-allocated all the time, or provers which outsource LDE computations to hardware etc.).

The trait could look something like this:
```Rust
pub trait TraceLde {
    type BaseField: StarkField;
    type HashFn: ElementHasher<BaseField = Self::BaseField>;

    /// Takes main trace columns as input, interpolates them into polynomials in coefficient form,
    /// evaluates the polynomials over the LDE domain, and commits to the polynomial evaluations.
    ///
    /// Returns a tuple containing the column polynomials in coefficient from and the commitment
    /// to the polynomial evaluations over the LDE domain.
    fn set_main_trace(
        main_trace: &ColMatrix<Self::BaseField>,
    ) -> (ColMatrix<Self::BaseField>, <Self::HashFn as Hasher>::Digest);

    /// Takes auxiliary trace segment columns as input, interpolates them into polynomials in
    /// coefficient form, evaluates the polynomials over the LDE domain, and commits to the
    /// polynomial evaluations.
    ///
    /// Returns a tuple containing the column polynomials in coefficient from and the commitment
    /// to the polynomial evaluations over the LDE domain.
    fn add_aux_segment<E: FieldElement<BaseField = Self::BaseField>>(
        aux_trace: &ColMatrix<E>,
    ) -> (ColMatrix<E>, <Self::HashFn as Hasher>::Digest);

    /// Reads current and next rows from the main trace segment into the specified frame.
    fn read_main_trace_frame_into(
        &self,
        lde_step: usize,
        frame: &mut EvaluationFrame<Self::BaseField>,
    );

    /// Reads current and next rows from the auxiliary trace segment into the specified frame.
    fn read_aux_trace_frame_into<E: FieldElement<BaseField = Self::BaseField>>(
        &self,
        lde_step: usize,
        frame: &mut EvaluationFrame<E>,
    );

    /// Returns trace table rows at the specified positions along with Merkle authentication paths
    /// from the commitment root to these rows.
[...]

PR #207, opened 2023-07-07
Support custom implementations of Trace LDE

grjte:
This PR addresses #180. It's not quite ready for merge, but I'd like to get feedback on the approach

- [x] switch the `TraceLde` struct to a trait
- [x] add a default implementation of `TraceLde`
- [x] update all examples
- [x] update docs
- [x] replace unit tests for DefaultTraceLde


prover/src/lib.rs
```diff
@@ -152,6 +159,13 @@ pub trait Prover {
     /// proof size, and proof security level.
     fn options(&self) -> &ProofOptions;
 
+    /// TODO: docs
+    fn new_trace_lde<E: FieldElement<BaseField = Self::BaseField>>(
+        &self,
+        trace_info: &TraceInfo,
+        blowup: usize,
+    ) -> Self::TraceLde<E>;
```
irakliyk, line 167: Looking at this now, I do wonder if we should instead replace the `TraceLde::set_main_trace()` with `TraceLde::new()` function which would look something like this:
```Rust
fn new(
    trace_info: &TraceInfo,
    main_trace: &ColMatrix<Self::BaseField>,
    domain: &StarkDomain<Self::BaseField>,
) -> Self
```
And then call this constructor instead of using `new_trace_lde()`.

This would get rid of all the annoying checks for whether the main trace has already been set. And if we wanted to pass in some custom info into the `new()` function, we could still do it via the `meta` property of `TraceInfo`.

If we do this, we would also need to add `get_main_trace_commitment()` method to `TraceLde` - but I think that shouldn't cause any issues.

irakliyk (changes requested):
Looks great! Thank you! This is basically how I thought about this as well. I left a few comments inline. Most of them are pretty minor, except for the last comment which is about potentially adding a `new()` constructor to `TraceLde` and using it instead of `Prover::new_trace_lde()`.

Regarding the tests: I think it would be great to have unit tests on `DefaultTraceLde` - so, I'd prefer if the removed tests were moved there. Looking at them, they may actually be a bit simpler now as it doesn't seem like we need to instantiate `Air` to test these behavior on `DefaultTraceLde`.

prover/src/trace/trace_lde/mod.rs
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+use crate::{ColMatrix, StarkDomain};
+use air::{proof::Queries, EvaluationFrame};
+use crypto::{ElementHasher, Hasher};
+use math::{FieldElement, StarkField};
+use utils::collections::Vec;
```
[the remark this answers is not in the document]
grjte: I've updated this, but I have a question about it - it seemed to me that in Winterfell the pattern seemed to be generally favoring imports from the specific crate, whereas in Miden we seem to import from `super` and import things to `super` when we need them downstream. I've done a hybrid here (and in response to your other comment), where I've imported from `super`, but not added any module-level imports for things that are only used in this file (i.e. those are still imported from the crate). Can you clarify what the ideal pattern is in each place?

prover/src/lib.rs
grjte: I've updated this, but I have a question - we now need to store the trace polynomials so that they can be retrieved during `get_main_trace_commitment`, and it gets cloned in that method. I wonder if it would be better to instead do something else, like returning the `TracePolyTable` from `TraceLde::new` as well

prover/src/trace/trace_lde/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+///
+/// Segments are stored in two groups:
+/// - Main segment: this is the first trace segment generated by the prover. Values in this segment
+///   will always be elements in the base field (even when an extension field is used).
+/// - Auxiliary segments: a list of 0 or more segments for traces generated after the prover
+///   commits to the first trace segment. Currently, at most 1 auxiliary segment is possible.
+pub trait TraceLde: Sync {
+    type BaseField: StarkField;
+    type HashFn: ElementHasher<BaseField = Self::BaseField>;
+    type AuxElement: FieldElement<BaseField = Self::BaseField>;
```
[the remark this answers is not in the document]
grjte: > Also, one question: do we need both BaseField and AuxElement? I'm fine with having both, but curious if having just one of them would work (without complicating things too much).

I tried to do it without this previously and I couldn't figure out a way around it after spending quite a bit of time. Now that we're switching to having `TraceLde::new` in the trait, I can revisit this. I'd like to get your eyes on the changes first though to make sure you do want to take the new approach

prover/src/lib.rs
irakliyk: Yes - I like this approach better. So, basically, `TraceLde::new()` would look like this, right?
```Rust
fn new(
    trace_info: &TraceInfo,
    main_trace: &ColMatrix<Self::BaseField>,
    domain: &StarkDomain<Self::BaseField>,
) -> (Self, TracePolyTable<Self::BaseField>)
```

I guess another approach is to move the management of `TracePolyTable` completely into `TraceLde` - but I think this is a bigger refactoring and if we do want to do it, we could do it in a different PR.

grjte: > I guess another approach is to move the management of TracePolyTable completely into TraceLde - but I think this is a bigger refactoring and if we do want to do it, we could do it in a different PR.

I considered this, but it introduces other complications, since we would still have to do cloning to initialize the DEEP composition poly. I think the simpler approach (returning `TracePolyTable`) will be cleaner

prover/src/trace/trace_lde/mod.rs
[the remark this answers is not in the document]
irakliyk: I think just having `ExtensionField` and replacing `BaseField` references with `ExtensionField::BaseField` should work - but if it causes issues, I think leaving as is is fine.

prover/src/trace/trace_lde/mod.rs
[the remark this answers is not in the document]
irakliyk: I am not sure if I've followed any well-defined pattern here. More like, if the parent module already has some elements, import them from there, otherwise get them from the crate - but I'm not even sure if that's what I've always done. I think the way you have it now is probably how I would have done it as well.

prover/src/lib.rs
```diff
@@ -136,6 +136,13 @@ pub trait Prover {
     /// PRNG to be used for generating random field elements.
     type RandomCoin: RandomCoin<BaseField = Self::BaseField, Hasher = Self::HashFn>;
 
+    // Trace low-degree extension for accessing the LDEs of trace segments and their commitments.
+    type TraceLde<E: FieldElement<BaseField = Self::BaseField>>: TraceLde<
+        BaseField = Self::BaseField,
+        ExtensionField = E,
+        HashFn = Self::HashFn,
+    >;
```
irakliyk, line 144: I'd probably say "for building" rather than "for accessing".

irakliyk (approved):
Looks good! Thank you! I left a couple small comments inline. We can probably merge and address these in a separate commit.

irakliyk:
Oh - one other thing: we should add an entry to the changelog file for this.

irakliyk (approved):

PR #209, opened 2023-08-08
feature: add conditional support for serde

hackaugusto:
This adds conditional support for serde. It allow serializing/deserializing the field elements, which is useful for downstream applications, for instance, to save test data.

The change is behind a feature flag to avoid impacting compilation times.

### DIFF
--- a/math/Cargo.toml
+++ b/math/Cargo.toml
@@ -33,6 +33,7 @@ default = ["std"]
 std = ["utils/std"]
 
 [dependencies]
+serde = { version = "1.0", features = [ "derive" ], optional = true, default-features = false }
 utils = { version = "0.6", path = "../utils/core", package = "winter-utils", default-features = false }
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


hackaugusto:
Simple test for the output format:

```rust
use serde::{Deserialize, Serialize};
use serde_json::json;
use winter_math::fields;

#[derive(Serialize, Deserialize)]
struct Data {
    d0: fields::f64::BaseElement,
    d1: fields::f62::BaseElement,
    d2: fields::f128::BaseElement,
}

fn main() {
    let v = Data {
        d0: 1234123u64.into(),
        d1: 1234123u64.into(),
        d2: 1234123u64.into(),
    };

    println!(
        "{}",
        serde_json::to_string(&v).expect("json serialization failed")
    );
}
```

The code using `into` and `from` annotations outputs:

```
{"d0":1234123,"d1":1234123,"d2":1234123}
```

In contrast, without them:

```
{"d0":5300517923007285,"d1":1483409960899751059,"d2":1234123}
```

irakliyk (approved):
Looks good! Thank you!

Issue #206, opened 2023-06-15
RandomCoin trait simplification


Al-Kindi-0:
The RandomCoin trait seems to be too flexible for its intended use. More specifically, the `reseed_with_int()` method is used only in the context of reseeding with the PoW produced through grinding. Moreover, the `draw_integers(...)` method is always run right after `reseed_with_int()` in order to generate query indices for the FRI protocol. Thus it seems reasonable to move the logic of `reseed_with_int()` inside of `draw_integers(...)`. More specifically, the signature of `draw_integers` should be:

```rust
fn draw_integers(
    &mut self,
    num_values: usize,
    domain_size: usize,
    pow_nonce: u64,
) -> Result<Vec<usize>, RandomCoinError> 
```

and the first thing `draw_integers` does is reseed with `pow_nonce` before generating the integers.

On the verifier side, the code could look like 
```rust
let pow_nonce = channel.read_pow_nonce();

// make sure the proof-of-work specified by the grinding factor is satisfied
if public_coin.check_leading_zeros(pow_nonce) < air.options().grinding_factor() {
    return Err(VerifierError::QuerySeedProofOfWorkVerificationFailed);
}

// draw pseudo-random query positions for the LDE domain from the public coin; in the
// interactive version of the protocol, the verifier sends these query positions to the prover,
// and the prover responds with decommitments against these positions for trace and constraint
// composition polynomial evaluations.
let query_positions = public_coin
    .draw_integers(air.options().num_queries(), air.lde_domain_size(), pow_nonce)
    .map_err(|_| VerifierError::RandomCoinError)?;
```

irakliyk:
Closed by #214.

PR #214, opened 2023-09-18
Refactor RandomCoin trait

Al-Kindi-0:
Addresses #206

### DIFF
--- a/crypto/src/random/default.rs
+++ b/crypto/src/random/default.rs
@@ -117,63 +117,8 @@ impl<B: StarkField, H: ElementHasher<BaseField = B>> RandomCoin for DefaultRando
         self.counter = 0;
     }
 
-    /// Reseeds the coin with the specified value by setting the new seed to hash(`seed` ||
-    /// `value`).
-    ///
-    /// # Examples
-    /// ```
-    /// # use winter_crypto::{RandomCoin, DefaultRandomCoin, Hasher, hashers::Blake3_256};
-    /// # use math::fields::f128::BaseElement;
-    /// // initial elements for seeding the random coin
-    /// let seed = &[BaseElement::new(1), BaseElement::new(2), BaseElement::new(3), BaseElement::new(4)];
-    ///
-    /// let mut coin1 = DefaultRandomCoin::<Blake3_256<BaseElement>>::new(seed);
-    /// let mut coin2 = DefaultRandomCoin::<Blake3_256<BaseElement>>::new(seed);
-    ///
-    /// // should draw the same element form both coins
-    /// let e1 = coin1.draw::<BaseElement>().unwrap();;
-    /// let e2 = coin2.draw::<BaseElement>().unwrap();;
-    /// assert_eq!(e1, e2);
-    ///
-    /// // after reseeding should draw different elements
-    /// coin2.reseed_with_int(42);
-    /// let e1 = coin1.draw::<BaseElement>().unwrap();;
-    /// let e2 = coin2.draw::<BaseElement>().unwrap();;
-    /// assert_ne!(e1, e2);
-    /// ```
-    fn reseed_with_int(&mut self, value: u64) {
-        self.seed = H::merge_with_int(self.seed, value);
-        self.counter = 0;
-    }
-
     // PUBLIC ACCESSORS
     // --------------------------------------------------------------------------------------------
-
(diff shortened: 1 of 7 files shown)


prover/src/channel.rs
```diff
@@ -137,18 +137,18 @@ where
...
     /// The positions are drawn from the public coin uniformly at random.
-    pub fn get_query_positions(&mut self) -> Vec<usize> {
+    pub fn get_query_positions(&mut self, nonce: u64) -> Vec<usize> {
         let num_queries = self.context.options().num_queries();
         let lde_domain_size = self.context.lde_domain_size();
         self.public_coin
-            .draw_integers(num_queries, lde_domain_size)
+            .draw_integers(num_queries, lde_domain_size, nonce)
             .expect("failed to draw query position")
     }
```
irakliyk, line 146: I wonder if we actually need to pass in the `nonce` as a parameter here. It seems like we should be able to get this from `self` as `self.pow_nonce`. Or does this complicate something?

prover/src/channel.rs
```diff
@@ -137,18 +137,18 @@ where
...
 
     /// Determines a nonce, which when hashed with the current seed of the public coin results
     /// in a new seed with the number of leading zeros equal to the grinding_factor specified
     /// in the proof options.
-    pub fn grind_query_seed(&mut self) {
+    pub fn grind_query_seed(&mut self) -> u64 {
```
irakliyk, line 151: If the above comment applies, then we can probably leave this method as it was before.

prover/src/lib.rs
```diff
@@ -414,10 +414,10 @@ pub trait Prover {
         let now = Instant::now();
 
         // apply proof-of-work to the query seed
-        channel.grind_query_seed();
+        let nonce = channel.grind_query_seed();
 
         // generate pseudo-random query positions
-        let query_positions = channel.get_query_positions();
+        let query_positions = channel.get_query_positions(nonce);
```
irakliyk, line 420: Similar comment to above: we might be able to avoid making changes here.

crypto/src/random/mod.rs
```diff
@@ -55,7 +47,8 @@ pub trait RandomCoin: Sync {
     /// PRNG.
     fn draw<E: FieldElement<BaseField = Self::BaseField>>(&mut self) -> Result<E, RandomCoinError>;
 
-    /// Returns a vector of unique integers selected from the range [0, domain_size).
+    /// Returns a vector of unique integers selected from the range [0, domain_size) after it reseeds
+    /// the coin with a nonce by setting the new seed to hash(`seed` || `nonce`).
```
irakliyk, line 51: nit: I wonder if we should not mention how exactly the reseeding happens as it concavely be done differently for different implementations of this trait.

crypto/src/random/default.rs
```diff
@@ -208,7 +153,8 @@ impl<B: StarkField, H: ElementHasher<BaseField = B>> RandomCoin for DefaultRando
         Err(RandomCoinError::FailedToDrawFieldElement(1000))
     }
 
-    /// Returns a vector of unique integers selected from the range [0, domain_size).
+    /// Returns a vector of unique integers selected from the range [0, domain_size) after reseeding
+    /// the PRNG with the specified `nonce`.
```
irakliyk, line 157: On the other hand, here I'd probably mention the exact methodology as this is a concrete implementation.

irakliyk (changes requested):
Look good! Thank you! I left a few comments inline. Most of them are pretty small. The only more significant one is about using `nonce` from an instance variable rather than passing it via parameters. But if this significantly complicates FRI verifier interface, we can keep it as is.

prover/src/channel.rs
Al-Kindi-0: Makes total sense, we can indeed use `self.pow_nonce` to simplify things. I have done that now.

crypto/src/random/default.rs
Al-Kindi-0: Makes sense.

irakliyk (approved):
All looks good! Thank you!

irakliyk:
> The clippy error seems similar to the one [we hit on plonky2](https://github.com/mir-protocol/plonky2/pull/1236). Setting manually `CARGO_INCREMENTAL: 0` for the clippy run may do the trick.

Thank you! This seems to have fixed it.

Issue #204, opened 2023-05-26
Update proven security to include IOP terms


Al-Kindi-0:
The way we compute the proven security in `get_proven_security` takes into account only the FRI soundness error part $\epsilon_{FRI}$. It would be great to include the terms coming from the protocol pre-FRI. 
I believe the update expression should be like:

$$
L^{+}\cdot\left(\frac{1}{|\mathbb{F}|} + \frac{d\cdot(k^+ - 1) + (k-1)}{|\mathbb{F}| - |D\cup H|}\right) + \epsilon_{FRI}
$$

where:

1. $H$ is the trace domain of size $k$
2. $D$ is the LDE domain of size $n$
3. $k^+ := k + 2$
4. $L^+ := \frac{m+0.5}{\sqrt{\rho^+}}$ where $m\geq 3$ is the Johnson proximity parameter which for simiplicity can be taken to be equal to $3$ and $\rho^+ := \frac{k^+}{n}$
5. $\mathbb{F}$ is the extension field.
6. $\epsilon_{FRI}$ is the soundness error bound for FRI run with proximity parameter $\theta^+ := 1 - \alpha^+ := 1 - (1 + \frac{1}{2m})\cdot \sqrt{\rho^+}$
7. $d$ is the number of constraint composition column polynomials.

pdg744:
<img width="351" alt="Screen Shot 2023-07-28 at 12 08 46 PM" src="https://github.com/facebook/winterfell/assets/15272444/ff5153d2-39ed-4278-a35b-1cb7aa293b6b">

Not sure I'm reading this right; it looks like you're sayihng $d$ is the number of constraints? Or maybe `constraint composition column polynomials` is the number of segment polynomials? 

Ulrich defines $d$ as the maximum degree of the constraints, but I think the expression in the error bound is actually the number of segment polynomials.

irakliyk:
Closed by #215.

PR #215, opened 2023-09-25
Proven security for protocol

Al-Kindi-0:
Addresses #204 
This PR adds the ability to compute the proven security of the protocol given a set of parameters. It uses [Habock](https://eprint.iacr.org/2022/1216.pdf) and some additional observations and correction made by @benediamond (Ulvetanna).
One annoying thing is related to the repeated use of the min function. This leads to a potential loss of up to 2 bits of soundness. It is possible to avoid this by adding some extra logic to compare the exponents but I am still looking for something easier.


air/src/proof/mod.rs
```diff
@@ -207,61 +207,126 @@ fn get_conjectured_security(
...
+    min - 1
+}
 
-    cmp::min(
-        cmp::min(pre_query_security, query_security) - 1,
-        collision_resistance,
-    )
+// Computes the largest proximity parameter m needed for Theorem 8
+// in https://eprint.iacr.org/2022/1216.pdf to work.
+fn upper_m(h: u64) -> f64 {
```
irakliyk, line 329: A couple of small nits:
- Let's use `///` for doc comments.
- For proper link formatting we should put links in `<>`. Here it would be `/// in <https://eprint.iacr.org/2022/1216.pdf> to work.`
- I would probably name the function `compute_upper_m()`.
- Let's add `#[cfg(feature = "std")]` attribute since we currently call this only in the `std` context.

Also, maybe add to the comment that for practical trace lengths (e.g., up to $2^{30}$) the returned value will be smaller than about 2300.

air/src/proof/mod.rs
```diff
@@ -207,61 +207,126 @@ fn get_conjectured_security(
...
+    if fri_err_bits < 1 {
+        return 0;
+    }
+    let fri_err_bits = fri_err_bits - 1;
+
+    // To apply Theorem 8 in https://eprint.iacr.org/2022/1216.pdf, we need to apply FRI with
+    // a slightly larger agreement parameter alpha.
+    // More concretely, we need alpha > rho_plus.sqrt() where rho_plus is the rate in function field
+    // F(Z) and defined as (trace_domain_size + 2.0) / lde_domain_size .
+    // This means that the range of m needs to be restricted in order to insure that
```
irakliyk, line 290: Should "insure" be "ensure"?

air/src/proof/mod.rs
```diff
@@ -207,61 +207,126 @@ fn get_conjectured_security(
...
+    ) as u32
+}
 
-    // compute security we get by executing multiple query rounds
-    let security_per_query = 0.5 * blowup_plus_bits - (1.0 + 1.0 / (2.0 * m)).log2();
-    let mut query_security = (security_per_query * num_fri_queries) as u32;
+/// Computes proven security level for the specified proof parameters for a fixed
+/// value of the proximity parameter m.
+#[cfg(feature = "std")]
+fn proven_security_protocol_m(
```
irakliyk, line 251: Should we mention that this is in the list-decoding regime?

Also small nit: maybe we should rename this to `get_proven_security_for_m()`?

air/src/proof/mod.rs
```diff
@@ -207,61 +207,126 @@ fn get_conjectured_security(
...
+                options,
+                base_field_bits,
+                lde_domain_size,
+                trace_domain_size,
+                a as usize,
+            )
+        })
+        .expect(
+            "Should not fail since m_max is larger than m_min
+                             for all trace sizes of length greater than 4",
```
irakliyk, line 233: minor nit: I'd probably put the whole string on a single line (otherwise, error formatting could look weird).

irakliyk (approved):
Looks great! Thank you! I left a few small comments inline. We can merge once they are addressed.

Also, I'm curious, for some common sets of parameters, how do results given by this new approximation differ from the previous method?

air/src/proof/mod.rs
Al-Kindi-0: Done!

air/src/proof/mod.rs
Al-Kindi-0: Makes sense!

air/src/proof/mod.rs
Al-Kindi-0: Fixed, thanks!

air/src/proof/mod.rs
Al-Kindi-0: Sounds good!
I am not sure about the upper bound you mentioned, I think that for traces of length $2^{22}$ an upper bound is around 2 million. It might make sense to put a cap on the return value as usually the optimal $m$ is in the lower range of `[3,..., upper_m(h)]`.

air/src/proof/mod.rs
```diff
@@ -32,6 +32,7 @@ pub use table::Table;
 // ================================================================================================
 
 const GRINDING_CONTRIBUTION_FLOOR: u32 = 80;
+const PROXIMITY_PARAMETER_UPPER: u64 = 1000;
```
irakliyk, line 35: A couple of minor nits:

- I'd probably name this `MAX_PROXIMITY_PARAMETER`.
- Let's add `#[cfg(feature = "std")]` attribute for this constant.

air/src/proof/mod.rs
```diff
@@ -204,58 +205,129 @@ fn get_conjectured_security(
...
     collision_resistance: u32,
 ) -> u32 {
-    let extension_field_bits = (base_field_bits * options.field_extension().degree()) as f64;
+    let m_min: usize = 3;
+    let m_max = compute_upper_m(trace_domain_size);
+
+    // We cap the range to 1000 as the optimal m value will be in the lower range of [m_min, m_max]
+    // since increasing m too much will lead to a deterioration in the FRI commit soundness making
+    // any benefit gained in the FRI query soundess mute.
+    let m_max = cmp::min(m_max as u64, PROXIMITY_PARAMETER_UPPER);
```
irakliyk, line 223: nit: I would probably move this into `compute_upper_m` function.

irakliyk (approved):
Looks good! Thank you! I left a couple of non-blocking comments inline.

irakliyk:
> Maybe a little bit out of the context, but there is some "degradation" in computing the security level for composite/recursive protocols. If $\epsilon_1$ is the error for the "outer" proving system and $\epsilon_2$ is the error for the "inner" proving system, one can easily prove that $\epsilon_1 + \epsilon_2$ (sum of negligible terms is negligible) is the error for the composite/recursive system. 
Assuming a sufficiently large number of recursive steps, one can achieve the degradation of the original security level by some bits. E.g., say $\epsilon_1 = \epsilon_2 = \frac{1}{2^{128}}$, then a one-depth recursive system has $\epsilon_1 + \epsilon_2 = \frac{1}{2^{127}}$ soundness error (127-bit security level) even tho the original systems had 128-bit security level.
>
> Since it is very easy, probably it's a good idea to have the formula as well for the composite/recursive system.

Interesting! Would we need to estimate the depth of recursion for this?

In general, I'm thinking that maybe we should create a separate issue to address any remaining improvements we can make to security estimation.

hecmas:
> Interesting! Would we need to estimate the depth of recursion for this?

If you want to be as accurate as possible in your security estimations, then I would say yes!  

All of the concerns regarding the security of the recursive/composite proof system are very good explained in [this](https://youtu.be/0LW-qeVe6QI?t=424&feature=shared) video from Dan Boneh.

air/src/proof/mod.rs
Al-Kindi-0: Done!

air/src/proof/mod.rs
Al-Kindi-0: Done!

Issue #183, opened 2023-03-29
Enable proven security computations for `no-std`


irakliyk:
In #182 proven security estimation was disabled in `no-std` setting because there is an issue with floating point computations. It would be great to figure out how to enable it back.

One option is to rely on core intrinsics - but these are available only on nightly. Maybe there is a different way to do this.

irakliyk:
Closed by #218.

PR #218, opened 2023-10-04
Protocol proven security in no-std

Al-Kindi-0:
Addresses #183 
This solution relies on the `fixed` crate for fixed point arithmetic. It also uses an approximation for the `log(1 + x)` function as well as other upper/lower bounds. To compensate for the pessimistic bounds in the no-std version, some `-1` subtractions after `cmp::min` were removed. Surprisingly this lead to results which are closer to the results using infinite precision integers, at least in the cases that were tested.
Still the logic seems a bit over complicated and this means that there are probably some corner cases that were overlooked.

### DIFF
--- a/air/Cargo.toml
+++ b/air/Cargo.toml
@@ -22,6 +22,7 @@ std = ["crypto/std", "fri/std", "math/std", "utils/std"]
 [dependencies]
 crypto = { version = "0.6", path = "../crypto", package = "winter-crypto", default-features = false }
 fri = { version = "0.6", path = "../fri", package = "winter-fri", default-features = false }
+libm = "0.2.8"
 math = { version = "0.6", path = "../math", package = "winter-math", default-features = false }
 utils = { version = "0.6", path = "../utils/core", package = "winter-utils", default-features = false }
 
--- a/air/src/proof/mod.rs
+++ b/air/src/proof/mod.rs
@@ -32,8 +32,6 @@ pub use table::Table;
 // ================================================================================================
 
 const GRINDING_CONTRIBUTION_FLOOR: u32 = 80;
-
-#[cfg(feature = "std")]
 const MAX_PROXIMITY_PARAMETER: u64 = 1000;
 
 // STARK PROOF
@@ -114,10 +112,6 @@ impl StarkProof {
                 H::COLLISION_RESISTANCE,
             )
         } else {
-            #[cfg(not(feature = "std"))]
-            panic!("proven security level is not available in no_std mode");
-
-            #[cfg(feature = "std")]
             get_proven_security(
                 self.context.options(),
                 self.context.num_modulus_bits(),
@@ -208,7 +202,6 @@ fn get_conjectured_security(
 }
 
 /// Estimates proven security level for the specified proof parameters.
-#[cfg(feature = "std")]
 fn get_proven_security(
     options: &ProofOptions,
     base_field_bits: u32,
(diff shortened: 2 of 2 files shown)


air/src/proof/mod.rs
```diff
@@ -207,61 +217,276 @@ fn get_conjectured_security(
...
-    // compute security we get by executing multiple query rounds
-    let security_per_query = 0.5 * blowup_plus_bits - (1.0 + 1.0 / (2.0 * m)).log2();
-    let mut query_security = (security_per_query * num_fri_queries) as u32;
+    min - 1
+}
+
+/// Computes proven security level for the specified proof parameters for a fixed
+/// value of the proximity parameter m.
+#[cfg(not(feature = "std"))]
+fn proven_security_protocol_m_nostd(
```
irakliyk, line 380: Question: instead of creating a separate version of this entire function, could we create separate version of just `log2()` function? For example:

```Rust
#[cfg(feature = "std")]
fn log2(value: f64) -> f64 {
    value.log2()
}

#[cfg(not(feature = "std"))]
fn log2(value: f64) -> f64 {
    // our custom implementation
}
```

air/Cargo.toml
```diff
@@ -24,6 +24,8 @@ crypto = { version = "0.6", path = "../crypto", package = "winter-crypto", defau
 fri = { version = "0.6", path = "../fri", package = "winter-fri", default-features = false }
 math = { version = "0.6", path = "../math", package = "winter-math", default-features = false }
 utils = { version = "0.6", path = "../utils/core", package = "winter-utils", default-features = false }
+fixed = "1.20.0"
+typenum = "1.17.0"
```
irakliyk, line 28: Instead of using these two libraries, I wonder if it would be easier to use [libm](https://crates.io/crates/libm) wrapper. It seems to be even more popular than `fixed` library (7M vs. 180K recent downloads) and seems to provide the [log2](https://docs.rs/libm/latest/libm/fn.log2.html) function we looking for.

I haven't tested it, but maybe `log2` implementation in `libm` is close enough to Rust's standard library that we don't need to do anything else?

irakliyk (commented):
Thank you! Looks good! This is not a full review yet, but I left a couple of comments which may affect the approach.

air/Cargo.toml
Al-Kindi-0: Actually, using `libm` makes total sense. It simplifies the implementation considerably and reduces the surface of errors. Thank you!

air/src/proof/mod.rs
Al-Kindi-0: Makes total sense! I went ahead and did the same for the rest of the operations we make use of. Let me know if the approach agrees with what you proposed.

air/Cargo.toml
```diff
@@ -24,6 +24,7 @@ crypto = { version = "0.6", path = "../crypto", package = "winter-crypto", defau
 fri = { version = "0.6", path = "../fri", package = "winter-fri", default-features = false }
 math = { version = "0.6", path = "../math", package = "winter-math", default-features = false }
 utils = { version = "0.6", path = "../utils/core", package = "winter-utils", default-features = false }
+libm = "0.2.8"
```
irakliyk, line 27: minor nit: could we put dependencies in the alphabetical order?

irakliyk (approved):
Looks good! Thank you! I left one minor nit inline.

One question: are there any difference for the results we get in `std` and `no_std` contexts?

air/Cargo.toml
Al-Kindi-0: Sure!

Al-Kindi-0:
For the examples I run, no difference at all.

Issue #160, opened 2023-03-02
Potential vulnerability if not checking proof security level


irakliyk:
This issue was brought up by Quan Thoi Minh Nguyen from Google.

The core of the issue is how we define the verifier interface [here](https://github.com/facebook/winterfell/blob/main/verifier/src/lib.rs#L84). Basically, the verifier accepts a `StarkProof` as an input and `StarkProof` struct contains `ProofOptions` struct. In turn, `ProofOptions` contains a bunch of protocol parameters which, in the end, define the security level of the proof. This setup could allow a potential attacker to send a fake proof at a very low security level (e.g., just a few bits), and the verifier would accept this proof.

So, it is up to the users of this library to make sure they check security level of the proof before passing it to the `verify()` function.  While this works, it is probably not ideal as people might forget that this needs to be done leading to unintended concequences.

I can think of 3 ways of changing the verifier interface to make this type of bugs less likely.

#### Option 1: return security level from the `verify()` function
We could change the signature of the `verify()` function to something like this:
```Rust
pub fn verify<AIR: Air, HashFn: ElementHasher<BaseField = AIR::BaseField>>(
    proof: StarkProof,
    pub_inputs: AIR::PublicInputs,
) -> Result<u32, VerifierError>
```
Where the returned `u32` would specify the security level of the verified proof.

#### Option 2: pass `ProofOptions` struct separately
We could change the signature of the `verify()` function to something like this:
```Rust
pub fn verify<AIR: Air, HashFn: ElementHasher<BaseField = AIR::BaseField>>(
    proof: StarkProof,
    options: ProofOptions,
    pub_inputs: AIR::PublicInputs,
) -> Result<(), VerifierError>
```
This would also require the we remove `ProofOptions` from `StarkProof`.

#### Option 3: specify minimum security level
We could change the signature of the `verify()` function to something like this:
```Rust
pub fn verify<AIR: Air, HashFn: ElementHasher<BaseField = AIR::BaseField>>(
    proof: StarkProof,
    pub_inputs: AIR::PublicInputs,
    min_security_level: u32,
) -> Result<(), VerifierError>
```

My thinking is that option 1 is not really explicit enough and the return value could be ignored by the caller just as easily. Option 2 makes the interface less ergonomic and difficult to use, in my opinion. So, my preference would be with option 3 - but would love to know if others think differently.

daira:
> So, it is up to the users of this library to make sure they check security level of the proof before passing it to the `verify()` function.

[...]

PR #219, opened 2023-10-13
Add acceptable options argument to verification

Al-Kindi-0:
Addresses #160

### DIFF
--- a/verifier/src/errors.rs
+++ b/verifier/src/errors.rs
@@ -41,6 +41,17 @@ pub enum VerifierError {
     /// constraint evaluation queries do not represent a polynomial of the degree expected by the
     /// verifier.
     FriVerificationFailed(fri::VerifierError),
+    /// This error occurs when the parameters, that were used to generate the proof, do not provide
+    /// a conjectured security level greater than or equal to the conjectured security level
+    /// expected by the verifier.
+    InsufficientConjecturedSecurity(u32, u32),
+    /// This error occurs when the parameters, that were used to generate the proof, do not provide
+    /// a proven security level greater than or equal to the proven security level expected by
+    /// the verifier.
+    InsufficientProvenSecurity(u32, u32),
+    /// This error occurs when the parameters, that were used to generate the proof, do not match
+    /// any of the set of parameters expected by the verifier.
+    UnacceptableProofOptions,
 }
 
 impl fmt::Display for VerifierError {
@@ -74,6 +85,13 @@ impl fmt::Display for VerifierError {
             Self::FriVerificationFailed(err) => {
                 write!(f, "verification of low-degree proof failed: {err}")
             }
+            Self::InsufficientConjecturedSecurity(minimal_security, proof_security)=> {
+                write!(f, "insufficient proof security level: expected at least {minimal_security} bits of conjectured security, but was {proof_security} bits")
+            }
+            Self::InsufficientProvenSecurity(minimal_security, proof_security)=> {
+                write!(f, "insufficient proof security level: expected at least {minimal_security} bits of proven security, but was {proof_security} bits")
+            }
+            Self::UnacceptableProofOptions => {write!(f, "invalid proof options: security parameters do not match the acceptable parameter set")}
         }
     }
 }
--- a/verifier/src/lib.rs
+++ b/verifier/src/lib.rs
@@ -46,11 +46,12 @@ use math::{
 };
 
 pub use utils::{
-    ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable, SliceReader,
+    collections::Vec, ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
+    SliceReader,
 };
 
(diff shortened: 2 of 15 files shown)


verifier/src/errors.rs
```diff
@@ -74,6 +77,13 @@ impl fmt::Display for VerifierError {
             Self::FriVerificationFailed(err) => {
                 write!(f, "verification of low-degree proof failed: {err}")
             }
+            Self::InsufficientConjecturedSecurity(minimal_security,proofs_security)=> {
+                write!(f, "verification failed due to parameters not providing enough conjectured security: expected at least {minimal_security} bits of security but the proof provides only {proofs_security} bits of conjectured security")
+            }
```
irakliyk, line 82: nit: I would probable be a bit more concise than this. For example: "insufficient proof security level: expected at least {minimal_security} bits of conjectured security, but was {proof_security} bits"

Also, I would add a space after the comma in "minimal_security,proofs_security" on line 80.

verifier/src/errors.rs
```diff
@@ -74,6 +77,13 @@ impl fmt::Display for VerifierError {
...
+            Self::InsufficientProvenSecurity(minimal_security,proofs_security)=> {
+                write!(f, "verification failed due to parameters not providing enough proven security: expected at least {minimal_security} bits of security but the proof provides only {proofs_security} bits of proven security")
```
irakliyk, line 84: Similar comment as above.

verifier/src/errors.rs
```diff
@@ -74,6 +77,13 @@ impl fmt::Display for VerifierError {
...
+            }
+            Self::UnacceptableProofOptions => {write!(f, "verification failed due to parameters not providing enough security: the proof was generated with a set of parameters not accepted by the verifier")}
```
irakliyk, line 86: I would rephrase it a bit since this could fail not only if security is too low but also if the security level is higher or the same, but just with a different parameter set. Maybe something like: "invalid proof options: security parameters do not match the acceptable parameter set".

verifier/src/lib.rs
```diff
@@ -287,3 +317,13 @@ where
...
 }
+
+// ACCEPTABLE OPTIONS
+// ================================================================================================
+// Specifies either the minimal, conjectured or proven, security level or a set of
+// `ProofOptions` that are acceptable by the verification procedure.
+pub enum AcceptableOptions {
+    MinConjecturedSecurity(u32),
+    MinProvenSecurity(u32),
+    OptionSet(Vec<ProofOptions>),
```
irakliyk, line 328: Since this is a part of the public API, we should add doc comments for each variant.

Also, to get rid of the errors in `no_std`, we should import `Vec` form `utils::collections`.

verifier/src/lib.rs
```diff
@@ -70,25 +70,55 @@ pub use errors::VerifierError;
...
+                        proof_security,
+                    ));
+                }
+            }
+            AcceptableOptions::OptionSet(options) => {
+                if !options.iter().any(|opt| opt == proof.options()) {
+                    return Err(VerifierError::UnacceptableProofOptions);
+                }
+            }
+        }
```
irakliyk, line 120: I would probably move this logic into a function implemented on `AcceptableOptions`. Maybe something like:

```Rust
impl AcceptableOptions {
    pub fn validate(&self, proof: &StarkProof) -> Result<(), VerifierError> {
        ...
    }
}
```

irakliyk (changes requested):
Looks good! Thank you! I left a few small comments inline.

verifier/src/lib.rs
```diff
@@ -70,25 +71,31 @@ pub use errors::VerifierError;
...
+    acceptable_options: &AcceptableOptions,
 ) -> Result<(), VerifierError> 
 where 
     AIR: Air, 
     HashFn: ElementHasher<BaseField = AIR::BaseField>,
     RandCoin: RandomCoin<BaseField = AIR::BaseField, Hasher = HashFn>,
 {
+    // check that `proof` was generated with an acceptable set of parameters from the point of view
+    // of the verifier
+    let _ = acceptable_options.validate::<HashFn>(&proof);
```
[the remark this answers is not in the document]
Al-Kindi-0: My bad! Fixed now!

verifier/src/lib.rs
```diff
@@ -287,3 +294,47 @@ where
...
+    /// Minimal acceptable conjectured security level
+    MinConjecturedSecurity(u32),
+    /// Minimal acceptable proven security level
+    MinProvenSecurity(u32),
+    /// Set of acceptable proof parameters
+    OptionSet(Vec<ProofOptions>),
+}
+
+impl AcceptableOptions {
+    pub fn validate<H: Hasher>(&self, proof: &StarkProof) -> Result<(), VerifierError> {
```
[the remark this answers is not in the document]
Al-Kindi-0: Done!

irakliyk (approved):
All looks good! Thank you!

Issue #216, opened 2023-09-26
Remove duplicate query check in FRI


Al-Kindi-0:
The analysis of FRI doesn't require the FRI query indices to not include duplicates i.e. it only requires that they are sampled randomly but doesn't require the sampling to be without replacement.
Thus it makes sense to simplify the FRI query generation by removing the duplicate check.

irakliyk:
Makes sense! This would mean removing [these lines](https://github.com/facebook/winterfell/blob/main/crypto/src/random/default.rs#L223-L225), right?

Al-Kindi-0:
Yes! I would say even starting with line 215 but we would need to change a few extra things in addition.

PR #220, opened 2023-10-17
Remove duplicate random queries check

Al-Kindi-0:
Addresses #216

### DIFF
--- a/crypto/src/merkle/mod.rs
+++ b/crypto/src/merkle/mod.rs
@@ -206,7 +206,6 @@ impl<H: Hasher> MerkleTree<H> {
     /// * Number of provided indexes is greater than 255.
     /// * Any of the provided indexes are greater than or equal to the number of leaves in the
     ///   tree.
-    /// * List of indexes contains duplicates.
     pub fn prove_batch(&self, indexes: &[usize]) -> Result<BatchMerkleProof<H>, MerkleTreeError> {
         if indexes.is_empty() {
             return Err(MerkleTreeError::TooFewLeafIndexes);
@@ -214,13 +213,10 @@ impl<H: Hasher> MerkleTree<H> {
         if indexes.len() > proofs::MAX_PATHS {
             return Err(MerkleTreeError::TooManyLeafIndexes(proofs::MAX_PATHS, indexes.len()));
         }
-        let mut indexes = indexes.to_vec();
-        indexes.sort_unstable();
-        indexes.dedup();
 
-        let index_map = map_indexes(&indexes, self.depth())?;
-        let indexes = normalize_indexes(&indexes);
-        let mut leaves = vec![H::Digest::default(); index_map.len()];
+        let index_map = map_indexes(indexes, self.depth())?;
+        let mut leaves = vec![H::Digest::default(); indexes.len()];
+        let indexes = normalize_indexes(indexes);
         let mut nodes: Vec<Vec<H::Digest>> = Vec::with_capacity(indexes.len());
 
         // populate the proof with leaf node values
@@ -311,7 +307,6 @@ impl<H: Hasher> MerkleTree<H> {
     /// * Number of provided indexes is greater than 255.
     /// * Any of the specified `indexes` is greater than or equal to the number of leaves in the
     ///   tree from which the batch proof was generated.
-    /// * List of indexes contains duplicates.
     /// * Any of the paths in the batch proof does not resolve to the specified `root`.
     pub fn verify_batch(
         root: &H::Digest,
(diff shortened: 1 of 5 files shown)


irakliyk (changes requested):
Looks good! Thank you! Let's rebase this after #219 is merged.

Also, we should remove `DuplicateLeafIndex` error and associated code.

Al-Kindi-0:
I had to adapt the Merkle batch proving code in order to account for the possibility of duplicate indices. I spent some time thinking about the best way to do so without changing the current algorithm for batch proving and I couldn't think of anything better than just deduplicate at the outset.
Let me know if you see a better way to do so.

crypto/src/merkle/mod.rs
```diff
@@ -214,9 +214,12 @@ impl<H: Hasher> MerkleTree<H> {
         if indexes.len() > proofs::MAX_PATHS {
             return Err(MerkleTreeError::TooManyLeafIndexes(proofs::MAX_PATHS, indexes.len()));
         }
+        let mut indexes = indexes.to_vec();
+        indexes.sort_unstable();
+        indexes.dedup();
```
irakliyk, line 219: I'm not sure this part is needed as both `map_indexes()` and `normalize_indexes()` should naturally de-duplicate the data.

I think here we should add the following:
- Update comments to this function to describe the new behavior and remove the obsolete error condition.
- Similarly update comments for [verify_batch()](https://github.com/facebook/winterfell/blob/main/crypto/src/merkle/mod.rs#L313), [BatchMerkleProof::get_root()](https://github.com/facebook/winterfell/blob/main/crypto/src/merkle/proofs.rs#L123), and maybe other function.
- Add comments to `map_indexes()` and `normalize_indexes()`.

irakliyk, line 219: I'm not sure this part is needed as both `map_indexes()` and `normalize_indexes()` should naturally de-duplicate the data.

I think here we should add the following:
- Update comments to this function to describe the new behavior and remove the obsolete error condition.
- Similarly update comments for [verify_batch()](https://github.com/facebook/winterfell/blob/main/crypto/src/merkle/mod.rs#L313), [BatchMerkleProof::get_root()](https://github.com/facebook/winterfell/blob/main/crypto/src/merkle/proofs.rs#L123), and maybe other function.
- Add comments to `map_indexes()` and `normalize_indexes()`.

irakliyk (changes requested):
Looks good! Thank you! I couldn't think of a better way either - but I think this works fine.

I added a comment inline mostly about updating comments to reflect the new behavior. Also, let's add a line to the changelog for this.

crypto/src/merkle/mod.rs
Al-Kindi-0: Agreed, but modulo a small modification.
I have included the updates, let me know if that works.

crypto/src/merkle/mod.rs
```diff
@@ -216,8 +215,8 @@ impl<H: Hasher> MerkleTree<H> {
         }
 
         let index_map = map_indexes(indexes, self.depth())?;
+        let mut leaves = vec![H::Digest::default(); indexes.len()];
```
irakliyk, line 218: Wouldn't this create a bigger vector than we may need? Basically, if there are duplicate indexes, we'd allocate more a bigger vector than the number of leaves we'll eventually have. Or am I missing something?

Al-Kindi-0: That's correct and without that we would have an out-of-range index error here https://github.com/facebook/winterfell/blob/0b79eaef9b561995f750e123edec419fab83f5ea/crypto/src/merkle/mod.rs#L230

irakliyk: I looked into this a bit more, and while this works, I'm not sure this is the right solution. One of the issues is that contrary to what I said above `map_indexes()` function behaves differently for inputs with and without duplicates. But if we fix `map_indexes()` the verification procedure starts failing because of how we serialize/deserialize proofs. Specifically, when we serialize a proof, we always assume that the number of openings is the same as the number of queries specified by the protocol. So, we must serialize extra "dummy" values so that they can be deserialized properly. This is why this change works.

I think the proper solution is to update serialization code so that it serializes the actual number of openings. But this is a bigger change. I'll need to think about the best way to implement it.

Al-Kindi-0: I see, my question is then why not just deduplicate the indices at the start of `prove_batch`?

irakliyk: This seems to work only if we don't de-duplicate indexes in [BatchMerkleProof::get_root()](https://github.com/facebook/winterfell/blob/main/crypto/src/merkle/proofs.rs#L123) method. In theory, we should do this as well - but then the tests stop passing. So, I think this approach is fragile.

I've proposed an alternative approach in #224. Let me know what you think.

irakliyk:
Superseded by #224.

PR #223, opened 2023-10-23
Moved constraint composition into constraint evaluator

irakliyk:
This PR moves constraints composition step into the constraint evaluator. Thus, the constraint evaluator now always returns a single column with all constraint evaluations merged (previously, it would return constraint evaluations merged by unique divisors).


prover/src/lib.rs
```diff
@@ -326,46 +327,28 @@ pub trait Prover {
...
-        let now = Instant::now();
-        let composition_poly =
-            constraint_evaluations.into_poly(air.context().num_constraint_composition_columns())?;
-        #[cfg(feature = "std")]
-        debug!(
-            "Converted constraint evaluations into {} composition polynomial columns of degree {} in {} ms",
-            composition_poly.num_columns(),
-            composition_poly.column_degree(),
-            now.elapsed().as_millis()
+        // first, build a commitment to the evaluations of the composition polynomial columns
```
Al-Kindi-0, line 345: nit: maybe add constraint to "composition polynomial columns" in order to retain some consistency?

Al-Kindi-0 (commented):
Looks good, thank you!

prover/src/lib.rs
irakliyk: agreed! added!

PR #224, opened 2023-10-23
Allow duplicate query positions

irakliyk:
This PR is an alternative to #220 (and as such it also addresses #216). However, the approach taken here is de-duplicate query positions explicitly at the prover and the verifier, rather than doing it implicitly during Merkle path generation.

### DIFF
--- a/prover/src/channel.rs
+++ b/prover/src/channel.rs
@@ -136,13 +136,21 @@ where
     /// Returns a set of positions in the LDE domain against which the evaluations of trace and
     /// constraint composition polynomials should be queried.
     ///
-    /// The positions are drawn from the public coin uniformly at random.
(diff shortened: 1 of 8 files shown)


prover/src/channel.rs
```diff
@@ -136,13 +136,21 @@ where
...
         let num_queries = self.context.options().num_queries();
         let lde_domain_size = self.context.lde_domain_size();
-        self.public_coin
+        let mut positions = self
+            .public_coin
             .draw_integers(num_queries, lde_domain_size, self.pow_nonce)
-            .expect("failed to draw query position")
+            .expect("failed to draw query position");
+
+        // remove an duplicate positions from the list
```
Al-Kindi-0, line 149: typo: an -> any

Al-Kindi-0 (commented):
Looks good, thank you!

PR #225, opened 2023-10-30
implement `Ord` for `BaseElement`

plafer:
Implements `Ord` for all `BaseElement`s. One benefit is that they can be used as the key to a `BTreeMap` (or any other data structure that requires ordering).

### DIFF
--- a/math/src/field/f64/mod.rs
+++ b/math/src/field/f64/mod.rs
@@ -52,7 +52,7 @@ const ELEMENT_BYTES: usize = core::mem::size_of::<u64>();
 ///
 /// Internal values represent x * R mod M where R = 2^64 mod M and x in [0, M).
 /// The backing type is `u64` but the internal values are always in the range [0, M).
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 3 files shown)


math/src/field/f64/mod.rs
```diff
@@ -52,7 +52,7 @@ const ELEMENT_BYTES: usize = core::mem::size_of::<u64>();
 ///
 /// Internal values represent x * R mod M where R = 2^64 mod M and x in [0, M).
 /// The backing type is `u64` but the internal values are always in the range [0, M).
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialOrd, Ord)]
```
Nashtare, line 55: I think deriving `PartialOrd` directly relies on lexicographic ordering, which has little meaning when dealing with possibly unreduced elements or in Montgomery form?

plafer: Do you have any recommendations on how to implement this properly, if at all possible?

Nashtare: I would assume comparing reduced elements is the way to go (same as in implementations of `PartialEq`) though maybe @irakliyk has a different solution?

plafer:
Closing after speaking with @irakliyk, as this is no longer needed in miden-node (and @irakliyk preferred not to have fields be `Ord` since they don't have an inherent order).

PR #228, opened 2023-11-16
change `Debug` format for fields

plafer:
Implements `Debug` by hand for the base fields such that it outputs a human-readable integer.

The current behavior makes the fields unintuitive when debugging `Vec<BaseElement>`. Specifically, `Vec` doesn't implement `Display`, so we're forced to use the `Debug` output:

```rust
let v: Vec<BaseElement> = vec![BaseElement::from(1u64)];

println!("{v:?}");

// current output: [ BaseElement(<Montgomery form of 1>) ]
// new output: [ 1 ]
```

I believe most users will be confused by the current output, and would expect to read an integer (as I was).

### DIFF
--- a/math/src/field/f128/mod.rs
+++ b/math/src/field/f128/mod.rs
@@ -50,7 +50,7 @@ const ELEMENT_BYTES: usize = core::mem::size_of::<u128>();
 ///
 /// Internal values are stored in their canonical form in the range [0, M). The backing type is
 /// `u128`.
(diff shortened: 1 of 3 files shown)


irakliyk (approved):
Looks good! Thank you!

irakliyk:
`winter-math` v0.7.1, which includes this change, has also now been published to crates.io.

PR #231, opened 2023-11-30
Stark proof serializable deserializable

hackaugusto:
- adds size hints to some serializations
- implement serialization and deserialization traits for StarkProof (using existing code)
- add blanket implementation for integers and option type

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -272,10 +272,18 @@ impl FieldExtension {
     }
 }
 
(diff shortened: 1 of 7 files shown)


hackaugusto:
> The only thing I'm not sure about is serializing usize values. These could have different number of bytes on different platforms and so something serialized on one platform may fail to deserialize on the other.

good point, removed it.

PR #233, opened 2023-12-01
serde: add default deserializers implementations

hackaugusto:
The `Deserializer` trait does not have an associated type, all methods return `Self`. That means it is not possible to have a blanket implementation for slices `&[T]`, because the signature is required to be `fn read_from<R: ByteReader>(source: &mut R) -> Result<&[T], DeserializationError>;`. I can't add a associated type without making a breaking change because default associated types are unstable https://github.com/rust-lang/rfcs/blob/master/text/2532-associated-type-defaults.md. I also can't define a blanket implementation for `Vec<T>` because the default serializer does not include the length of the vector.

edit: also note:

- the serializer doesn't return errors, if one decides to encode an `usize` into `u8` there is nothing to do besides panicking, this could be an attack vector.
- the `write_batch_into` is the same as the vec implementation, should probably be removed since it doesn't have a clear receiver and there is equivalent functionality

### DIFF
--- a/utils/core/src/serde/mod.rs
+++ b/utils/core/src/serde/mod.rs
@@ -197,3 +197,44 @@ pub trait Deserializable: Sized {
         Ok(result)
     }
 }
+
+impl Deserializable for () {
+    fn read_from<R: ByteReader>(_source: &mut R) -> Result<Self, DeserializationError> {
+        Ok(())
+    }
+}
+
+impl Deserializable for u8 {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        source.read_u8()
+    }
+}
+
+impl Deserializable for u16 {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        source.read_u16()
+    }
+}
+
+impl Deserializable for u32 {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        source.read_u32()
+    }
+}
+
+impl Deserializable for u64 {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        source.read_u64()
+    }
+}
+
+impl<T: Deserializable> Deserializable for Option<T> {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let contains = source.read_bool()?;
+
+        match contains {
+            true => Ok(Some(T::read_from(source)?)),
+            false => Ok(None),
+        }
+    }
+}
--- a/utils/core/src/serde/mod.rs
+++ b/utils/core/src/serde/mod.rs
@@ -113,6 +113,19 @@ impl<T: Serializable> Serializable for &Vec<T> {
     }
 }
 
+impl Serializable for String {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        target.write_u64(self.len() as u64);
+        target.write_bytes(self.as_bytes());
+    }
+}
+
+impl Serializable for &String {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        (*self).write_into(target)
+    }
+}
+
 impl<T: Serializable, const N: usize> Serializable for Vec<[T; N]> {
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
         let source = flatten_slice_elements(self);
@@ -236,5 +249,17 @@ impl<T: Deserializable> Deserializable for Option<T> {
             true => Ok(Some(T::read_from(source)?)),
             false => Ok(None),
         }
+    }
+}
+
+impl Deserializable for String {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let length = source
+            .read_u64()?
+            .try_into()
+            .map_err(|err| DeserializationError::InvalidValue(format!("{err}",)))?;
+        let data = source.read_vec(length)?;
+
+        String::from_utf8(data).map_err(|err| DeserializationError::InvalidValue(format!("{err}")))
     }
 }
--- a/utils/core/src/serde/mod.rs
+++ b/utils/core/src/serde/mod.rs
@@ -53,6 +53,90 @@ impl Serializable for () {
     fn write_into<W: ByteWriter>(&self, _target: &mut W) {}
 }
 
+impl<T1> Serializable for (T1,)
+where
+    T1: Serializable,
+{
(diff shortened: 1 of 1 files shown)


utils/core/src/serde/mod.rs
```diff
@@ -113,17 +197,16 @@ impl<T: Serializable> Serializable for &Vec<T> {
     }
 }
 
-impl<T: Serializable, const N: usize> Serializable for Vec<[T; N]> {
```
hackaugusto, line 116: I added a blanket implementation for `impl<T, C> Serializable for [T; C]`. Together with the existing blanked implementation for `Vec<T>`, `&Vec<T>`, and `&[T]`, it made these 3 implementations obsolete.

utils/core/src/serde/mod.rs
```diff
@@ -53,6 +53,90 @@ impl Serializable for () {
     fn write_into<W: ByteWriter>(&self, _target: &mut W) {}
 }
 
+impl<T1> Serializable for (T1,)
```
irakliyk, line 56: minor nit: do we need to have a comma here after `T1`?

utils/core/src/serde/mod.rs
```diff
@@ -113,17 +197,16 @@ impl<T: Serializable> Serializable for &Vec<T> {
...
 }
 
-impl<T: Serializable, const N: usize> Serializable for &Vec<[T; N]> {
+impl Serializable for &String {
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
-        let source = flatten_slice_elements(self);
-        T::write_batch_into(source, target);
+        (*self).write_into(target)
     }
 }
```
irakliyk, line 211: This deviates somewhat from how we serialize lists in that we try not to make assumptions data types used to serialize the number of elements. I think we can leave it up to the callers to figure out how exactly they want to serialize a given string.

utils/core/src/serde/mod.rs
```diff
@@ -197,3 +279,160 @@ pub trait Deserializable: Sized {
...
+
+impl Deserializable for () {
+    fn read_from<R: ByteReader>(_source: &mut R) -> Result<Self, DeserializationError> {
+        Ok(())
+    }
+}
+
+impl<T1> Deserializable for (T1,)
```
irakliyk, line 289: minor nit: do we need the comma after `T1`?

utils/core/src/serde/mod.rs
```diff
@@ -197,3 +279,160 @@ pub trait Deserializable: Sized {
...
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let length = source
+            .read_u64()?
+            .try_into()
+            .map_err(|err| DeserializationError::InvalidValue(format!("{err}",)))?;
+        let data = source.read_vec(length)?;
+
+        String::from_utf8(data).map_err(|err| DeserializationError::InvalidValue(format!("{err}")))
+    }
+}
```
irakliyk, line 424: Similar comment to one of the previous comments re strings: I'd rather not make opinionated choices about which data type to use for serializing string lengths.

utils/core/src/serde/mod.rs
```diff
@@ -197,3 +279,160 @@ pub trait Deserializable: Sized {
...
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let data: Vec<T> = T::read_batch_from(source, C)?;
+
+        // SAFETY: the call above only returns a Vec if there are `C` elements, this conversion
+        // always succeeds
+        let res = data.try_into().unwrap_or_else(|v: Vec<T>| {
+            panic!("Expected a Vec of length {} but it was {}", C, v.len())
+        });
+
+        Ok(res)
```
irakliyk, line 436: It is a bit unfortunate that we have to allocate a vector here to deserialize the data - but I wonder if the compiler is smart enough to figure out that this is not needed.

irakliyk (changes requested):
Thank you! Looks good! I left a few small comments inline.

utils/core/src/serde/mod.rs
hackaugusto: yes:

> The syntax for tuple expressions is a parenthesized, comma separated list of expressions, called the tuple initializer operands. 1-ary tuple expressions require a comma after their tuple initializer operand to be disambiguated with a [parenthetical expression](https://doc.rust-lang.org/reference/expressions/grouped-expr.html).

https://doc.rust-lang.org/reference/expressions/tuple-expr.html

utils/core/src/serde/mod.rs
hackaugusto: https://github.com/facebook/winterfell/pull/233#discussion_r1417277215

utils/core/src/serde/mod.rs
hackaugusto: I don't understand the problem. You don't want these implementations? You don't want to include the string length?

hackaugusto: I guess the issue is with the use of `as_bytes`? You don't want to encode the data as `utf8`?

The biggest issue here is that the user can **not** implement this trait, because of the orphan rules. So here we either have to make the choice or not give an implementation to the user.

When the user is writing their own `Serializable`, they can also deviate from this by just not calling it.

irakliyk: I would prefer to remove these implementations as serializing strings without length is probably not a good idea, and serializing them with length is not consistent with the overall approach (e.g., different from how we serialize slices, vectors etc.).

I don't think this affects the usability too much as the user can serialize strings manually pretty easily - e.g.,:
```Rust
let x = "hello world".to_string();

target.write_u64(x.len() as u64);
target.write_bytes(x.as_bytes());
```

On the deserialization front, I think we can improve the ergonomics by adding `read_string()` method to `ByteReader` interface. It could look something like:
```Rust
fn read_string(&mut self, len: usize) -> Result<String, DeserializationError> {
    let data = source.read_vec(len)?;
    String::from_utf8(data).map_err(|err| DeserializationError::InvalidValue(format!("{err}")))
}
```

hackaugusto: > I would prefer to remove these implementations as serializing strings without length is probably not a good idea, and serializing them with length is not consistent with the overall approach (e.g., different from how we serialize slices, vectors etc.).

The question is, does it make sense to serialize slices and vectors without length too? And if this is done for consistency sake, why can't the String be serialized without the length too?

> I don't think this affects the usability too much as the user can serialize strings manually pretty easily - e.g.,:

Your call

> On the deserialization front, I think we can improve the ergonomics by adding read_string() method to ByteReader interface. It could look something like:

This can be done in another PR.

Removed the default implementation

irakliyk (approved):
Looks good! Thank you!

PR #236, opened 2023-12-18
bugfix: StarkProof is not always the last element in a serialization buffer

hackaugusto:
This prevents other data structure to be serialized in the same buffer, which happen naturally for structs with additional field elements.


irakliyk (approved):
Thank you for fixing this! Looks good.

PR #238, opened 2023-12-21
Implement serialization and deserialization for `usize` type

Overcastan:
This PR adds [vint64](https://docs.rs/vint64/latest/vint64/) variable-length serialization and deserialization implementation for `usize` type.

### DIFF
--- a/utils/core/Cargo.toml
+++ b/utils/core/Cargo.toml
@@ -22,3 +22,6 @@ std = []
 
 [dependencies]
 rayon = { version = "1.8", optional = true }
+
+[dev-dependencies]
+proptest = "1.3"
--- a/utils/core/src/serde/byte_reader.rs
+++ b/utils/core/src/serde/byte_reader.rs
@@ -92,6 +92,41 @@ pub trait ByteReader {
         Ok(u64::from_le_bytes(bytes))
     }
 
+    /// Returns a usize value read from `self` in [vint64](https://docs.rs/vint64/latest/vint64/)
+    /// format.
+    ///
+    /// # Errors
+    /// Returns a [DeserializationError] if:
+    /// * usize value could not be read from `self`.
+    /// * encoded value is greater than `usize` maximum value on a given platform.
+    fn read_usize(&mut self) -> Result<usize, DeserializationError> {
+        let first_byte = self.peek_u8()?;
+        let length = first_byte.trailing_zeros() as usize + 1;
+
+        let result = if length == 9 {
+            // 9-byte special case
+            self.read_u8()?;
+            let value = self.read_array::<8>()?;
+            u64::from_le_bytes(value)
+        } else {
+            let mut encoded = [0u8; 8];
+            let value = self.read_slice(length)?;
+            encoded[..length].copy_from_slice(value);
+            u64::from_le_bytes(encoded) >> length
+        };
+
+        // check if the result value is within acceptable bounds for `usize` on a given platform
+        if result > usize::MAX as u64 {
+            return Err(DeserializationError::InvalidValue(format!(
+                "Encoded value must be less than {}, but {} was provided",
+                usize::MAX,
+                result
+            )));
+        }
+
+        Ok(result as usize)
+    }
+
     /// Returns a u128 value read from `self` in little-endian byte order.
     ///
     /// # Errors
--- a/utils/core/src/serde/byte_writer.rs
+++ b/utils/core/src/serde/byte_writer.rs
@@ -62,6 +62,24 @@ pub trait ByteWriter: Sized {
         self.write_bytes(&value.to_le_bytes());
     }
 
+    /// Writes a usize value in [vint64](https://docs.rs/vint64/latest/vint64/) format into `self`.
+    ///
+    /// # Panics
+    /// Panics if the value could not be written into `self`.
+    fn write_usize(&mut self, value: usize) {
+        let length = encoded_len(value);
+
+        // 9-byte special case
+        if length == 9 {
+            // length byte is zero in this case
+            self.write_u8(0);
+            self.write(value.to_le_bytes());
+        } else {
+            let encoded_bytes = ((value << 1 | 1) << (length - 1)).to_le_bytes();
+            self.write_bytes(&encoded_bytes[..length]);
+        }
+    }
+
     /// Writes a serializable value into `self`.
     ///
     /// # Panics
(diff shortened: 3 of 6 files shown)


utils/core/src/serde/byte_writer.rs
```diff
@@ -62,6 +62,32 @@ pub trait ByteWriter: Sized {
...
+        } else {
+            (usize::BITS - value.leading_zeros() + OCTET_VALUE_SIZE - 1) / OCTET_VALUE_SIZE
+        };
+        for i in 0..chunks_num {
+            // get the value of i'th 7 bit chunk
+            let mut octet_value = (value >> (OCTET_VALUE_SIZE * i) & OCTET_BIT_MASK as usize) as u8;
+            // if this chunk is not last, cange continuation bit to 1
+            if i != chunks_num - 1 {
+                octet_value += OCTET_CONTINUATION_BIT;
+            }
```
irakliyk, line 84: nit: should "cange" but "change"?

utils/core/src/serde/byte_reader.rs
```diff
@@ -92,6 +92,26 @@ pub trait ByteReader {
...
+        loop {
+            let octet = self.read_u8()?;
+            // get the value from the octet and accumulate it in the result
+            result += (octet & OCTET_BIT_MASK) as usize * 128usize.pow(offset);
+            // if the continuation bit equals 0, finish reading
+            if octet >> OCTET_VALUE_SIZE == 0 {
+                break;
+            }
+            offset += 1;
+        }
```
irakliyk, line 111: We should probably check if the value we read is within acceptable bounds for `usize` on a given platform - and if not, return an error.

utils/core/src/tests.rs
```diff
@@ -112,6 +112,29 @@ fn write_serializable() {
     assert_eq!(234567u128, reader.read_u128().unwrap());
 }
 
+#[test]
+fn write_serializable_usize() {
```
irakliyk, line 116: Could we also add a proptest-based test for this?

utils/core/src/serde/byte_writer.rs
```diff
@@ -62,6 +62,32 @@ pub trait ByteWriter: Sized {
...
+    fn write_usize(&mut self, value: usize) {
+        let mut uleb128_bytes: Vec<u8> = Vec::new();
+
+        // get the number of 7 bit chunks sufficient for value encoding
+        let chunks_num = if value == 0 {
+            1
+        } else {
+            (usize::BITS - value.leading_zeros() + OCTET_VALUE_SIZE - 1) / OCTET_VALUE_SIZE
+        };
+        for i in 0..chunks_num {
```
irakliyk, line 78: minor nit: I'd add a blank line above this line.

irakliyk (changes requested):
Looks good! Thank you! I left a few comments inline - the main one is about more error checking on decoding.

Also, is this the same format as what protobuf uses for variable integer encoding?

utils/core/Cargo.toml
```diff
@@ -21,4 +21,5 @@ default = ["std"]
 std = []
 
 [dependencies]
+proptest = "1.3"
```
irakliyk, line 24: This should probably be a dev dependency.

utils/core/src/serde/byte_reader.rs
```diff
@@ -92,6 +92,36 @@ pub trait ByteReader {
...
+        loop {
+            let octet = self.read_u8()?;
+            // get the value from the octet and accumulate it in the result
+            result += (octet & OCTET_BIT_MASK) as u64 * 128u64.pow(offset);
+            // if the continuation bit equals 0, finish reading
+            if octet >> OCTET_VALUE_SIZE == 0 {
+                break;
+            }
+            offset += 1;
+        }
```
irakliyk, line 111: If someone encodes a malicious value (where continuation bit is always one for many many bits) - how would this work? Should we maybe always break after 9 iterations?

utils/core/src/serde/byte_writer.rs
```diff
@@ -62,6 +62,33 @@ pub trait ByteWriter: Sized {
         self.write_bytes(&value.to_le_bytes());
     }
 
+    /// Writes a usize value in ULEB128 format into `self`.
+    ///
+    /// # Panics
+    /// Panics if the value could not be written into `self`.
+    fn write_usize(&mut self, value: usize) {
+        let mut uleb128_bytes: Vec<u8> = Vec::new();
```
irakliyk, line 70: Can this be implemented without allocating a vector?

utils/core/src/serde/byte_reader.rs
```diff
@@ -92,6 +92,46 @@ pub trait ByteReader {
...
+                break;
+            }
+
+            // return an error if the number of chunks is greater than maximum amount required for
+            // usize encoding
+            if offset > 9 {
+                return Err(DeserializationError::InvalidValue(
+                    "Too many chunks for usize encoding".to_string(),
+                ));
+            }
```
irakliyk, line 118: Could we add a test which triggers this condition?

utils/core/src/serde/byte_writer.rs
```diff
@@ -83,3 +103,22 @@ impl ByteWriter for Vec<u8> {
...
+        8..=14 => 8,
+        15..=21 => 7,
+        22..=28 => 6,
+        29..=35 => 5,
+        36..=42 => 4,
+        43..=49 => 3,
+        50..=56 => 2,
+        57..=64 => 1,
+        _ => unreachable!(),
+    }
```
irakliyk, line 123: Would this be equivalent to something like this:
```Rust
let zeros = value.leading_zeros() as usize;
let len = zeros.saturating_sub(1) / 7;
9 - core::cmp::min(len, 8)
```

Overcastan: Indeed! I suppose, that this implementation works a little bit slower, but it looks much better for sure!

utils/core/src/serde/byte_writer.rs
```diff
@@ -83,3 +103,13 @@ impl ByteWriter for Vec<u8> {
         self.extend_from_slice(values);
     }
 }
+
+// HELPER FUNCTIONS
+// ================================================================================================
+
+/// Returns the length of the value in vint64 enoding.
```
irakliyk, line 110: typo: "enoding" should probably be "encoding".

utils/core/src/serde/byte_writer.rs
```diff
@@ -62,6 +62,26 @@ pub trait ByteWriter: Sized {
...
+        // 9-byte special case
+        if length == 9 {
+            // length byte is zero in this case
+            self.write_u8(0);
+            self.write(value.to_le_bytes());
+        } else {
+            let encoded_bytes = ((value << 1 | 1) << (length - 1)).to_le_bytes();
+            for byte in encoded_bytes.iter().take(length) {
+                self.write_u8(*byte)
+            }
```
irakliyk, line 81: Could we use `self::write_bytes(&encoded_bytes[..length]);` here instead?

irakliyk (approved):
Looks great! Thank you! I left a couple of small comments inline. After these are addressed, let's squash the commits and merge.

utils/core/src/serde/byte_writer.rs
irakliyk: I didn't benchmark, but looking at the generated assembly, I think this should actually be faster as compared to the `match` approach. Or, at least, there is much less branching in the logic.

PR #239, opened 2024-01-04
Remove `Serializable` and `Deserializable` implementations from slices and vecs

irakliyk:
This PR removes current implementations of `Serializable` and `Deserilizable` traits from slices and vectors. The motivation for this is that the way sterilization worked for these structs was somewhat confusing (slices and vectors were serialized without their lengths). Instead, we can use explicit calls to `write_batch()` and `read_batch()` methods to accomplish the same thing.

In the future, we can reimplement `Serializable` and `Deserializable` on slices and vecs, but this time we would serialize them together with the number of elements (for which we can use `usize` serialization implemented in #238).

### DIFF
--- a/utils/core/src/serde/byte_writer.rs
+++ b/utils/core/src/serde/byte_writer.rs
@@ -62,6 +62,14 @@ pub trait ByteWriter: Sized {
         self.write_bytes(&value.to_le_bytes());
     }
 
+    /// Writes a u128 value in little-endian byte order into `self`.
+    ///
+    /// # Panics
+    /// Panics if the value could not be written into `self`.
+    fn write_u128(&mut self, value: u128) {
+        self.write_bytes(&value.to_le_bytes());
+    }
+
     /// Writes a usize value in [vint64](https://docs.rs/vint64/latest/vint64/) format into `self`.
     ///
     /// # Panics
@@ -87,6 +95,15 @@ pub trait ByteWriter: Sized {
     fn write<S: Serializable>(&mut self, value: S) {
         value.write_into(self)
     }
+
+    /// Serializes all `elements` and writes the resulting bytes into `self`.
+    ///
+    /// This method does not write any metadata (e.g. number of serialized elements) into `self`.
+    fn write_many<S: Serializable>(&mut self, elements: &[S]) {
+        for element in elements {
+            element.write_into(self);
+        }
+    }
 }
 
 // BYTE WRITER IMPLEMENTATIONS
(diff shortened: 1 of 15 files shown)


utils/core/src/serde/byte_writer.rs
```diff
@@ -87,6 +95,15 @@ pub trait ByteWriter: Sized {
...
     }
+
+    /// Serializes all `elements` and writes the resulting bytes into `self`.
+    ///
+    /// This method does not write any metadata (e.g. number of serialized elements) into `self`.
+    fn write_batch<S: Serializable>(&mut self, elements: &[S]) {
+        for element in elements {
+            element.write_into(self);
+        }
+    }
```
plafer, line 106: With these methods, we have an unfortunate "circular dependency" on `Serializable` and `ByteWriter`. IMO, `ByteWriter` should only know how to serialize primitive values, and `Serializable` would serve to serialize arbitrary types. This translates to:

1. Remove `write()`: we can instead use `serializable.write_into(&mut byte_writer)` instead
2. Remove `write_batch()`

Since there are multiple ways to serialize a batch of serializables (no metadata such as this implementation, prefix the length as u64, etc), I would instead export this as a standalone function `utils::core::serde::serialize_batch_no_metadata()` (or whatever we want to call it). We can later add other batch serialization strategies, and users can choose whichever one they want.

This is just my current intuition though, I haven't played with the idea, so I could be missing something too. What do you think?

An analogous comment applies to `ByteReader`.

plafer (commented):
Left a suggestion; otherwise LGTM as is

utils/core/src/serde/byte_writer.rs
irakliyk: Yeah - I kind of went back and forth with this. The main reason for having `write()` and `write_batch()` on `ByteWriter` is that it presents a more convenient interface. For example, serializing a vector of arrays would look as follows:
```Rust
// using ByteWriter
writer.write_batch(&my_vector_of_arrays);

// using Serializable (assuming our arrays are [u64; 4]
<[u64; 4]>.write_batch_into(&mut writer, &my_vector_of_arrays);
```

But maybe this convenience is overshadowed by the benefit of defining just one way of doing things.

irakliyk: > * Remove `write()`: we can instead use `serializable.write_into(&mut byte_writer)` instead
> * Remove `write_batch()`

I've implemented this, but looking at the code, decided to roll it back as it made the code a bit less regular and a bit more difficult to read. I don't mind revisiting this in the future though.

PR #242, opened 2024-01-12
Make dependence on number of openings explicit in security estimator

Al-Kindi-0:
As the title says and it corrects a few minor mistakes. The security estimates do not change (when rounding to bits) after testing with some of the examples.

### DIFF
--- a/air/src/proof/mod.rs
+++ b/air/src/proof/mod.rs
@@ -288,28 +288,7 @@ fn proven_security_protocol_for_m(
     let m = m as f64;
     let rho = 1.0 / options.blowup_factor() as f64;
     let alpha = (1.0 + 0.5 / m) * sqrt(rho);
-    let theta = 1.0 - alpha;
-    let max_deg = options.blowup_factor() as f64;
-
-    let lde_domain_size = (trace_domain_size * options.blowup_factor()) as f64;
-    let trace_domain_size = trace_domain_size as f64;
-
-    // Computes FRI commit-phase (i.e., pre-query) soundness error.
-    // This considers only the first term given in eq. 7 in https://eprint.iacr.org/2022/1216.pdf,
-    // i.e. 0.5 * (m + 0.5)^7 * n^2 / (rho^1.5.q) as all other terms are negligible in comparison.
-    let fri_commit_err_bits = extension_field_bits
-        - log2((0.5 * powf(m + 0.5, 7.0) / powf(rho, 1.5)) * powf(lde_domain_size, 2.0));
-
-    // Compute FRI query-phase soundness error
-    let fri_queries_err_bits =
-        options.grinding_factor() as f64 - log2(powf(1.0 - theta, num_fri_queries));
-
-    // Combined error for FRI
-    let fri_err_bits = cmp::min(fri_commit_err_bits as u64, fri_queries_err_bits as u64);
-    if fri_err_bits < 1 {
-        return 0;
-    }
-    let fri_err_bits = fri_err_bits - 1;
+    let max_deg = options.blowup_factor() as f64 + 1.0;
 
     // To apply Theorem 8 in https://eprint.iacr.org/2022/1216.pdf, we need to apply FRI with
     // a slightly larger agreement parameter alpha.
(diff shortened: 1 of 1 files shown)


irakliyk (commented):
Looks good! Thank you for updating this!

I think it would be good to add a test for `get_proven_security()` function at a few sample parameter sets. For example, we could use:`base_field_bits = 64`, `collision_resistance = 128`, `trace_domain_size = 2^20`, and then vary `num_queries`, `blowup_factor`, and `extension_degree` to get 96 bits or 128 bits of security.

These could be good "sanity checks" to make sure the numbers we are getting are as expected.

air/src/proof/mod.rs
```diff
@@ -404,3 +406,259 @@ pub fn ceil(value: f64) -> f64 {
...
+    fn get_96_bits_security() {
+        let field_extension = FieldExtension::Cubic;
+        let base_field_bits = BaseElement::MODULUS_BITS;
+        let fri_folding_factor = 8;
+        let fri_remainder_max_degree = 127;
+        let grinding_factor = 20;
+        let mut blowup_factor = 4;
+        let mut num_queries = 80;
+        let collision_resistance = 128;
+        let trace_length = 1 << 18;
```
irakliyk, line 426: minor nit: should we use `2_usize.pow(18)` to make the operation a bit more explicit?

air/src/proof/mod.rs
```diff
@@ -404,3 +406,259 @@ pub fn ceil(value: f64) -> f64 {
...
+            fri_remainder_max_degree as usize,
+        );
+        let security_1 =
+            get_proven_security(&options, base_field_bits, trace_length, collision_resistance);
+
+        assert_eq!(security_1, 97);
+
+        // increasing the blowup factor should increase the bits of security gained per query
+        blowup_factor <<= 1;
+        num_queries -= 27;
```
irakliyk, line 443: nit: I would probably redefine the variables here. For example:
```Rust
let blowup_factor = 8;
let num_queries = 53;
```

air/src/proof/mod.rs
```diff
@@ -404,3 +406,259 @@ pub fn ceil(value: f64) -> f64 {
...
+    fn get_128_bits_security() {
+        let field_extension = FieldExtension::Cubic;
+        let base_field_bits = BaseElement::MODULUS_BITS;
+        let fri_folding_factor = 8;
+        let fri_remainder_max_degree = 127;
+        let grinding_factor = 20;
+        let mut blowup_factor = 8;
+        let mut num_queries = 85;
+        let collision_resistance = 128;
+        let trace_length = 1 << 18;
```
irakliyk, line 468: Similar comment as above here and for tests below as well.

air/src/proof/mod.rs
```diff
@@ -404,3 +406,259 @@ pub fn ceil(value: f64) -> f64 {
...
+            fri_remainder_max_degree as usize,
+        );
+        let security_1 =
+            get_proven_security(&options, base_field_bits, trace_length, collision_resistance);
+
+        assert_eq!(security_1, 128);
+
+        // increasing the blowup factor should increase the bits of security gained per query
+        blowup_factor <<= 1;
+        num_queries -= 20;
```
irakliyk, line 485: Similar comments as above here and for tests below.

air/src/proof/mod.rs
```diff
@@ -404,3 +406,259 @@ pub fn ceil(value: f64) -> f64 {
...
+            blowup_factor,
+            grinding_factor,
+            field_extension,
+            fri_folding_factor as usize,
+            fri_remainder_max_degree as usize,
+        );
+        let security_1 =
+            get_proven_security(&options, base_field_bits, trace_length, collision_resistance);
+
+        assert_eq!(security_1, 67);
```
irakliyk, line 523: This is the highest security level we can get from quadratic extension field (without using unique decoding regime), right?

irakliyk (approved):
The tests look great! Thank you! I left a few small nits inline - but this is basically ready to merge.

air/src/proof/mod.rs
Al-Kindi-0: Done

air/src/proof/mod.rs
Al-Kindi-0: Yes indeed, for the specified trace length and blowup factor

irakliyk (approved):
All looks good! Thank you!

Issue #230, opened 2023-11-28
`f64` field: `BaseElement` should not be convertible from `u64` or `u128` without error


plafer:
We currently convert `From<u64>` and `From<128>` for `BaseElement` ([here](https://github.com/facebook/winterfell/blob/cd92409298f33d02a76fda9b981a0b4b3b2fce8c/math/src/field/f64/mod.rs#L516-L531)). I believe these should be `TryFrom<u64>` and `TryFrom<u128>`.

If the value is greater than `2^64 - 2^32 + 1`, we silently perform a modulus. This can potentially be a vector of attack. For example, in the Miden VM, `BaseElement` is sometimes used to represent the length of a list. When initializing the VM, we take the list length as a `u64`, and then into `BaseElement`. Basically, if the list is very long (`>= 2^64 - 2^32 + 1`), there will be a silent wrap, and we will end up processing just a few elements in the VM. Now I don't know of any hardware that can store any list of that kind of size, but I can see this wrapping used in conjunction with other vulnerabilities to create a hack.

This is a direct consequence of the fact that if I store `2^64 - 2^32 + 100` in a `u64`, I see this as being unrepresentable in a `BaseElement`, as opposed to thinking of it as equivalent to `100`.

In case some users need the current behavior, I would create a function, say `from_u64_with_wrapping()`, and have the canonical Rust conversion (i.e. `From`/`TryFrom`) be "safer".

Would love to hear your thoughts on this. Do you also see it as enabling vulnerabilities of some kind?

irakliyk:
Closed by #243.

PR #243, opened 2024-01-28
Remove `From` impls resulting in silent field element conversions

irakliyk:
This PR addresses #230 and also does the same for `f62` and `f128` field.

Other changes in this PR:
- Cleaned up code as per clippy suggestions.
- Made some functions `const` for `f64` field.
- Added `as_int()` implementation to the `f64` field struct.

### DIFF
--- a/air/src/proof/context.rs
+++ b/air/src/proof/context.rs
@@ -27,10 +27,22 @@ impl Context {
     // --------------------------------------------------------------------------------------------
     /// Creates a new context for a computation described by the specified field, trace info, and
     /// proof options.
+    ///
+    /// # Panics
+    /// Panics if either trace length or the LDE domain size implied by the trace length and the
+    /// blowup factor is greater then [u32::MAX].
     pub fn new<B: StarkField>(trace_info: &TraceInfo, options: ProofOptions) -> Self {
+        // TODO: return errors instead of panicking?
+
+        let trace_length = trace_info.length();
+        assert!(trace_length <= u32::MAX as usize, "trace length too big");
+
+        let lde_domain_size = trace_length * options.blowup_factor();
+        assert!(lde_domain_size <= u32::MAX as usize, "LDE domain size too big");
+
         Context {
             trace_layout: trace_info.layout().clone(),
-            trace_length: trace_info.length(),
+            trace_length,
             trace_meta: trace_info.meta().to_vec(),
             field_modulus_bytes: B::get_modulus_le_bytes(),
             options,
@@ -117,7 +129,7 @@ impl<E: StarkField> ToElements<E> for Context {
 
         // convert proof options and trace length to elements
         result.append(&mut self.options.to_elements());
-        result.push(E::from(self.trace_length as u64));
+        result.push(E::from(self.trace_length as u32));
 
         // convert trace metadata to elements; this is done by breaking trace metadata into chunks
         // of bytes which are slightly smaller than the number of bytes needed to encode a field
@@ -257,7 +269,7 @@ mod tests {
             BaseElement::from(1_u32),    // lower bits of field modulus
             BaseElement::from(u32::MAX), // upper bits of field modulus
             BaseElement::from(ext_fri),
-            BaseElement::from(grinding_factor as u32),
+            BaseElement::from(grinding_factor),
             BaseElement::from(blowup_factor as u32),
             BaseElement::from(num_queries as u32),
             BaseElement::from(trace_length as u32),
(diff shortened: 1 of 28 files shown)


math/src/field/f62/tests.rs
```diff
@@ -23,7 +23,7 @@ fn add() {
     assert_eq!(BaseElement::from(5u8), BaseElement::from(2u8) + BaseElement::from(3u8));
 
     // test overflow
-    let t = BaseElement::from(BaseElement::MODULUS - 1);
+    let t = BaseElement::new(BaseElement::MODULUS - 1);
```
plafer, line 26: Shouldn't `new()` return a `Result` (and internally run a `try_from()`)? Given that `u64` values over `M` are not valid.

air/src/proof/context.rs
```diff
@@ -27,10 +27,22 @@ impl Context {
...
     /// proof options.
+    ///
+    /// # Panics
+    /// Panics if either trace length or the LDE domain size implied by the trace length and the
+    /// blowup factor is greater then [u32::MAX].
     pub fn new<B: StarkField>(trace_info: &TraceInfo, options: ProofOptions) -> Self {
+        // TODO: return errors instead of panicking?
+
+        let trace_length = trace_info.length();
+        assert!(trace_length <= u32::MAX as usize, "trace length too big");
```
plafer, line 38: nit: I would use `try_into()` to convert `u32` to `usize`. For architectures where `usize < u32`, `u32::MAX as usize` will truncate, which probably doesn't make sense.

Follow-up: What is the reason for this check?

air/src/proof/context.rs
```diff
@@ -27,10 +27,22 @@ impl Context {
...
+
+        let lde_domain_size = trace_length * options.blowup_factor();
+        assert!(lde_domain_size <= u32::MAX as usize, "LDE domain size too big");
```
plafer, line 41: Similarly, `try_into()` is probably best here (depending on why `u32::MAX` is used as upper bound - see previous follow-up question)

air/src/proof/context.rs
```diff
@@ -117,7 +129,7 @@ impl<E: StarkField> ToElements<E> for Context {
 
         // convert proof options and trace length to elements
         result.append(&mut self.options.to_elements());
-        result.push(E::from(self.trace_length as u64));
+        result.push(E::from(self.trace_length as u32));
```
plafer, line 132: It's safer to avoid the `as` operator, which is very complex and can perform truncation. We should almost always use `try_into()`

math/src/fft/concurrent.rs
```diff
@@ -50,36 +50,47 @@ pub fn evaluate_poly_with_offset<B: StarkField, E: FieldElement<BaseField = B>>(
...
+pub fn interpolate_poly<B, E>(values: &mut [E], inv_twiddles: &[B])
 where
     B: StarkField,
     E: FieldElement<BaseField = B>,
 {
-    split_radix_fft(v, inv_twiddles);
-    let inv_length = E::inv((v.len() as u64).into());
-    v.par_iter_mut().for_each(|e| *e *= inv_length);
-    permute(v);
+    assert!(values.len() <= u32::MAX as usize, "too many values");
```
plafer, line 61: Similar comment about `as usize`

math/src/fft/concurrent.rs
```diff
@@ -50,36 +50,47 @@ pub fn evaluate_poly_with_offset<B: StarkField, E: FieldElement<BaseField = B>>(
...
+
+    split_radix_fft(values, inv_twiddles);
+    let inv_length = E::inv((values.len() as u32).into());
```
plafer, line 64: And `as u32`.

If we do this everywhere though, maybe we can address those in a different PR

plafer (commented):
I much prefer not having the `from()` implementation! But I wonder if we should make `new()` also return a `Result`?

Otherwise left a few nits

math/src/field/f62/tests.rs
irakliyk: One reason for keeping it this way is to provide a fast but "unsafe" method which would allow someone to instantiate a field element when they know for sure that the underlying data is valid. So, given a `u64` we have two options:

- If we don't know whether the underlying data is valid, we should use `try_from()`.
- If we do know that the underlying data is valid, we can use `new()`.

But this behavior should be properly documented - so, I added a note about silent conversion to this method.

air/src/proof/context.rs
irakliyk: Generally agreed - but two thoughts:
1. I think it may be OK for the purposes of this library to assume that `usize` is at least 32 bits (and if so, we an enforce this at compile time).
2. If we do want to make sure that the library works on 16-bit architectures, there are many more places which we'd need to update, and this probably deserves a separate PR.

Regarding this check: we currently assume that trace length can be at most $2^{32}$ rows long. Lifting this restriction will also require changes in a few places.

PR #244, opened 2024-02-06
Add `TryFrom<u64>` and `TryFrom<u128>` bounds to the `FieldElement` trait

Overcastan:
This small PR implements the `TryFrom<u64>` and `TryFrom<u128>` bounds for the structs which implements `FieldElement` trait, namely: 
- BaseElement(u64)
- BaseElement(u128)
- QuadExtension
- CubeExtension

### DIFF
--- a/math/src/field/extensions/cubic.rs
+++ b/math/src/field/extensions/cubic.rs
@@ -11,8 +11,10 @@ use core::{
     slice,
 };
 use utils::{
-    collections::Vec, string::ToString, AsBytes, ByteReader, ByteWriter, Deserializable,
-    DeserializationError, Randomizable, Serializable, SliceReader,
+    collections::Vec,
+    string::{String, ToString},
+    AsBytes, ByteReader, ByteWriter, Deserializable, DeserializationError, Randomizable,
+    Serializable, SliceReader,
 };
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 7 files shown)


math/src/field/traits.rs
```diff
@@ -50,6 +50,8 @@ pub trait FieldElement:
     + From<u16>
     + From<u8>
     + for<'a> TryFrom<&'a [u8]>
+    + TryFrom<u64>
+    + TryFrom<u128>
```
irakliyk, line 54: minor nit: I'd move these lines above `for<'a> TryFrom<&'a [u8]>`.

math/src/field/extensions/quadratic.rs
```diff
@@ -313,6 +315,34 @@ impl<B: ExtensibleField<2>> From<u8> for QuadExtension<B> {
     }
 }
 
+impl<B: ExtensibleField<2>> TryFrom<u64> for QuadExtension<B> {
+    type Error = String;
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        let felt = B::try_from(value);
```
irakliyk, line 322: minor nit: I'd probably name the variable `base_element` to more explicit. Or maybe just put this line directly into the match statement.

But also, can we not do `B::try_from(value)?;` to avoid the match statement below?

math/src/field/extensions/quadratic.rs
```diff
@@ -313,6 +315,34 @@ impl<B: ExtensibleField<2>> From<u8> for QuadExtension<B> {
...
+        }
+    }
+}
+
+impl<B: ExtensibleField<2>> TryFrom<u128> for QuadExtension<B> {
+    type Error = String;
+
+    fn try_from(value: u128) -> Result<Self, Self::Error> {
+        let felt = B::try_from(value);
+        match felt {
```
irakliyk, line 337: Same comment as above.

math/src/field/extensions/cubic.rs
```diff
@@ -319,6 +321,34 @@ impl<B: ExtensibleField<3>> From<u8> for CubeExtension<B> {
     }
 }
 
+impl<B: ExtensibleField<3>> TryFrom<u64> for CubeExtension<B> {
+    type Error = String;
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        let felt = B::try_from(value);
+        match felt {
```
irakliyk, line 329: Same comment as above.

math/src/field/extensions/cubic.rs
```diff
@@ -319,6 +321,34 @@ impl<B: ExtensibleField<3>> From<u8> for CubeExtension<B> {
...
+        }
+    }
+}
+
+impl<B: ExtensibleField<3>> TryFrom<u128> for CubeExtension<B> {
+    type Error = String;
+
+    fn try_from(value: u128) -> Result<Self, Self::Error> {
+        let felt = B::try_from(value);
+        match felt {
```
irakliyk, line 343: Same comment as above.

irakliyk (commented):
Looks good! Thank you! I left a few small comments/questions inline.

math/src/field/extensions/quadratic.rs
Overcastan: I decided to use `B::try_from(value)?;` because without it I have to manage the field modulus `M` constant, which is different for different `B`, so I'm not sure that we can avoid it

irakliyk (approved):
Looks good! Thank you!

Issue #240, opened 2024-01-05
Accomodating more expressive transition constraints


Al-Kindi-0:
Currently, support for more general transition constraints is lacking. Consider the case of the constraints enforced on the so-called Lagrange kernel in Appendix A [here](https://eprint.iacr.org/2023/1284.pdf). These boil down to:

1. The boundary constraint $c(X) -  (1 - r_0)\dots (1 - r_{\nu-1}) = 0$ at $X = 1$, where $\nu = \log_2(n)$ and $n$ is the trace length.
2. For each  $`{\kappa}\in\lbrace 1,\dots, \nu\rbrace`$, the transition constraint $`$`r_{\nu - \kappa}\cdot c(X) - (1 -r_{\nu - \kappa})\cdot c(g^{2^{\nu - \kappa} }\cdot X) = 0`$`$ holds on $2^{\kappa - 1}$-th roots of unity i.e. with divisor $X^{2^{\kappa - 1}  } - 1$.

Thus we have $\log_2(n)$ constraints each with its own divisor i.e., enforcement domain. I haven't included exception points to keep things simple, but these should be straightforward to accommodate, I believe.

Now the above introduces certain challenges to the way we currently do things. For example, for the OOD evaluations, we now need to provide, in addition to $c(z)$ and $c(g\cdot z)$, $c(g^{2^{\nu - \kappa} }\cdot z)$ for $`{\kappa}\in\lbrace 1,\dots, \nu - 1\rbrace`$.

The question now is: should we generalize the code base from the ground up or is there a way to "delegate" accommodating such relatively rare cases to a per case custom solution?

plafer:
I have a few preliminary thoughts on this. At a high-level, this design generalizes the codebase, while making the most common case of "transition constraint applies over the entire domain" convenient.

We would introduce a new object `TransitionConstraintGroup` that would encode a set of transition constraints who share a common enforcement domain. It would look *roughly* like:

```rs
struct TransitionConstraintGroup<A: Air> {
  // Note: each transition function *returns* its evaluation
  transition_functions: Vec<fn(&A, &EvaluationFrame, &[E]) -> E>,
  // `Domain` would be a type similar to the current `ConstraintDivisor`,
  // perhaps more user-friendly
  enforcement_domain: Domain
}

impl<A: Air> TransitionConstraintGroup<A: Air> {
  // Most commonly used constructor. Domain defaults to full trace length (minus 1).
  fn new(transition_functions: ...) -> Self { ... }
[...]

PR #245, opened 2024-02-15
Merge `TraceLayout` into `TraceInfo`

plafer:
This is in preparation for #240.

The initial reason for this PR was to prepare for the introduction of the Lagrange kernel. Specifically, `TraceLayout` stored the [number of random elements](https://github.com/facebook/winterfell/blob/2fa2d631e1516a8a9044496a97327e3d5d23954a/air/src/air/trace_info.rs#L161) `aux_segment_rands` needed by the auxiliary trace; with the Lagrange kernel transition constraint, this number is a function of the trace length (specifically `log(trace_length)`). So the trace length should be stored in the same struct as `aux_segment_rands`.

Ultimately I found that the cleanest solution was to collapse the `TraceLayout` into the `TraceInfo`. As an aside, a "trace layout" without all the layout information (i.e. missing trace length) didn't make sense to me. I also found that this change cleaned things up, specifically since `TraceInfo` was stored "in parts" in multiple places (new to a `TraceLayout`) - we now properly store a single `TraceInfo` object in these cases.

This change also remove a bunch of methods across the board, which in my opinion simplifies the general API. For example, the only way to get a trace length now is through a `TraceInfo` object (unless I missed some), and all traces return their `TraceInfo`.


prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -203,8 +203,8 @@ where
     }
 
     /// Returns the trace layout of the execution trace.
-    fn trace_layout(&self) -> &TraceLayout {
-        self.trace_info.layout()
+    fn trace_info(&self) -> &TraceInfo {
```
irakliyk, line 206: Comment for this method should be updated.

prover/src/trace/trace_lde/mod.rs
```diff
@@ -67,5 +67,5 @@ pub trait TraceLde<E: FieldElement>: Sync {
     fn blowup(&self) -> usize;
 
     /// Returns the trace layout of the execution trace.
-    fn trace_layout(&self) -> &TraceLayout;
+    fn trace_info(&self) -> &TraceInfo;
```
irakliyk, line 70: Comment for this method should be updated.

irakliyk (changes requested):
Thank you! Looks good! I left some comments inline - but nothing major.

The original motivation for having `TraceInfo` and `TraceLayout` be separate was that `TraceLayout` was supposed to be independent from a specific instance of a computation (i.e., would not depend on trace length). But I don't think this is actually needed in practice - so, merging them is a good improvement.

prover/src/trace/mod.rs
```diff
@@ -43,15 +43,8 @@ pub trait Trace: Sized {
 
     // REQUIRED METHODS
     // --------------------------------------------------------------------------------------------
-
-    /// Returns a description of how columns of this trace are arranged into trace segments.
-    fn layout(&self) -> &TraceLayout;
-
-    /// Returns the number of rows in this trace.
-    fn length(&self) -> usize;
```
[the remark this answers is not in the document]
plafer: Agreed; reintroduced it

math/src/utils/mod.rs
```diff
@@ -199,6 +199,25 @@ pub fn log2(n: usize) -> u32 {
...
 
+// CONVERSION FUNCTIONS
+// ================================================================================================
+
+/// Converts a slice of bytes into a field element. Pads the slice if it is smaller than the number
+/// of bytes needed to represent an element.
+///
+/// # Panics
+/// Panics if the length of `bytes` is smaller than the number of bytes needed to encode an element.
+pub fn bytes_to_element_with_padding<B: StarkField>(mut bytes: Vec<u8>) -> B {
```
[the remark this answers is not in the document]
plafer: Good idea - named it `from_byte_vec_with_padding()` to make it clear that it tries to add padding before conversion. Otherwise I found it confusing why that method existed on top of the `TryFrom` conversion from slice to `StarkField`

plafer:
@irakliyk should we fix the new redundant imports clippy errors in another PR? There are a lot of them (more than the CI shows), and it might require a change in how `core_utils` works (not sure).

math/src/field/traits.rs
```diff
@@ -258,11 +273,21 @@ pub trait StarkField: FieldElement<BaseField = Self> {
         Self::TWO_ADIC_ROOT_OF_UNITY.exp(power)
     }
 
-    /// Returns byte representation of the field modulus in little-endian byte order.
-    fn get_modulus_le_bytes() -> Vec<u8>;
+    /// Converts a slice of bytes into a field element. Pads the slice if it is smaller than the number
+    /// of bytes needed to represent an element.
+    ///
+    /// # Panics
+    /// Panics if the length of `bytes` is smaller than the number of bytes needed to encode an element.
```
irakliyk, line 280: We panic if the number of bytes is greater than (not smaller than) the number of bytes for the field elements, right?

Also, another possible panic condition is when the number of bytes is correct but the value contained in these bytes is not a valid field element, right?

air/src/proof/context.rs
```diff
@@ -109,36 +90,24 @@ impl<E: StarkField> ToElements<E> for Context {
     /// Converts this [Context] into a vector of field elements.
     ///
     /// The elements are laid out as follows:
-    /// - trace layout info [1 or more elements].
+    /// - trace info [1 or more elements].
```
irakliyk, line 93: Trace info takes "2 or more elements", right? (because it will have main segment info, and also now trace length).

math/src/field/traits.rs
```diff
@@ -230,6 +230,9 @@ pub trait FieldElement:
 /// the modulus of the field should be a prime number of the form `k` * 2^`n` + 1 (a Proth prime),
 /// where `n` is relatively large (e.g., greater than 32).
 pub trait StarkField: FieldElement<BaseField = Self> {
+    // CONSTANTS
+    // =============================================================================================
```
irakliyk, line 234: minor nit: we usually use `----` for internal section separators instead of `=====` (also applicable to other changes in this file).

math/src/field/traits.rs
```diff
@@ -258,11 +273,21 @@ pub trait StarkField: FieldElement<BaseField = Self> {
...
 
-    /// Returns a canonical integer representation of this field element.
-    fn as_int(&self) -> Self::PositiveInteger;
+        bytes.resize(Self::ELEMENT_BYTES, 0);
+
+        match Self::try_from(bytes.as_slice()) {
+            Ok(element) => element,
+            Err(_) => panic!("element deserialization failed"),
+        }
+    }
```
irakliyk, line 290: Potential optimization: I wonder if we should change the function signature to:
```Rust
fn from_bytes_with_padding(bytes: &[u8]) -> Self
```
This would avoid the need to call `to_vec()` for the users of this function which could save on heap allocations. Internally, we could use an array buffer like so:
```Rust
let mut buf = [0; Self::ELEMENT_BYTES];
buf[..bytes.len()].copy_from_slice(bytes);
```

air/src/air/trace_info.rs
```diff
@@ -288,13 +247,25 @@ impl<E: StarkField> ToElements<E> for TraceLayout {
             result.push(E::from(buf));
         }
 
+        result.push(E::from(self.trace_length as u32));
```
irakliyk, line 250: Let's add a comment here to state that we assume that trace length is never greater than $2^{32}$.

irakliyk (approved):
Looks good! Thank you! I left a few minor comments inline - but other than that, this is good to merge.

irakliyk:
> @irakliyk should we fix the new redundant imports clippy errors in another PR? There are a lot of them (more than the CI shows), and it might require a change in how `core_utils` works (not sure).

Yes - let's address these in a separate PR.

math/src/field/traits.rs
plafer: Oops; right.

Fixed

math/src/field/traits.rs
plafer: Ah yes - fixed

math/src/field/traits.rs
plafer: This gives me this error:

```
error: constant expression depends on a generic parameter
   --> math/src/field/traits.rs:286:27
    |
286 |         let mut buf = [0; Self::ELEMENT_BYTES];
    |                           ^^^^^^^^^^^^^^^^^^^
    |
    = note: this may fail depending on what value the parameter takes
```

I did a little digging, and this seems to be a part of Rust that is still rough around the edges. IIUC, the problem is currently that there is no syntax to write the trait bounds for `Self::ELEMENT_BYTES` that the compiler needs to ensure soundness.

air/src/air/trace_info.rs
plafer: done

math/src/field/traits.rs
irakliyk: I would probably still change the method signature and do `.to_vec()` internally. This way, when we can do it properly in the future, we can make the change only in one place.

plafer: fixed

plafer:
@irakliyk everything should be good to go now

irakliyk (approved):
All looks good! Thank you!

PR #247, opened 2024-02-16
Implement Lagrange kernel constraints

plafer:
Closes: #240

Builds on: #245

Done:
- failing test with `Air` which uses a Lagrange kernel auxiliary column
- verifier 

Left to do:
- [x] Prover populates `OodFrame` with Lagrange kernel "frame"
- [x] Prover evaluates Lagrange kernel transition and boundary constraints
- [x] Cleanup (including revisiting current API)

### DIFF
--- a/air/src/air/boundary/constraint.rs
+++ b/air/src/air/boundary/constraint.rs
@@ -88,22 +88,6 @@ where
         }
     }
 
-    /// Creates a new boundary constraint from the specified single assertion.
-    ///
-    /// # Panics
-    /// Panics if the assertion is not a single assertion (i.e. `assertion.values` has more than 1
-    /// value)
-    pub fn new_single(assertion: Assertion<F>, composition_coefficient: E) -> Self {
-        assert_eq!(assertion.values.len(), 1);
-
-        BoundaryConstraint {
-            column: assertion.column,
-            poly: assertion.values,
-            poly_offset: (0, F::BaseField::ONE),
-            cc: composition_coefficient,
-        }
-    }
-
     // PUBLIC ACCESSORS
     // --------------------------------------------------------------------------------------------
 
--- a/air/src/air/coefficients.rs
+++ b/air/src/air/coefficients.rs
@@ -66,7 +66,7 @@ impl<E: FieldElement> Default for AuxTraceRandElements<E> {
 pub struct ConstraintCompositionCoefficients<E: FieldElement> {
     pub transition: Vec<E>,
     pub boundary: Vec<E>,
-    pub lagrange: LagrangeConstraintsCompositionCoefficients<E>,
+    pub lagrange: Option<LagrangeConstraintsCompositionCoefficients<E>>,
 }
 
 /// Stores the constraint composition coefficients for the Lagrange kernel transition and boundary
@@ -74,7 +74,7 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
 #[derive(Debug, Clone)]
 pub struct LagrangeConstraintsCompositionCoefficients<E: FieldElement> {
     pub transition: Vec<E>,
-    pub boundary: Option<E>,
+    pub boundary: E,
 }
 // DEEP COMPOSITION COEFFICIENTS
 // ================================================================================================
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -209,7 +209,7 @@ impl<B: StarkField> AirContext<B> {
     /// The number of transition constraints is defined by the total number of transition constraint
     /// degree descriptors (for both the main and the auxiliary trace constraints). This number is
     /// used to determine how many transition constraint coefficients need to be generated for
-    /// merging transition constraints into a composition polynomial.
+    /// merging transition constraints into a constraint composition polynomial.
     pub fn num_transition_constraints(&self) -> usize {
         self.main_transition_constraint_degrees.len() + self.aux_transition_constraint_degrees.len()
     }
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -103,8 +103,6 @@ impl<B: StarkField> AirContext<B> {
         assert!(num_main_assertions > 0, "at least one assertion must be specified");
 
         if trace_info.is_multi_segment() {
-            // If the only auxiliary column is the Lagrange kernel one, then we don't require any
-            // other boundary/transition constraints
             assert!(
                 !aux_transition_constraint_degrees.is_empty(),
                 "at least one transition constraint degree must be specified for auxiliary trace segments"
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -177,7 +177,7 @@ impl<B: StarkField> AirContext<B> {
 
     /// Returns length of the execution trace for an instance of a computation.
     ///
-    // This is guaranteed to be a power of two greater than or equal to 8.
+    /// This is guaranteed to be a power of two greater than or equal to 8.
     pub fn trace_len(&self) -> usize {
         self.trace_info.length()
     }
--- a/air/src/air/divisor.rs
+++ b/air/src/air/divisor.rs
@@ -44,20 +44,24 @@ impl<B: StarkField> ConstraintDivisor<B> {
     ///
     /// For transition constraints, the divisor polynomial $z(x)$ is always the same:
     ///
-    /// $$
-    /// z(x) = \frac{x^n - 1}{ \prod_{i=1}^k (x - g^{n-i})}
-    /// $$
+    /// $$ z(x) = \frac{x^n - 1}{ \prod_{i=1}^k (x - g^{n-i})} $$
     ///
-    /// where, $n$ is the length of the execution trace, $g$ is the generator of the trace
-    /// domain, and $k$ is the number of exemption points. The default value for $k$ is $1$.
+    /// where, $n$ is the length of the execution trace, $g$ is the generator of the trace domain,
+    /// and $k$ is the number of exemption points. The default value for $k$ is $1$.
     ///
     /// The above divisor specifies that transition constraints must hold on all steps of the
-    /// execution trace except for the last $k$ steps.
-    pub fn from_transition(trace_length: usize, num_exemptions: usize) -> Self {
-        let exemptions = (trace_length - num_exemptions..trace_length)
-            .map(|step| get_trace_domain_value_at::<B>(trace_length, step))
+    /// constraint enforcement domain except for the last $k$ steps. The constraint enforcement
+    /// domain is the entire trace in the case of transition constraints, but only a subset of the
+    /// trace for Lagrange kernel transition constraints.
+    pub fn from_transition(
+        constraint_enforcement_domain_size: usize,
+        num_exemptions: usize,
+    ) -> Self {
+        let exemptions = (constraint_enforcement_domain_size - num_exemptions
+            ..constraint_enforcement_domain_size)
+            .map(|step| get_trace_domain_value_at::<B>(constraint_enforcement_domain_size, step))
             .collect();
-        Self::new(vec![(trace_length, B::ONE)], exemptions)
+        Self::new(vec![(constraint_enforcement_domain_size, B::ONE)], exemptions)
     }
 
     /// Builds a divisor for a boundary constraint described by the assertion.
--- a/air/src/air/lagrange/boundary.rs
+++ b/air/src/air/lagrange/boundary.rs
@@ -33,7 +33,7 @@ where
         numerator / denominator
     }
 
-    /// Returns the evaluation of the boundary constraint numerator.
+    /// Returns the evaluation of the boundary constraint numerator, multiplied by the composition coefficient.
     ///
     /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)` for some `x`
     pub fn evaluate_numerator_at(&self, frame: &LagrangeKernelEvaluationFrame<E>) -> E {
--- a/air/src/air/lagrange/boundary.rs
+++ b/air/src/air/lagrange/boundary.rs
@@ -23,7 +23,7 @@ where
         }
     }
 
-    /// Returns the evaluation of the boundary constraint at `x`.
+    /// Returns the evaluation of the boundary constraint at `x`, multiplied by the composition coefficient.
     ///
     /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)`
     pub fn evaluate_at(&self, x: E, frame: &LagrangeKernelEvaluationFrame<E>) -> E {
--- a/air/src/air/lagrange/boundary.rs
+++ b/air/src/air/lagrange/boundary.rs
@@ -23,7 +23,8 @@ where
         }
     }
 
-    /// Returns the evaluation of the boundary constraint at `x`, multiplied by the composition coefficient.
+    /// Returns the evaluation of the boundary constraint at `x`, multiplied by the composition
+    /// coefficient.
     ///
     /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)`
     pub fn evaluate_at(&self, x: E, frame: &LagrangeKernelEvaluationFrame<E>) -> E {
@@ -33,9 +34,11 @@ where
         numerator / denominator
     }
 
-    /// Returns the evaluation of the boundary constraint numerator, multiplied by the composition coefficient.
+    /// Returns the evaluation of the boundary constraint numerator, multiplied by the composition
+    /// coefficient.
     ///
-    /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)` for some `x`
+    /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)` for
+    /// some `x`
     pub fn evaluate_numerator_at(&self, frame: &LagrangeKernelEvaluationFrame<E>) -> E {
         let trace_value = frame.inner()[0];
         let constraint_evaluation = trace_value - self.assertion_value;
--- a/air/src/air/lagrange/boundary.rs
+++ b/air/src/air/lagrange/boundary.rs
@@ -15,7 +15,7 @@ impl<E> LagrangeKernelBoundaryConstraint<E>
 where
     E: FieldElement,
 {
-    /// Creates a new Lagrange kernel boundary constraint from the specified single assertion.
+    /// Creates a new Lagrange kernel boundary constraint.
     pub fn new(composition_coefficient: E, lagrange_kernel_rand_elements: &[E]) -> Self {
         Self {
             assertion_value: Self::assertion_value(lagrange_kernel_rand_elements),
--- a/air/src/air/lagrange/mod.rs
+++ b/air/src/air/lagrange/mod.rs
@@ -8,7 +8,7 @@ mod transition;
 use math::FieldElement;
 pub use transition::LagrangeKernelTransitionConstraints;
 
-use crate::{AirContext, LagrangeConstraintsCompositionCoefficients};
+use crate::LagrangeConstraintsCompositionCoefficients;
 
 /// Represents the Lagrange kernel transition and boundary constraints.
 pub struct LagrangeKernelConstraints<E: FieldElement> {
@@ -20,14 +20,12 @@ pub struct LagrangeKernelConstraints<E: FieldElement> {
 impl<E: FieldElement> LagrangeKernelConstraints<E> {
     /// Constructs a new [`LagrangeKernelConstraints`].
     pub fn new(
-        context: &AirContext<E::BaseField>,
         lagrange_composition_coefficients: LagrangeConstraintsCompositionCoefficients<E>,
         lagrange_kernel_rand_elements: &[E],
         lagrange_kernel_col_idx: usize,
     ) -> Self {
         Self {
             transition: LagrangeKernelTransitionConstraints::new(
-                context,
                 lagrange_composition_coefficients.transition,
             ),
             boundary: LagrangeKernelBoundaryConstraint::new(
--- a/air/src/air/lagrange/mod.rs
+++ b/air/src/air/lagrange/mod.rs
@@ -14,6 +14,7 @@ use crate::{AirContext, LagrangeConstraintsCompositionCoefficients};
 pub struct LagrangeKernelConstraints<E: FieldElement> {
     pub transition: LagrangeKernelTransitionConstraints<E>,
     pub boundary: LagrangeKernelBoundaryConstraint<E>,
+    pub lagrange_kernel_col_idx: usize,
 }
 
 impl<E: FieldElement> LagrangeKernelConstraints<E> {
@@ -22,6 +23,7 @@ impl<E: FieldElement> LagrangeKernelConstraints<E> {
         context: &AirContext<E::BaseField>,
         lagrange_composition_coefficients: LagrangeConstraintsCompositionCoefficients<E>,
         lagrange_kernel_rand_elements: &[E],
+        lagrange_kernel_col_idx: usize,
     ) -> Self {
         Self {
             transition: LagrangeKernelTransitionConstraints::new(
@@ -32,6 +34,7 @@ impl<E: FieldElement> LagrangeKernelConstraints<E> {
                 lagrange_composition_coefficients.boundary,
                 lagrange_kernel_rand_elements,
             ),
+            lagrange_kernel_col_idx,
         }
     }
 }
--- a/air/src/air/lagrange/mod.rs
+++ b/air/src/air/lagrange/mod.rs
@@ -5,4 +5,33 @@ mod frame;
 pub use frame::LagrangeKernelEvaluationFrame;
 
 mod transition;
+use math::FieldElement;
 pub use transition::LagrangeKernelTransitionConstraints;
+
+use crate::{AirContext, LagrangeConstraintsCompositionCoefficients};
+
+/// Represents the Lagrange kernel transition and boundary constraints.
+pub struct LagrangeKernelConstraints<E: FieldElement> {
+    pub transition: LagrangeKernelTransitionConstraints<E>,
+    pub boundary: LagrangeKernelBoundaryConstraint<E>,
+}
+
+impl<E: FieldElement> LagrangeKernelConstraints<E> {
+    /// Constructs a new [`LagrangeKernelConstraints`].
+    pub fn new(
+        context: &AirContext<E::BaseField>,
+        lagrange_composition_coefficients: LagrangeConstraintsCompositionCoefficients<E>,
+        lagrange_kernel_rand_elements: &[E],
+    ) -> Self {
+        Self {
+            transition: LagrangeKernelTransitionConstraints::new(
+                context,
+                lagrange_composition_coefficients.transition,
+            ),
+            boundary: LagrangeKernelBoundaryConstraint::new(
+                lagrange_composition_coefficients.boundary,
+                lagrange_kernel_rand_elements,
+            ),
+        }
+    }
+}
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -382,13 +382,13 @@ pub trait Air: Send + Sync {
     fn get_boundary_constraints<E: FieldElement<BaseField = Self::BaseField>>(
         &self,
         aux_rand_elements: &AuxTraceRandElements<E>,
-        boundary_composition_coefficients: &[E],
+        composition_coefficients: &[E],
     ) -> BoundaryConstraints<E> {
         BoundaryConstraints::new(
             self.context(),
             self.get_assertions(),
             self.get_aux_assertions(aux_rand_elements),
-            boundary_composition_coefficients,
+            composition_coefficients,
         )
     }
 
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -526,10 +526,8 @@ pub trait Air: Send + Sync {
 
         let lagrange = if self.context().has_lagrange_kernel_aux_column() {
             let mut lagrange_kernel_t_coefficients = Vec::new();
-            if self.context().has_lagrange_kernel_aux_column() {
-                for _ in 0..self.context().trace_len().ilog2() {
-                    lagrange_kernel_t_coefficients.push(public_coin.draw()?);
-                }
+            for _ in 0..self.context().trace_len().ilog2() {
+                lagrange_kernel_t_coefficients.push(public_coin.draw()?);
             }
 
             let lagrange_kernel_boundary = public_coin.draw()?;
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -301,7 +301,6 @@ pub trait Air: Send + Sync {
     ) -> Option<LagrangeKernelConstraints<E>> {
         self.context().lagrange_kernel_aux_column_idx().map(|col_idx| {
             LagrangeKernelConstraints::new(
-                self.context(),
                 lagrange_composition_coefficients,
                 lagrange_kernel_rand_elements,
                 col_idx,
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -170,7 +170,7 @@ const MIN_CYCLE_LENGTH: usize = 2;
 /// trait:
 /// * The [AirContext] struct returned from [Air::context()] method must be instantiated using
 ///   [AirContext::new_multi_segment()] constructor. When building AIR context in this way, you
-///   will need to provide a [TraceLayout] which describes the shape of a multi-segment execution
+///   will need to provide a [`crate::TraceInfo`] which describes the shape of a multi-segment execution
 ///   trace.
 /// * Override [Air::evaluate_aux_transition()] method. This method is similar to the
 ///   [Air::evaluate_transition()] method but it also accepts two extra parameters:
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -25,7 +25,7 @@ pub use transition::{EvaluationFrame, TransitionConstraintDegree, TransitionCons
 
 mod lagrange;
 pub use lagrange::{
-    LagrangeKernelBoundaryConstraint, LagrangeKernelEvaluationFrame,
+    LagrangeKernelBoundaryConstraint, LagrangeKernelConstraints, LagrangeKernelEvaluationFrame,
     LagrangeKernelTransitionConstraints,
 };
 
@@ -291,6 +291,24 @@ pub trait Air: Send + Sync {
 
     // PROVIDED METHODS
     // --------------------------------------------------------------------------------------------
+
+    /// Returns a new [`LagrangeKernelConstraints`] if a Lagrange kernel auxiliary column is present
+    /// in the trace, or `None` otherwise.
+    fn get_lagrange_kernel_constraints<E: FieldElement<BaseField = Self::BaseField>>(
+        &self,
+        lagrange_composition_coefficients: LagrangeConstraintsCompositionCoefficients<E>,
+        lagrange_kernel_rand_elements: &[E],
+    ) -> Option<LagrangeKernelConstraints<E>> {
+        if self.context().has_lagrange_kernel_aux_column() {
+            Some(LagrangeKernelConstraints::new(
+                self.context(),
+                lagrange_composition_coefficients,
+                lagrange_kernel_rand_elements,
+            ))
+        } else {
+            None
+        }
+    }
 
     /// Returns values for all periodic columns used in the computation.
     ///
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -299,15 +299,14 @@ pub trait Air: Send + Sync {
         lagrange_composition_coefficients: LagrangeConstraintsCompositionCoefficients<E>,
         lagrange_kernel_rand_elements: &[E],
     ) -> Option<LagrangeKernelConstraints<E>> {
-        if self.context().has_lagrange_kernel_aux_column() {
-            Some(LagrangeKernelConstraints::new(
+        self.context().lagrange_kernel_aux_column_idx().map(|col_idx| {
+            LagrangeKernelConstraints::new(
                 self.context(),
                 lagrange_composition_coefficients,
                 lagrange_kernel_rand_elements,
-            ))
-        } else {
-            None
-        }
+                col_idx,
+            )
+        })
     }
 
     /// Returns values for all periodic columns used in the computation.
--- a/air/src/air/mod.rs
+++ b/air/src/air/mod.rs
@@ -32,12 +32,11 @@ pub use lagrange::{
 mod coefficients;
 pub use coefficients::{
     AuxTraceRandElements, ConstraintCompositionCoefficients, DeepCompositionCoefficients,
+    LagrangeConstraintsCompositionCoefficients,
 };
 
 mod divisor;
 pub use divisor::ConstraintDivisor;
-
-use self::coefficients::LagrangeConstraintsCompositionCoefficients;
 
 #[cfg(test)]
 mod tests;
@@ -504,33 +503,33 @@ pub trait Air: Send + Sync {
             t_coefficients.push(public_coin.draw()?);
         }
 
-        let mut lagrange_kernel_t_coefficients = Vec::new();
-        if self.context().has_lagrange_kernel_aux_column() {
-            for _ in 0..self.context().trace_len().ilog2() {
-                lagrange_kernel_t_coefficients.push(public_coin.draw()?);
-            }
-        }
-
         let mut b_coefficients = Vec::new();
         for _ in 0..self.context().num_assertions() {
             b_coefficients.push(public_coin.draw()?);
         }
 
-        let lagrange_kernel_boundary = {
+        let lagrange = if self.context().has_lagrange_kernel_aux_column() {
+            let mut lagrange_kernel_t_coefficients = Vec::new();
             if self.context().has_lagrange_kernel_aux_column() {
-                Some(public_coin.draw()?)
-            } else {
-                None
+                for _ in 0..self.context().trace_len().ilog2() {
+                    lagrange_kernel_t_coefficients.push(public_coin.draw()?);
+                }
             }
+
+            let lagrange_kernel_boundary = public_coin.draw()?;
+
+            Some(LagrangeConstraintsCompositionCoefficients {
+                transition: lagrange_kernel_t_coefficients,
+                boundary: lagrange_kernel_boundary,
+            })
+        } else {
+            None
         };
 
         Ok(ConstraintCompositionCoefficients {
             transition: t_coefficients,
             boundary: b_coefficients,
-            lagrange: LagrangeConstraintsCompositionCoefficients {
-                transition: lagrange_kernel_t_coefficients,
-                boundary: lagrange_kernel_boundary,
-            },
+            lagrange,
         })
     }
 
--- a/air/src/air/trace_info.rs
+++ b/air/src/air/trace_info.rs
@@ -288,10 +288,10 @@ impl Serializable for TraceInfo {
 }
 
 impl Deserializable for TraceInfo {
-    /// Reads [TraceLayout] from the specified `source` and returns the result.
+    /// Reads [`TraceInfo`] from the specified `source` and returns the result.
     ///
     /// # Errors
-    /// Returns an error of a valid [TraceLayout] struct could not be read from the specified
+    /// Returns an error of a valid [`TraceInfo`] struct could not be read from the specified
     /// `source`.
     fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
         let main_segment_width = source.read_u8()? as usize;
--- a/air/src/proof/context.rs
+++ b/air/src/proof/context.rs
@@ -215,9 +215,9 @@ mod tests {
             fri_remainder_max_degree as usize,
         );
         let trace_info = TraceInfo::new_multi_segment(
-            main_width as usize,
-            [aux_width as usize],
-            [aux_rands as usize],
+            main_width,
+            [aux_width],
+            [aux_rands],
             trace_length,
             vec![],
         );
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -112,7 +112,7 @@ impl OodFrame {
     ///
     /// # Errors
     /// Returns an error if:
-    /// * Valid [EvaluationFrame]s for the specified `main_trace_width` and `aux_trace_width`
+    /// * Valid [`crate::EvaluationFrame`]s for the specified `main_trace_width` and `aux_trace_width`
     ///   could not be parsed from the internal bytes.
     /// * A vector of evaluations specified by `num_evaluations` could not be parsed from the
     ///   internal bytes.
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -62,8 +62,8 @@ impl OodFrame {
 
         let mut result = vec![];
         for col in 0..trace_states.num_columns() {
-            result.push(trace_states.current_frame[col]);
-            result.push(trace_states.next_frame[col]);
+            result.push(trace_states.current_row[col]);
+            result.push(trace_states.next_row[col]);
         }
 
         // there are 2 frames: current and next
@@ -218,8 +218,8 @@ impl Deserializable for OodFrame {
 /// will be evaluated at `z`, `gz`, `g^2 z`, ... `g^(2^(v-1)) z`, where `v == log(trace_len)`, and
 /// stored in `lagrange_kernel_frame`.
 pub struct OodFrameTraceStates<E: FieldElement> {
-    current_frame: Vec<E>,
-    next_frame: Vec<E>,
+    current_row: Vec<E>,
+    next_row: Vec<E>,
     lagrange_kernel_frame: Option<LagrangeKernelEvaluationFrame<E>>,
 }
 
@@ -233,25 +233,25 @@ impl<E: FieldElement> OodFrameTraceStates<E> {
         assert_eq!(current_frame.len(), next_frame.len());
 
         Self {
-            current_frame,
-            next_frame,
+            current_row: current_frame,
+            next_row: next_frame,
             lagrange_kernel_frame,
         }
     }
 
     /// Returns the number of columns for the current and next frames.
     pub fn num_columns(&self) -> usize {
-        self.current_frame.len()
+        self.current_row.len()
     }
 
     /// Returns the current frame.
     pub fn current_frame(&self) -> &[E] {
-        &self.current_frame
+        &self.current_row
     }
 
     /// Returns the next frame.
     pub fn next_frame(&self) -> &[E] {
-        &self.next_frame
+        &self.next_row
     }
 
     /// Returns the Lagrange kernel frame, if any.
--- a/prover/src/constraints/evaluator/boundary.rs
+++ b/prover/src/constraints/evaluator/boundary.rs
@@ -34,11 +34,10 @@ impl<E: FieldElement> BoundaryConstraints<E> {
     pub fn new<A: Air<BaseField = E::BaseField>>(
         air: &A,
         aux_rand_elements: &AuxTraceRandElements<E>,
-        boundary_composition_coefficients: &[E],
+        composition_coefficients: &[E],
     ) -> Self {
         // get constraints from the AIR instance
-        let source =
-            air.get_boundary_constraints(aux_rand_elements, boundary_composition_coefficients);
+        let source = air.get_boundary_constraints(aux_rand_elements, composition_coefficients);
 
         // initialize a map of twiddles here so that we can keep track of already computed
         // twiddles; this helps us avoid building twiddles over and over again for constraints
--- a/prover/src/constraints/evaluator/default.rs
+++ b/prover/src/constraints/evaluator/default.rs
@@ -322,7 +322,7 @@ where
                 );
             }
 
-            evaluator.evaluate_lagrange_kernel_constraints_2::<A, T>(
+            evaluator.evaluate_lagrange_kernel_constraints::<A, T>(
                 trace,
                 lagrange_kernel_aux_column_idx,
                 domain,
--- a/prover/src/constraints/evaluator/default.rs
+++ b/prover/src/constraints/evaluator/default.rs
@@ -10,7 +10,7 @@ use super::{
 };
 use air::{
     Air, AuxTraceRandElements, ConstraintCompositionCoefficients, EvaluationFrame,
-    LagrangeKernelEvaluationFrame, TransitionConstraints,
+    TransitionConstraints,
 };
 use alloc::vec::Vec;
 use math::FieldElement;
@@ -306,21 +306,6 @@ where
                 .context()
                 .lagrange_kernel_aux_column_idx()
                 .expect("expected Lagrange kernel aux column index to be present");
-
-            // this will be used to convert steps in constraint evaluation domain to steps in
-            // LDE domain
-            let lde_shift = domain.ce_to_lde_blowup().trailing_zeros();
-
-            let mut lagrange_kernel_column_frames =
-                vec![LagrangeKernelEvaluationFrame::<E>::new_empty(); domain.ce_domain_size()];
-
-            for (step, frame) in lagrange_kernel_column_frames.iter_mut().enumerate() {
-                trace.read_lagrange_kernel_frame_into(
-                    step << lde_shift,
-                    lagrange_kernel_aux_column_idx,
-                    frame,
-                );
-            }
 
             evaluator.evaluate_lagrange_kernel_constraints(
                 trace,
--- a/prover/src/constraints/evaluator/default.rs
+++ b/prover/src/constraints/evaluator/default.rs
@@ -282,14 +282,26 @@ where
         }
     }
 
+    /// If present, evaluates the Lagrange kernel constraints over the constraint evaluation domain.
+    /// The evaluation of each constraint (both boundary and transition) is divided by its divisor,
+    /// multiplied by its composition coefficient, the result of which is added to
+    /// `combined_evaluations_accumulator`.
+    ///
+    /// Specifically, `combined_evaluations_accumulator` is a buffer whose length is the size of the
+    /// constraint evaluation domain, where each index contains combined evaluations of other
+    /// constraints in the system.
     fn evaluate_lagrange_kernel_constraints<T: TraceLde<E>>(
         &self,
         trace: &T,
         domain: &StarkDomain<A::BaseField>,
-        accumulator: &mut [E],
+        combined_evaluations_accumulator: &mut [E],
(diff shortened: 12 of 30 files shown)


winterfell/src/tests.rs
```diff
@@ -0,0 +1,206 @@
+use super::*;
+use prover::{
+    crypto::{hashers::Blake3_256, DefaultRandomCoin},
+    math::{fields::f64::BaseElement, FieldElement},
+    matrix::ColMatrix,
+};
+
+#[cfg(not(feature = "std"))]
+use alloc::{vec, vec::Vec};
```
plafer, line 9: Note: I wasn't able to make the imports work when using `winter-utils` as a `dev-dependency`, so I opted for this

plafer:
@irakliyk @Al-Kindi-0 this is ready for review!

The PR has a bigger diff than I originally anticipated. The main reason is that many existing abstractions did not apply to the Lagrange kernel column. For example, I had to create a `LagrangeKernelEvaluationFrame`, `LagrangeKernelBoundaryConstraint`, and `LagrangeKernelTransitionConstraints`, since the existing `EvaluationFrame`, `BoundaryConstraints` and `TransitionConstraints` made assumptions that didn't apply to the Lagrange kernel column. It would be nice to clean this up, but probably in another PR?

Related: #80

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,64 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
 }
+
+// LAGRANGE KERNEL EVALUATION FRAME
+// ================================================================================================
+
+/// The evaluation frame for the Lagrange kernel.
+///
+/// The Lagrange kernel's evaluation frame is different from [`EvaluationFrame`]. Specifically,
+/// - it only contains evaluations from the Lagrange kernel column
+///   (compared to all rows in the case of [`EvaluationFrame`])
```
Al-Kindi-0, line 90: should this be "all columns"?

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,64 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
+/// - The column is evaluated at points `z`, `gz`, `g^2 z`, ..., `g^(2^(v-1)) z`,
+///   where `z` is an arbitrary point, and `g` is the trace domain generator
```
Al-Kindi-0, line 92: We could just say that z is the OOD point to be more specific

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,64 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
+#[derive(Debug, Clone)]
+pub struct LagrangeKernelEvaluationFrame<E: FieldElement> {
+    frame: Vec<E>,
+}
+
+impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
+    // CONSTRUCTORS
+    // --------------------------------------------------------------------------------------------
+
+    /// Constructs a Lagrange kernel evaluation frame from the raw column polynomial evaluations
```
Al-Kindi-0, line 102: minor nit: evaluations -> evaluations.

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,64 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
+    pub fn new(frame: Vec<E>) -> Self {
+        Self { frame }
+    }
+
+    /// Constructs the frame from the Lagrange kernel trace column polynomial coefficients, and an evaluation point.
```
Al-Kindi-0, line 107: minor nit: "..., and an evaluation" -> "... for an evaluation point"

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,64 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
+            frame.push(lagrange_poly_at_x);
+        }
+
+        Self { frame }
+    }
+
+    // ACCESSORS
+    // --------------------------------------------------------------------------------------------
+
+    /// Returns a reference to the inner frame
```
Al-Kindi-0, line 131: minor nit: frame -> frame.

air/src/air/transition/mod.rs
```diff
@@ -171,3 +173,64 @@ impl<E: FieldElement> TransitionConstraints<E> {
...
     }
 }
+
+// LAGRANGE KERNEL TRANSITION CONSTRAINTS INFO
+// ================================================================================================
+
+/// Represents the transition constraints for the Lagrange kernel column, as well as the random
+/// coefficients used to linearly combine all the constraints.
+///
+/// There are `log(trace_len)` constraints, each with their own divisor, as described in
```
Al-Kindi-0, line 183: should it be "its" instead of "their"? Not sure to be honest

air/src/proof/ood_frame.rs
```diff
@@ -128,6 +169,10 @@ impl Serializable for OodFrame {
         target.write_u16(self.trace_states.len() as u16);
         target.write_bytes(&self.trace_states);
 
+        // write lagrange kernel column trace rows
```
Al-Kindi-0, line 172: nit: Lagrange

air/src/proof/ood_frame.rs
```diff
@@ -149,13 +194,68 @@ impl Deserializable for OodFrame {
         let num_trace_state_bytes = source.read_u16()? as usize;
         let trace_states = source.read_vec(num_trace_state_bytes)?;
 
+        // read lagrange kernel column trace rows
```
Al-Kindi-0, line 197: nit: Lagrange

air/src/proof/ood_frame.rs
```diff
@@ -149,13 +194,68 @@ impl Deserializable for OodFrame {
...
+            lagrange_kernel_trace_states,
             evaluations,
         })
     }
 }
+
+// OOD FRAME TRACE STATES
+// ================================================================================================
+
+/// Stores the trace evaluations evaluated at `z` and `gz`, where `z` is a random Field element. If
```
Al-Kindi-0, line 216: nit: we should probably say: "... trace evaluations at ..."

prover/src/constraints/evaluator/default.rs
```diff
@@ -107,8 +110,26 @@ where
         #[cfg(debug_assertions)]
         evaluation_table.validate_transition_degrees();
 
-        // combine all evaluations into a single column and return
-        evaluation_table.combine()
+        // combine all evaluations into a single column
+        let combined_evaluations = if self.air.trace_info().has_lagrange_kernel_aux_column() {
+            // if present, linearly combine the lagrange kernel evaluations too
```
Al-Kindi-0, line 115: nit: Lagrange

prover/src/constraints/evaluator/default.rs
```diff
@@ -257,6 +313,56 @@ where
...
+        // LDE domain
+        let lde_shift = domain.ce_to_lde_blowup().trailing_zeros();
+        for step in 0..domain.ce_domain_size() {
+            let lagrange_kernel_column_frame =
+                trace.get_lagrange_kernel_column_frame(step << lde_shift);
+
+            let domain_point = domain.get_ce_x_at(step);
+
+            let transition_constraints_combined_evals = {
+                let mut transition_evals = E::zeroed_vector(log2(domain.trace_length()) as usize);
```
Al-Kindi-0, line 337: could we move this outside the loop? would it even matter performance wise?

prover/src/lib.rs
```diff
@@ -382,7 +388,7 @@ pub trait Prover {
 
             // evaluate trace and constraint polynomials at the OOD point z, and send the results to
             // the verifier. the trace polynomials are actually evaluated over two points: z and
-            // z * g, where g is the generator of the trace domain.
+            // z * g, where g is the generator of the trace domain. Additionally, if the Lagrange kernel auxiliary column is present, we also evaluate that column over the points: z, z * g, z * g^2, z * g^4, ..., z * g^(2^(v-1)), where v = log(trace_len).
```
Al-Kindi-0, line 391: nit: could we fix the width of the comment?

Al-Kindi-0 (commented):
Looks great! Thank you @plafer !
Once we agree on #257 , we should come back and revisit some portions of this PR.

air/src/air/transition/frame.rs
plafer: Yup - fixed

air/src/air/transition/mod.rs
plafer: According to ChatGPT, it should be "its"

>The correct version is: "There are 42 constraints, each with its own divisor."
>
>In this sentence, "each" is referring to each constraint individually, so "its" is the appropriate possessive pronoun to use, as it agrees with the singular noun "constraint."

Fixed.

air/src/proof/ood_frame.rs
plafer: Agreed, fixed

prover/src/constraints/evaluator/default.rs
plafer: This is a tiny buffer, so it wouldn't affect performance. I kept the same pattern as `Air::evaluate_transition()` - i.e. zero out the buffer before passing it in the method. Technically, in both cases, zeroing out shouldn't matter because the method *should* touch every array index. IMO, whether we zero out the buffer or not, I mainly care that we be consistent.

Also, this is one of the reasons why I typically stay away from out parameters - this sort of ambiguity tends to create bugs.

air/src/air/transition/frame.rs
Al-Kindi-0: I see, it is used for both situations. Then I would probably use x instead of z in order to avoid confusions as z is, implicitly, used throughout the codebase to denote the OOD point.

irakliyk (commented):
Thank you! Looks good! This is not a full review yet - but I wanted to leave some preliminary comments.

One general observation: I wonder if it is possible to make Lagrange constraint evaluation a bit more "self-contained". For example, it may be possible to return something like this in the `Air` trait:
```Rust
trait Air {
    ...

    fn get_lagrange_kernel_evaluator(&self) -> Option<LagrangeKernelEvaluator>

    ...
}
```
This `LagrangeKernelEvaluator` would be responsible for evaluating the constraints for a given `LagrangeKernelEvaluationFrame`.

I haven't digested the code enough to say for sure if this is a good idea or not though.

air/src/air/divisor.rs
```diff
@@ -54,10 +54,6 @@ impl<B: StarkField> ConstraintDivisor<B> {
     /// The above divisor specifies that transition constraints must hold on all steps of the
     /// execution trace except for the last $k$ steps.
     pub fn from_transition(trace_length: usize, num_exemptions: usize) -> Self {
-        assert!(
-            num_exemptions > 0,
-            "invalid number of transition exemptions: must be greater than zero"
-        );
```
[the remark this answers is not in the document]
plafer: The Lagrange kernel transition constraints don't have any exemptions ([here](https://github.com/plafer/winterfell/blob/8544e17037c3d4c2391d4cded3693e161180aa48/air/src/air/transition/mod.rs#L206))

air/src/air/trace_info.rs
```diff
@@ -28,6 +28,7 @@ pub struct TraceInfo {
     main_segment_width: usize,
     aux_segment_widths: [usize; NUM_AUX_SEGMENTS],
     aux_segment_rands: [usize; NUM_AUX_SEGMENTS],
+    lagrange_kernel_aux_column_idx: Option<usize>,
```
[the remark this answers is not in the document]
plafer: I agree that it doesn't belong in `TraceInfo` - I probably put it there initially to solve some immediate problems.

However, I moved it to `AirContext`, since the Lagrange column index is an Air parameter. Otherwise `AirContext` is incomplete IMO: it holds main/aux transition degrees, the assertions, and trace information, but nothing about the Lagrange column.

Do you agree? I could also add a method `Air::lagrange_kernel_aux_column_idx()` that forwards to the internal context?

prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -8,7 +8,9 @@ use super::{
     TraceInfo, TraceLde, TracePolyTable,
 };
 use crate::{RowMatrix, DEFAULT_SEGMENT_WIDTH};
+use air::LagrangeKernelEvaluationFrame;
 use crypto::MerkleTree;
+use math::log2;
```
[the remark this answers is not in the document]
plafer: I removed the `log2()` function in math and switched everything to use `ilog2()`

air/src/air/mod.rs
```diff
@@ -285,6 +291,69 @@ pub trait Air: Send + Sync {
     // PROVIDED METHODS
     // --------------------------------------------------------------------------------------------
 
+    /// Evaluates the Lagrange kernel transition constraints over the specified Lagrange kernel
+    /// evaluation frame and stores them in `result`.
+    fn evaluate_lagrange_kernel_aux_transition<F, E>(
```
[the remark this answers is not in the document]
plafer: I agree. I called it `LagrangeKernelTransitionConstraints::evaluate_and_combine()`, since since `evaluate()` is used across repo to mean just evaluating the numerator

air/src/proof/ood_frame.rs
```diff
@@ -9,18 +9,25 @@ use utils::{
...
+use crate::LagrangeKernelEvaluationFrame;
 
 // OUT-OF-DOMAIN FRAME
 // ================================================================================================
+
+/// Represents an [`OodFrame`] where the trace and constraint evaluations have been parsed out.
+pub struct ParsedOodFrame<E> {
+    pub ood_trace_evaluations: Vec<E>,
+    pub ood_lagrange_kernel_trace_evaluations: Option<Vec<E>>,
+    pub ood_constraint_evaluations: Vec<E>,
```
irakliyk, line 21: minor nit: do we need the `ood` prefix here?

air/src/proof/ood_frame.rs
```diff
@@ -149,13 +194,68 @@ impl Deserializable for OodFrame {
...
 }
+
+// OOD FRAME TRACE STATES
+// ================================================================================================
+
+/// Stores the trace evaluations at `z` and `gz`, where `z` is a random Field element. If
+/// the Air contains a Lagrange kernel auxiliary column, then that column interpolated polynomial
+/// will be evaluated at `z`, `gz`, `g^2 z`, ... `g^(2^(v-1)) z`, where `v == log(trace_len)`, and
+/// stored in `lagrange_kernel_frame`.
+pub struct OodFrameTraceStates<E: FieldElement> {
```
irakliyk, line 220: Is this struct still used somewhere?

plafer: It is used [here](https://github.com/plafer/winterfell/blob/a3377d936b32b196e4a59a87ad39dc27cedffcb8/prover/src/trace/poly_table.rs#L80), [here](https://github.com/plafer/winterfell/blob/a3377d936b32b196e4a59a87ad39dc27cedffcb8/air/src/proof/ood_frame.rs#L57) and [here](https://github.com/plafer/winterfell/blob/a3377d936b32b196e4a59a87ad39dc27cedffcb8/prover/src/composer/mod.rs#L66)

air/src/proof/ood_frame.rs
plafer: Removed

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,65 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
+        let g = E::from(E::BaseField::get_root_of_unity(log_trace_len));
+
+        let mut frame = Vec::with_capacity(log_trace_len as usize + 1);
+
+        // push c(z)
+        frame.push(polynom::eval(lagrange_kernel_col_poly, z));
+
+        // push `c(gz)`, `c(z * g^2)`, `c(z * g^4)`, ..., `c(z * g^(2^(v-1)))`
+        for i in 0..log_trace_len {
+            let x = g.exp_vartime(2_u32.pow(i).into()) * z;
```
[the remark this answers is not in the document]
plafer: Actually we're increasing the power nonlinearly. The powers are: 1, 2, 4, 8, 16, ... . We could try to optimize this a bit, but we're only doing this `log(trace_len)` times, i.e. always a relatively few number of times. I think the added complexity wouldn't be worth the marginal performance improvement?

air/src/proof/ood_frame.rs
```diff
@@ -38,29 +46,48 @@ impl OodFrame {
...
+    /// rows interleaved. This is done so that reseeding of the random coin needs to be done only
+    /// once as opposed to once per each row.
     ///
     /// # Panics
     /// Panics if evaluation frame has already been set.
-    pub fn set_trace_states<E: FieldElement>(&mut self, trace_states: &[Vec<E>]) -> Vec<E> {
+    pub fn set_trace_states<E: FieldElement>(
+        &mut self,
+        trace_states: &OodFrameTraceStates<E>,
+    ) -> Vec<E> {
```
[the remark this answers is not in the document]
plafer: I actually like that an Ood frame contains both the "regular" current/next frames, as well as the Lagrange kernel frame. I think of it as one unit; all the data necessary to check the out-of-domain point in one place. I'm curious why you were thinking about separating them though.

> (with this, we may be able to avoid the need for `OodFrameTraceStates` struct)

How do you see we get rid of it? We'll still need to store the current & next frames.

But I also have as mental note to reduce the number of "ood frame" structs, but I was leaving it for another PR :).

plafer:
@irakliyk ready for another review! I still have the `read_lagrange_kernel_frame_into()` comment to address - which I agree with and will implement first thing tomorrow.

air/src/air/transition/frame.rs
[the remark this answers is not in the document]
irakliyk: Since this is used only for the OOD frame, using `exp_vartime()` is probably fine - but also, I don't think this would add much complexity.

If the powers are $g$, $g^2$, $g^4$, $g^8$ etc. wouldn't this be just:
```Rust
let mut g_exp = g;
for i in 0..log_trace_len {
    let x = g_exp * z;
    let lagrange_poly_at_x = polynom::eval(lagrange_kernel_col_poly, x);

    frame.push(lagrange_poly_at_x);
    g_exp *= g_exp;
}
```

air/src/air/mod.rs
```diff
@@ -285,6 +291,24 @@ pub trait Air: Send + Sync {
...
 
+    /// Returns the random elements to use in the lagrange kernel.
+    ///
+    /// The return slice stores the least significant random element first. For example, for a trace
+    /// length of 8, there will be 3 random elements, such that
+    /// - return[0] = r0
+    /// - return[1] = r1
+    /// - return[2] = r2
+    /// where `return` is the returned slice, and r = (r2, r1, r0).
+    fn lagrange_kernel_rand_elements<'r, F, E>(&self, aux_rand_elements: &'r [E]) -> &'r [E]
```
irakliyk, line 302: I'm not sure we need this function just yet. I think for now we should assume that $r$-values are exogenous to the system. In the next PR, we can figure out where these should come from.

air/src/air/mod.rs
```diff
@@ -357,13 +381,22 @@ pub trait Air: Send + Sync {
     fn get_boundary_constraints<E: FieldElement<BaseField = Self::BaseField>>(
         &self,
         aux_rand_elements: &AuxTraceRandElements<E>,
-        composition_coefficients: &[E],
+        boundary_composition_coefficients: &[E],
+        lagrange_kernel_boundary_coefficient: Option<E>,
     ) -> BoundaryConstraints<E> {
```
irakliyk, line 386: Do we need to integrate Lagrange kernel boundary constraints into other boundary constraints? In my initial thinking, we'd treat all Lagrange constraints separately. Specifically, instead of having 2 types of constraints as we currently do, we'd have:

1. Transition constraints (same as now).
2. Boundary constraints (same as now).
3. Lagrange kernel constraints - this would include boundary and transition constraints relevant to the Lagrange column.

verifier/src/channel.rs
```diff
@@ -407,11 +433,18 @@ impl<E: FieldElement> TraceOodFrame<E> {
...
+            {
                 current_aux[i] = a[0];
                 next_aux[i] = a[1];
             }
             Some(EvaluationFrame::from_rows(current_aux, next_aux))
         }
     }
+
+    /// Returns the Lagrange kernel evaluation frame, if any.
+    pub fn lagrange_kernel_column_frame(&self) -> Option<&LagrangeKernelEvaluationFrame<E>> {
```
irakliyk, line 447: minor nit: `lagrange_kernel_frame()` is probably sufficient (i.e., no need for "column").

prover/src/constraints/evaluator/default.rs
```diff
@@ -131,18 +152,44 @@ where
...
+            .lagrange_kernel_boundary
+            .map(|lagrange_kernel_boundary_coefficient| {
+                let lagrange_kernel_aux_rand_elements =
+                    air.lagrange_kernel_rand_elements(aux_rand_elements.get_segment_elements(0));
+
+                LagrangeKernelBoundaryConstraint::new(
+                    lagrange_kernel_boundary_coefficient,
+                    lagrange_kernel_aux_rand_elements,
+                )
+            });
```
irakliyk, line 185: Instead of separating Lagrange boundary and transition constraints here, I wonder if we should treat them as a single unit - e.g., `LagrangeKernelConstraints` instantiated something like:
```Rust
let lagrange_kernel_constraints =
 air.context().lagrange_kernel_aux_column_idx().map(|_| {
     LagrangeKernelConstraints::new(
         air.context(),
         lagrange_rand_elements, // the r-values
         composition_coefficients.lagrange,
     )
 });
```
And then we split into evaluating Lagrange boundary and transition constraints internally within this struct.

I guess the main question is how to get the $r$-values here.

prover/src/constraints/evaluator/default.rs
```diff
@@ -233,8 +280,19 @@ where
...
+                self.air.context().lagrange_kernel_aux_column_idx(),
+                self.air.trace_info(),
+            ) {
+                evaluations[0] += self.evaluate_aux_transition(
+                    &main_frame,
+                    &aux_frame,
+                    step,
+                    &mut ta_evaluations,
+                );
+            }
```
irakliyk, line 295: If there are no other aux constraints - shouldn't we just call `evaluate_fragment_main()` instead of calling `evaluate_fragment_full()` and then having this check?

irakliyk (commented):
Thank you! Looks good! Still not a full review - but I left some more comments inline.

air/src/air/coefficients.rs
```diff
@@ -65,7 +65,9 @@ impl<E: FieldElement> Default for AuxTraceRandElements<E> {
 #[derive(Debug, Clone)]
 pub struct ConstraintCompositionCoefficients<E: FieldElement> {
     pub transition: Vec<E>,
+    pub lagrange_kernel_transition: Vec<E>,
     pub boundary: Vec<E>,
+    pub lagrange_kernel_boundary: Option<E>,
 }
```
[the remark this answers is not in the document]
plafer: What benefits do you see in doing this? I typically don't like this style, since the type becomes much less descriptive, and more bug-prone. To understand what the `lagrange` field is, I have to go see how it is populated in the code. And I need to keep in my head that `lagrange[0]` is different from all other `lagrange[i]`, and lookout for bugs e.g. that loop over all indices and apply a "transition constraint operation".

air/src/air/transition/frame.rs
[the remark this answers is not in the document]
plafer: bleh, you're right. I changed it to use this code snippet.

prover/src/constraints/evaluator/default.rs
plafer: I since removed `trace_aux_segment_has_only_lagrange_kernel_column()`, since we don't want to support `Air` that only have a Lagrange kernel column (from previous comment). So the previous behavior where we call `evaluate_aux_transition()` unconditionally is now restored.

prover/src/trace/trace_lde/mod.rs
```diff
@@ -56,6 +56,13 @@ pub trait TraceLde<E: FieldElement>: Sync {
     /// Reads current and next rows from the auxiliary trace segment into the specified frame.
     fn read_aux_trace_frame_into(&self, lde_step: usize, frame: &mut EvaluationFrame<E>);
 
+    /// Returns the Lagrange kernel frame starting at the current row (as defined by `lde_step`).
+    ///
+    /// Note that unlike [`EvaluationFrame`], the Lagrange kernel frame includes only the Lagrange
+    /// kernel column (as opposed to all columns).
+    fn get_lagrange_kernel_column_frame(&self, lde_step: usize)
+        -> LagrangeKernelEvaluationFrame<E>;
```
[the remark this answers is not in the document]
plafer: done

air/src/air/boundary/constraint.rs
```diff
@@ -142,3 +162,54 @@ where
...
+            let trace_value = frame.inner()[0];
+            let constraint_evaluation = trace_value - self.assertion_value;
+
+            constraint_evaluation * self.composition_coefficient
+        };
+
+        let denominator = x - E::ONE;
+
+        numerator / denominator
+    }
```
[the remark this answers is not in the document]
plafer: done

air/src/air/context.rs
```diff
@@ -119,6 +132,14 @@ impl<B: StarkField> AirContext<B> {
             );
         }
 
+        // validate Lagrange kernel aux column, if any
+        if let Some(lagrange_kernel_aux_column_idx) = lagrange_kernel_aux_column_idx {
+            assert!(lagrange_kernel_aux_column_idx < trace_info.get_aux_segment_width(0), "Lagrange kernel column index out of bounds: index={}, but only {} columns in segment", lagrange_kernel_aux_column_idx, trace_info.get_aux_segment_width(0));
+
+            let min_aux_segment_rands = trace_info.length().ilog2();
+            assert!(trace_info.get_aux_segment_rand_elements(0) >= min_aux_segment_rands as usize, "Lagrange kernel column requires log(trace_length) random elements. Got: {}, but need at least {}", trace_info.get_aux_segment_rand_elements(0), min_aux_segment_rands);
```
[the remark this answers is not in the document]
plafer: Removed

air/src/air/mod.rs
plafer: removed

air/src/air/coefficients.rs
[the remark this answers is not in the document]
plafer: Replaced both fields with a single `LagrangeConstraintsCompositionCoefficients` one

plafer:
Added a benchmark that runs the prover on a trace with a Lagrange kernel column, with trace lengths `2^14`, `2^15`, `2^16`. On my Macbook Pro M2, the results were:

- trace length $2^{14}$: [426.12 ms 426.70 ms 427.21 ms]
- trace length $2^{15}$: [906.85 ms 908.47 ms 910.41 ms]
- trace length $2^{16}$: [1.9335 s 1.9348 s 1.9363 s]

prover/src/constraints/evaluator/default.rs
plafer: I think we should instead generalize the existing `TransitionConstraints` and `BoundaryConstraints`, as detailed in https://github.com/facebook/winterfell/issues/261

air/src/air/mod.rs
plafer: Similar to other comments, I think we should merge them with `TransitionConstraints` and `BoundaryConstraints` (https://github.com/facebook/winterfell/issues/261)

air/src/air/transition/mod.rs
```diff
@@ -171,3 +172,88 @@ impl<E: FieldElement> TransitionConstraints<E> {
...
+
+        transition_evaluations
+            .iter()
+            .zip(self.lagrange_constraint_coefficients.iter())
+            .zip(self.divisors.iter())
+            .fold(E::ZERO, |acc, ((&const_eval, &coef), divisor)| {
+                let z = divisor.evaluate_at(x);
+
+                acc + (coef.mul_base(const_eval) / z.into())
+            })
```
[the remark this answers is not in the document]
plafer: Fixed. I ended up encapsulating the logic in a `LagrangeConstraintsBatchEvaluator` that lives next to `DefaultConstraintEvaluator`. The idea is that this is an algorithm that is specific to the prover (hence lives in the prover crate), and runs on the CPU (hence lives next to `DefaultConstraintEvaluator`).

plafer:
@irakliyk this is ready for review!

Benchmarks after optimization:

- trace length $2^{14}$: [81.791 ms 81.859 ms 81.918 ms]
- trace length $2^{15}$: [164.98 ms 165.21 ms 165.57 ms]
- trace length $2^{16}$: [352.78 ms 353.17 ms 353.56 ms]

A 5x speedup!

air/src/air/mod.rs
irakliyk: I think trying to merge them into current `TransitionConstraints` and `BoundaryConstraints` will result in more complexity than required. It is also not clear to me that the strategies we use for evaluating (and optimizing) transition/boundary constraints will be applicable to Lagrange kernel constraints. So, as mention in the previous comment, my preference would be to keep Lagrange kernel constraints as encapsulated as possible and not treat them as special cases of boundary/transition constrains.

Of course, if later on we decided that it is a better approach, we can always refactor the codebase - but for now, I think it would be good to implement end-to-end handling of Lagrange column separately.

Concretely, we could add a method to the `Air` trait which would look something like this:
```Rust
fn get_lagrange_kernel_constraints<E: FieldElement<BaseField = Self::BaseField>>(
    &self,
    rand_elements: &Vec<E>, // these would be the r-values
    composition_coefficients: &LagrangeConstraintsCompositionCoefficients<E>,
) -> Option<LagrangeKernelConstraints<E>>
```
This method would return `None` if there is no Lagrange kernel column in the trace, and `LagrangeKernelConstraints` otherwise. `LagrangeKernelConstraints` would have something like `evaluate_at()` method, which would evaluate all Lagrange kernel constraints (boundary and transition) for a given frame. (it could also have other methods which would help the prover to perform evaluations in a more optimal way).

air/src/air/transition/frame.rs
```diff
@@ -77,3 +79,80 @@ impl<E: FieldElement> EvaluationFrame<E> {
...
+///
+/// The Lagrange kernel's evaluation frame is different from [`EvaluationFrame`]. Specifically,
+/// - it only contains evaluations from the Lagrange kernel column
+///   (compared to all columns in the case of [`EvaluationFrame`])
+/// - The column is evaluated at points `x`, `gx`, `g^2 x`, ..., `g^(2^(v-1)) x`,
+///   where `x` is an arbitrary point, and `g` is the trace domain generator
+#[derive(Debug, Clone)]
+pub struct LagrangeKernelEvaluationFrame<E: FieldElement> {
+    frame: Vec<E>,
+}
```
irakliyk, line 96: I would move this to `src/air/lagrange/frame.rs` module.

air/src/air/transition/mod.rs
```diff
@@ -171,3 +172,137 @@ impl<E: FieldElement> TransitionConstraints<E> {
...
+
+/// Represents the transition constraints for the Lagrange kernel column, as well as the random
+/// coefficients used to linearly combine all the constraints.
+///
+/// There are `log(trace_len)` constraints, each with its own divisor, as described in
+/// [this issue](https://github.com/facebook/winterfell/issues/240).
+pub struct LagrangeKernelTransitionConstraints<E: FieldElement> {
+    lagrange_constraint_coefficients: Vec<E>,
+    divisors: Vec<ConstraintDivisor<E::BaseField>>,
+}
```
irakliyk, line 187: I would move this to `src/air/lagrange/transition.rs` module.

Also, I'm not sure we need to use `ConstraintDivisor` structs here. The divisors for Lagrange kernel constraints are pretty simple and I think it might be simpler (and probably more efficient) to handle them in a specialized way here.

prover/src/constraints/evaluator/default.rs
```diff
@@ -130,19 +150,29 @@ where
...
+            &aux_rand_elements,
+            &composition_coefficients.boundary,
+            composition_coefficients.lagrange.boundary,
+        );
+
+        let lagrange_constraints_evaluator = LagrangeConstraintsBatchEvaluator::new(
+            air,
+            aux_rand_elements.clone(),
+            composition_coefficients,
+        );
```
irakliyk, line 169: I think using `aux_rand_elements` here is fine for now - but my current understanding is that the $r$-values for the Lagrange kernel column will be separate from these. But this is something we'll address in the next PR.

prover/src/constraints/evaluator/default.rs
```diff
@@ -39,6 +40,7 @@ pub struct DefaultConstraintEvaluator<'a, A: Air, E: FieldElement<BaseField = A:
     air: &'a A,
     boundary_constraints: BoundaryConstraints<E>,
     transition_constraints: TransitionConstraints<E>,
+    lagrange_constraints_evaluator: LagrangeConstraintsBatchEvaluator<E>,
```
irakliyk, line 43: I would make this an optional field - this way, we won't have to handle optionality inside `LagrangeConstraintsBatchEvaluator`.

prover/src/constraints/evaluator/default.rs
```diff
@@ -339,3 +405,191 @@ where
...
+
+/// Contains a specific strategy for evaluating the Lagrange kernel boundary and transition
+/// constraints where the divisors' evaluation is batched.
+///
+/// Specifically, [`batch_inversion`] is used to reduce the number of divisions performed.
+struct LagrangeConstraintsBatchEvaluator<E: FieldElement> {
+    lagrange_kernel_boundary_constraint: Option<LagrangeKernelBoundaryConstraint<E>>,
+    lagrange_kernel_transition_constraints: Option<LagrangeKernelTransitionConstraints<E>>,
+    aux_rand_elements: AuxTraceRandElements<E>,
+}
```
irakliyk, line 420: I would move this to `src/constraints/evaluator/lagrange.rs` module. Also, as mentioned in the previous comment, if we make the whole filed for `LagrangeConstraintsBatchEvaluator` optional, we won't need to make the internal fields optional.

Lastly, `aux_rand_elements` are the $r$-values, right? If so, I'd change this filed to `rand_elements: Vec<E>` as these values will be different from `AuxTraceRandElements`.

prover/src/constraints/evaluator/default.rs
```diff
@@ -257,6 +288,41 @@ where
...
+        let mut lagrange_kernel_column_frames =
+            vec![LagrangeKernelEvaluationFrame::<E>::new_empty(); domain.ce_domain_size()];
+
+        for (step, frame) in lagrange_kernel_column_frames.iter_mut().enumerate() {
+            trace.read_lagrange_kernel_frame_into(
+                step << lde_shift,
+                lagrange_kernel_aux_column_idx,
+                frame,
+            );
+        }
```
irakliyk, line 315: If I'm understanding this correctly, this would store a frame per row in the constraint evaluation domain, right? If so, this is probably too expensive from memory standpoint. For example, for a trace with $2^20$ columns, this would be equivalent to having extra 20 columns in the auxiliary trace, and this would significantly increase RAM requirements (e.g., in some cases doubling them).

This interferes a bit with batch inversion of denominators approach (would have been really nice if we could pre-compute them for all constraints in a single vector) - but I think we can still make it work if we change the methodology slightly: we still evaluate one constraint at a time, but instead of working with evaluation frames, we work directly with the trace (e.g., read elements needed only for the current constraint). Maybe there are other ways of doing this too.

irakliyk (changes requested):
Thank you! I left a few more comments inline. The main ones are about segregating Lagrange kernel constraints from the rest of the structure.

Also, could you rebase this from the latest `main` - handling of `no_std` changed there slight.

Regarding benchmarks: what exactly are these measuring (i.e., is it just Lagrange column constraint evaluation - or something else)? If this is just for evaluating constraints, we'll need to dig deeper into how to optimize this as it seems like the impact on prover performance would be quite bad (i.e., it will slow down proof generation for something like Miden VM by 10% - 20% where generating proof for $2^{16}$ trace takes about about 2.5 seconds on a single core).

plafer:
> Also, could you rebase this from the latest main - handling of no_std changed there slight.

I merged `main` in, but it seems like it made the diff against `next` bigger. Is this intended?

> Regarding benchmarks: what exactly are these measuring (i.e., is it just Lagrange column constraint evaluation - or something else)? If this is just for evaluating constraints, we'll need to dig deeper into how to optimize this as it seems like the impact on prover performance would be quite bad

It is [measuring the entire](https://github.com/plafer/winterfell/blob/40a3b06dd51f62cfbf5a3f536e91480981202a86/prover/benches/lagrange_kernel.rs#L33) `Prover::prove()` function - not just the Lagrange constraint evaluation. I did this because the prover API is likely more stable than any internal Lagrange kernel API for constraint evaluation - but I can change it to just benchmark the Lagrange constraint evaluation if that is more useful. I personally used it more as a means of knowing if any given commit did indeed improve general performance, as opposed to knowing the absolute Lagrange kernel constraint evaluation time.

air/src/air/mod.rs
plafer: I had separated them already; I had forgotten to update the `BoundaryConstraints` structs (the Lagrange constraints part was not actually used anymore). Now fixed.

The `LagrangeKernelConstraints` part of the comment still needs to be addressed though.

air/src/air/coefficients.rs
```diff
@@ -66,8 +66,16 @@ impl<E: FieldElement> Default for AuxTraceRandElements<E> {
...
+    pub lagrange: LagrangeConstraintsCompositionCoefficients<E>,
 }
 
+/// Stores the constraint composition coefficients for the Lagrange kernel transition and boundary
+/// constraints.
+#[derive(Debug, Clone)]
+pub struct LagrangeConstraintsCompositionCoefficients<E: FieldElement> {
+    pub transition: Vec<E>,
+    pub boundary: Option<E>,
+}
```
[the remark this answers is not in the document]
plafer: done

prover/src/constraints/evaluator/default.rs
plafer: done

irakliyk (changes requested):
Thank you! Looks good! This is not a full review - but I did leave some more comments inline.

air/src/air/divisor.rs
```diff
@@ -54,10 +54,6 @@ impl<B: StarkField> ConstraintDivisor<B> {
...
         let exemptions = (trace_length - num_exemptions..trace_length)
             .map(|step| get_trace_domain_value_at::<B>(trace_length, step))
             .collect();
```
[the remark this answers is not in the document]
plafer: This would re-introduce some Lagrange kernel concepts into the rest of the API (something we've been trying to avoid elsewhere). I also don't think this `assert!` is necessary, since `AirContext`'s constructor already checks it (e.g. for [main trace](https://github.com/plafer/winterfell/blob/9ea6c2df597e1fc110a93b6f94058de18a840b96/air/src/air/context.rs#L103)).

air/src/air/lagrange/boundary.rs
```diff
@@ -0,0 +1,60 @@
...
+    assertion_value: E,
+    composition_coefficient: E,
+}
+
+impl<E> LagrangeKernelBoundaryConstraint<E>
+where
+    E: FieldElement,
+{
+    /// Creates a new Lagrange kernel boundary constraint from the specified single assertion.
+    pub fn new(composition_coefficient: E, lagrange_kernel_rand_elements: &[E]) -> Self {
```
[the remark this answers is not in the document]
plafer: fixed

air/src/air/lagrange/boundary.rs
```diff
@@ -0,0 +1,60 @@
...
+        let numerator = self.evaluate_numerator_at(frame);
+        let denominator = self.evaluate_denominator_at(x);
+
+        numerator / denominator
+    }
+
+    /// Returns the evaluation of the boundary constraint numerator.
+    ///
+    /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)` for some `x`
+    pub fn evaluate_numerator_at(&self, frame: &LagrangeKernelEvaluationFrame<E>) -> E {
```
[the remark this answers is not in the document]
plafer: Updated

air/src/air/mod.rs
```diff
@@ -357,13 +382,13 @@ pub trait Air: Send + Sync {
...
+        boundary_composition_coefficients: &[E],
     ) -> BoundaryConstraints<E> {
         BoundaryConstraints::new(
             self.context(),
             self.get_assertions(),
             self.get_aux_assertions(aux_rand_elements),
-            composition_coefficients,
+            boundary_composition_coefficients,
         )
     }
```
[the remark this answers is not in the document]
plafer: fixed

air/src/proof/context.rs
```diff
@@ -215,9 +215,9 @@ mod tests {
             fri_remainder_max_degree as usize,
         );
         let trace_info = TraceInfo::new_multi_segment(
-            main_width,
-            [aux_width],
-            [aux_rands],
+            main_width as usize,
+            [aux_width as usize],
+            [aux_rands as usize],
```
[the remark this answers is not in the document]
plafer: Fixed. I would have assumed `clippy` would catch that

prover/src/constraints/evaluator/boundary.rs
```diff
@@ -34,10 +34,11 @@ impl<E: FieldElement> BoundaryConstraints<E> {
     pub fn new<A: Air<BaseField = E::BaseField>>(
         air: &A,
         aux_rand_elements: &AuxTraceRandElements<E>,
-        composition_coefficients: &[E],
+        boundary_composition_coefficients: &[E],
     ) -> Self {
         // get constraints from the AIR instance
-        let source = air.get_boundary_constraints(aux_rand_elements, composition_coefficients);
+        let source =
+            air.get_boundary_constraints(aux_rand_elements, boundary_composition_coefficients);
```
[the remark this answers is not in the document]
plafer: removed prefix

verifier/src/lib.rs
```diff
@@ -200,15 +200,17 @@ where
     let ood_trace_frame = channel.read_ood_trace_frame();
     let ood_main_trace_frame = ood_trace_frame.main_frame();
     let ood_aux_trace_frame = ood_trace_frame.aux_frame();
+    let ood_lagrange_kernel_column_frame = ood_trace_frame.lagrange_kernel_frame();
```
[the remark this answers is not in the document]
plafer: Renamed to `ood_lagrange_kernel_frame` (which is what I assume you meant?)

plafer:
Addressed comments, and further optimized constraint evaluation. 

## Timing
For a trace length of $2^{20}$, timings are:

**trace column building:** 84ms
**precompute transition divisors time:** 11ms
**invert transition divisors time:** 34ms
**invert boundary divisors time:** 19ms
**evaluate and merge constraints time:** 387ms

**TOTAL lagrange constraints eval time:** 452ms

# Benchmarks
Benchmarks are also now significantly faster:

- trace length $2^{16}$: [163.01 ms 165.96 ms 168.60 ms]
- trace length $2^{20}$: [3.0080 s 3.0399 s 3.0690 s]

prover/src/constraints/evaluator/lagrange.rs
```diff
@@ -0,0 +1,289 @@
...
+        let divisor_evals_inv = {
+            let divisor_evaluator = TransitionDivisorEvaluator::<E>::new(
+                num_lagrange_transition_constraints,
+                domain.offset(),
+            );
+
+            // The number of divisor evaluations is
+            // `ce_domain_size + ce_domain_size/2 + ce_domain_size/4 + ... + ce_domain_size/(log(ce_domain_size)-1)`,
+            // which is slightly smaller than `ce_domain_size * 2`
+            let mut divisor_evals: Vec<E> = Vec::with_capacity(domain.ce_domain_size() * 2);
```
irakliyk, line 174: Probably won't change things much, but we frequently use `uninit_vector()` function to skip memory initialization when the whole vector will be overwritten.

prover/src/constraints/evaluator/lagrange.rs
```diff
@@ -0,0 +1,289 @@
...
+    ) -> Self
+    where
+        E: FieldElement<BaseField = A::BaseField>,
+    {
+        Self {
+            lagrange_kernel_constraints: LagrangeKernelConstraints::new(
+                air.context(),
+                lagrange_composition_coefficients,
+                &lagrange_kernel_rand_elements,
+            ),
```
irakliyk, line 34: Why not use `air.get_lagrange_kernel_constraints()` here instead of `LagrangeKernelConstraints::new()`?

prover/src/constraints/evaluator/default.rs
```diff
@@ -107,8 +110,28 @@ where
...
+                        lagrange_kernel_combined_evals.len()
+                    );
+                    main_and_aux_evaluations
+                        .into_iter()
+                        .zip(lagrange_kernel_combined_evals)
+                        .map(|(eval_1, eval_2)| eval_1 + eval_2)
+                        .collect()
+                }
+                None => main_and_aux_evaluations,
+            }
```
irakliyk, line 131: This is probably fine for now, but we may be able to optimize this a bit more in the future. Specifically: right now we first compute the entire vector with Lagrange constraint evaluations, and then add the values to the main/aux evaluations. By refactoring this a bit (and passing `main_and_aux_evaluations` into the Lagrange constraint evaluator we could avoid the second loop (i.e., we'd treat `main_and_aux_evaluations` as an accumulator to which Lagrange constraint evaluations get added immediately as they are computed for each row).

Let's add a `TODO` here and also create an issue for this.

air/src/air/lagrange/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+use math::FieldElement;
+pub use transition::LagrangeKernelTransitionConstraints;
+
+use crate::{AirContext, LagrangeConstraintsCompositionCoefficients};
+
+/// Represents the Lagrange kernel transition and boundary constraints.
+pub struct LagrangeKernelConstraints<E: FieldElement> {
+    pub transition: LagrangeKernelTransitionConstraints<E>,
+    pub boundary: LagrangeKernelBoundaryConstraint<E>,
+}
```
irakliyk, line 17: I wonder if the Lagrange column index should be a part of this struct. Would slightly simplify the code in a few places.

verifier/src/evaluator.rs
```diff
@@ -42,13 +46,17 @@ pub fn evaluate_constraints<A: Air, E: FieldElement<BaseField = A::BaseField>>(
...
-        air.evaluate_aux_transition(
-            main_trace_frame,
-            aux_trace_frame,
-            &periodic_values,
-            &aux_rand_elements,
-            &mut t_evaluations2,
-        );
+        // `t_evaluations_2` would be empty here if there is only 1 auxiliary column, and that
+        // column is the Lagrange kernel column
+        if !t_evaluations2.is_empty() {
```
irakliyk, line 51: Should this be an error as a trace with just one Lagrange column is not really useful?

prover/src/constraints/evaluator/lagrange.rs
plafer: `Vec::with_capacity()` doesn't initialize the memory, as discussed in the [Guarantees section of the docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees) (see the diagram). So unless I'm missing something, the following 2 approaches generate the same code:
- construct vector with `Vec::with_capacity()`, and `Vec::push()` elements one by one
- construct vec with `uninit_vector()`, and write elements one by one with `v[i] = ...;`

Since `Vec::with_capacity()` is canonical, and doesn't involve exposing uninitialize memory, I think it should be preferred over `uninit_vector()`. Unless I'm missing something?

air/src/air/lagrange/boundary.rs
```diff
@@ -0,0 +1,60 @@
...
+        Self {
+            assertion_value: Self::assertion_value(lagrange_kernel_rand_elements),
+            composition_coefficient,
+        }
+    }
+
+    /// Returns the evaluation of the boundary constraint at `x`.
+    ///
+    /// `frame` is the evaluation frame of the Lagrange kernel column `c`, starting at `c(x)`
+    pub fn evaluate_at(&self, x: E, frame: &LagrangeKernelEvaluationFrame<E>) -> E {
```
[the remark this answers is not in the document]
plafer: done

prover/src/constraints/evaluator/lagrange.rs
plafer: Changed to `air.get_lagrange_kernel_constraints()`; however since it returns an `Option<>`, we have to use an `expect()`, which arguably is not as nice?

verifier/src/evaluator.rs
plafer: Oops, forgot to remove that one. Fixed

air/src/air/lagrange/mod.rs
plafer: done

prover/src/constraints/evaluator/default.rs
plafer: I implemented it and got the exact same performance (within noise bounds):

```
trace column building: 84ms
precompute trans divisors time: 11ms
invert trans divisors time: 34ms
invert boundary divisors time: 21ms
evaluate and merge constraints time: 404ms

TOTAL lagrange constraints eval time: 472ms
```

Since the first one is more readable and in a functional style, I would suggest we revert that [commit](https://github.com/facebook/winterfell/pull/247/commits/7c74d894a835f3734da38cb2cee02723605cb7d9). What do you think?

prover/src/constraints/evaluator/lagrange.rs
irakliyk: In the testing I did a while back (so, things may have changed since then) using un-initialized vector and then setting values via `v[I] = x` was noticeably faster than doing `Vec::with_capacity() ` and then using `.push()` method. Was not a huge difference, but iirc it was somewhere between 10% and 20%.

plafer: Oh yes you're right, I just benchmarked it, and indeed I using `uninit_vector()` was 40% faster! I will change the code to use it.

plafer: Actually it makes the code much more complex, since

1. we need to compute the exact size of the vector, which is

```rust
let mut divisor_evals = unsafe {
    uninit_vector(
        domain.ce_domain_size() * 2
            - (domain.ce_domain_size()
                / 2_usize.pow(num_lagrange_transition_constraints as u32 - 1)),
    )
};
```

2. the index to which we need to write to is a complex function of `trans_constraint_idx` and `step`, so we'd be better off introducing an `idx` variable which we increment after each write (which makes the code that much harder to read).

I tried it, ran the benchmarks, and it didn't change the performance. So I would just leave it as is.

irakliyk (changes requested):
Thank you! Looks good! I left some more comments inline - but these are mostly nits.

One thing I'm still trying to think through is whether we should change anything about how we handle OOD frame structs. But I think this is the last potential change (and I'm not sure if anything should be changed at all on this front).

@Al-Kindi-0 - could you give this another review?

prover/src/constraints/evaluator/default.rs
irakliyk: I don't have a strong preference here - so, fine with either of the approaches.

prover/src/trace/poly_table.rs
```diff
@@ -64,11 +72,27 @@ impl<E: FieldElement> TracePolyTable<E> {
...
+        let lagrange_kernel_frame = self.lagrange_kernel_column_idx.map(|col_idx| {
+            let lagrange_kernel_col_poly = self.aux_segment_polys[0].get_column(col_idx);
+
+            LagrangeKernelEvaluationFrame::from_lagrange_kernel_column_poly(
+                lagrange_kernel_col_poly,
+                z,
+            )
+        });
+
+        OodFrameTraceStates::new(current_frame, next_frame, lagrange_kernel_frame)
```
irakliyk, line 95: A few more thoughts about OOD frame structs.

It seems to me that we have two structs which contain basically identical data: `OodFrameTraceStates` (in the `air`  crate, but used primarily in the prover) and `TraceOodFrame` (used in the verifier). I wonder if we can unify these. For example, we could:

1. Replace the use of `TraceOodFrame` with `OodFrameTraceStates`. We might need to add some more methods to `OodFrameTraceStates` to do this - but I am not seeing any fundamental reasons why we can't do that.
2. Rename `OodFrameTraceStates` into `TraceOodFrame` (I like this name a little better).
3. Update the filed names as I mentioned in one of the previous comments. The struct can then look like this:

```Rust
pub struct TraceOodFrame <E: FieldElement> {
    current_row: Vec<E>,
    next_row: Vec<E>,
    lagrange_kernel_frame: Option<LagrangeKernelEvaluationFrame<E>>,
}
```

Then, `TracePolyTable::get_ood_frame()` method would return this struct and it would be passed to the prover channel.

One the verifier side, I think `OodFrame::parse()` could return this struct as well. The signature could look something like this:
```Rust
pub fn parse<E: FieldElement>(
    self,
    main_trace_width: usize,
    aux_trace_width: usize,
    num_evaluations: usize,
) -> Result<(TraceOodFrame<E>, Vec<E>), DeserializationError>
```

This should also nicely encapsulate the "interleaving" logic between current and next rows which is currently split between `OodFrame` and `TraceOodFrame`.

Al-Kindi-0 (commented):
Looks great, thank you!
Left some small comments inline.
There is one thing which should conclude this PR and that is updating the portion related to the DEEP composition polynomial. I am not sure if this should go into this PR (as this one is already big enough and the remaining changes are quite substantial). I think a good improvement to discuss that will be useful for the remaining changes is to see if we can do away with using the coefficient form of the trace column polynomials. This will have some performance advantages, I believe, as well as make some of the changes required to accommodate the Lagrange kernel easy.

air/src/air/mod.rs
```diff
@@ -285,6 +292,23 @@ pub trait Air: Send + Sync {
...
+        lagrange_kernel_rand_elements: &[E],
+    ) -> Option<LagrangeKernelConstraints<E>> {
+        self.context().lagrange_kernel_aux_column_idx().map(|col_idx| {
+            LagrangeKernelConstraints::new(
+                self.context(),
+                lagrange_composition_coefficients,
+                lagrange_kernel_rand_elements,
+                col_idx,
+            )
+        })
```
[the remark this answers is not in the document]
plafer: Agreed. I actually just fully removed it, and removed the `assert!()` statement that checks that the length of `lagrange_composition_coefficients` is `log(trace_len)`. I find it cleaner to just have `LagrangeKernelConstraints` assume that the arguments are well-formed (and leaving that concern to other parts of the codebase).

air/src/air/mod.rs
```diff
@@ -501,9 +525,28 @@ pub trait Air: Send + Sync {
             b_coefficients.push(public_coin.draw()?);
         }
 
+        let lagrange = if self.context().has_lagrange_kernel_aux_column() {
+            let mut lagrange_kernel_t_coefficients = Vec::new();
+            if self.context().has_lagrange_kernel_aux_column() {
+                for _ in 0..self.context().trace_len().ilog2() {
+                    lagrange_kernel_t_coefficients.push(public_coin.draw()?);
+                }
+            }
```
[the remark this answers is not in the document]
plafer: Oops, it wasn't; that was just an editing error.

prover/src/constraints/evaluator/lagrange.rs
```diff
@@ -0,0 +1,289 @@
...
+            let mut current_slice_len = domain.ce_domain_size();
+            for i in 1..num_indices {
+                let next_precompute_index = slice_indices_precomputes[i - 1] + current_slice_len;
+                slice_indices_precomputes.push(next_precompute_index);
+
+                current_slice_len /= 2;
+            }
+
+            slice_indices_precomputes
+        };
```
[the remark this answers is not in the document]
plafer: Right. `get_inverse_divisor_eval()` is called for each index of the constraint evaluation domain, and computing this slice index requires 1 or more divisions (unless there's a closed form expression that I'm not aware of).

So without this precomputation, we'd be back to running a division at every index.

air/src/proof/ood_frame.rs
```diff
@@ -149,13 +194,68 @@ impl Deserializable for OodFrame {
...
+// ================================================================================================
+
+/// Stores the trace evaluations at `z` and `gz`, where `z` is a random Field element. If
+/// the Air contains a Lagrange kernel auxiliary column, then that column interpolated polynomial
+/// will be evaluated at `z`, `gz`, `g^2 z`, ... `g^(2^(v-1)) z`, where `v == log(trace_len)`, and
+/// stored in `lagrange_kernel_frame`.
+pub struct OodFrameTraceStates<E: FieldElement> {
+    current_frame: Vec<E>,
+    next_frame: Vec<E>,
+    lagrange_kernel_frame: Option<LagrangeKernelEvaluationFrame<E>>,
```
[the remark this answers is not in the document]
plafer: Agreed with `current_row` and `next_row` (updated). However I prefer `lagrange_kernel_frame`, as we already use "Lagrange kernel frame" elsewhere in the codebase to refer to these "Lagrange kernel values" - so this way it's more consistent IMO.

prover/src/trace/poly_table.rs
plafer: I am all for this. Should we create a new issue though? This suggestion is pretty self-contained, and this PR is already very big.

air/src/air/context.rs
```diff
@@ -100,10 +103,12 @@ impl<B: StarkField> AirContext<B> {
         assert!(num_main_assertions > 0, "at least one assertion must be specified");
 
         if trace_info.is_multi_segment() {
+            // If the only auxiliary column is the Lagrange kernel one, then we don't require any
+            // other boundary/transition constraints
```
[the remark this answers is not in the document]
plafer: removed

prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -174,6 +175,28 @@ where
...
+        let frame = frame.frame_mut();
+        frame.truncate(0);
+
+        let aux_segment = &self.aux_segment_ldes[0];
+
+        frame.push(aux_segment.get(lagrange_kernel_aux_column_idx, lde_step));
+
+        let frame_length = self.trace_info.length().ilog2() as usize + 1;
+        for i in 0..frame_length - 1 {
+            let shift = self.blowup() * 2_u32.pow(i as u32) as usize;
```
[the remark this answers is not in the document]
plafer: Switched to `(1 << i)` - this is arguably more readable due to the removal of both `as _`.

The benchmarks indeed didn't improve though

irakliyk (approved):
All looks good! Thank you! As the next steps, we should do the following:

1. Make a small refactoring mentioned [here](https://github.com/facebook/winterfell/pull/247#discussion_r1537123827) and also remove the incomplete machinery related to supporting multiple auxiliary trace segments.
2. Complete the ability to run the full GKR protocol after the main trace commitment is computed (discussed in [#257](https://github.com/facebook/winterfell/issues/257). This may require refactoring how our prover/verifier channels work.
3. Implement proper handling of DEEP composition for the Lagrange constraints.

PR #251, opened 2024-02-23
Support into iterator

hackaugusto:
Removes the `Sized` bound, it is no longer necessary, since https://github.com/facebook/winterfell/pull/239 .
The change above allows for `impl<T: Serializable> Serializable for &T`.
The change above allows to change `write_many` to be generic and receive a `IntoIterator` without breaking the serialization for `[T; C]`

The removal of the `Sized` bound also makes the trait object save, and allows for type erasure.

### DIFF
--- a/utils/core/src/serde/byte_writer.rs
+++ b/utils/core/src/serde/byte_writer.rs
@@ -100,7 +100,11 @@ pub trait ByteWriter: Sized {
     /// Serializes all `elements` and writes the resulting bytes into `self`.
     ///
     /// This method does not write any metadata (e.g. number of serialized elements) into `self`.
-    fn write_many<S: Serializable>(&mut self, elements: &[S]) {
+    fn write_many<S, T>(&mut self, elements: T)
+    where
+        T: IntoIterator<Item = S>,
+        S: Serializable,
+    {
         for element in elements {
             element.write_into(self);
         }
--- a/utils/core/src/serde/mod.rs
+++ b/utils/core/src/serde/mod.rs
@@ -16,7 +16,7 @@ pub use byte_writer::ByteWriter;
 // ================================================================================================
 
 /// Defines how to serialize `Self` into bytes.
-pub trait Serializable: Sized {
+pub trait Serializable {
     // REQUIRED METHODS
     // --------------------------------------------------------------------------------------------
     /// Serializes `self` into bytes and writes these bytes into the `target`.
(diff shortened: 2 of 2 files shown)


utils/core/src/serde/byte_writer.rs
```diff
@@ -100,7 +100,11 @@ pub trait ByteWriter: Sized {
     /// Serializes all `elements` and writes the resulting bytes into `self`.
     ///
     /// This method does not write any metadata (e.g. number of serialized elements) into `self`.
-    fn write_many<S: Serializable>(&mut self, elements: &[S]) {
+    fn write_many<S, T>(&mut self, elements: T)
+    where
+        T: IntoIterator<Item = S>,
+        S: Serializable,
+    {
```
irakliyk, line 107: Question: is there a reason not to define this as:
```Rust
fn write_many<S, T>(&mut self, elements: T)
where
    T: Iterator<Item = S>,
    S: Serializable,
{
```

hackaugusto: `IntoIterator` covers more use cases, and basically remove the need for the user to do `write_many(vec.iter())`, a `write_many(vec)` will suffice. At the same time `write_many(vec.iter())` is still supported, since an [`Iterator` implements this trait](https://doc.rust-lang.org/core/iter/trait.IntoIterator.html#impl-IntoIterator-for-I)

irakliyk: Ah - so, if we didn't want to pass the ownership of the object to `write_many()`, we could still write `write_many(vec.iter())`, right?

hackaugusto: `vec.iter()` works, also just `&vec`. Here example:

```rust
use std::fmt::Debug;

fn iter<T, I>(vs: T)
where
    T: IntoIterator<Item = I>,
    I: Debug,
{
    for v in vs {
        print!("{:?} ", v);
    }
    print!("\n");
}

fn main() {
    let vs = vec![1, 2, 3, 4];

    iter(&vs);
    iter(&vs);
    iter(vs.iter());
    iter(vs.iter());
}
```

Will print:

```
1 2 3 4
1 2 3 4
1 2 3 4
```

hackaugusto: One important note, `IntoIterator` is a trait that specifies how to convert a type to an `Iterator`. And that can be owning or not. It is not like `into_iter` that by convention is the owning version of the iterator.

hackaugusto:
@irakliyk yes! The motivation for this was the BTreeMap serialization. This also allows that to be serialized with no copying (whereas now one has to copy the map to a vec and then serialize that)

PR #253, opened 2024-02-28
serde: add support for some of the alloc collections

hackaugusto:
This adds a generic implementation of the traits `Serializable` and `Deserializable` for the following types:

- Vec
- BTreeMap
- BTreeSet

Unfortunately I could use `IntoIterator` because that would conflict with the existing implementation for tuples, and the issue can not be solved without [trait specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html)


irakliyk (approved):
Looks good! Thank you!

PR #254, opened 2024-02-29
Remove Griffin hash

### LINKED ISSUE #252: Suggestion: Remove outdated griffin hash implementation
The current Griffin hash implementation was a deviation from an older design of the primitive family, which latest design has been completely broken for high branch count (typically the number of branches at use here), see https://eprint.iacr.org/2024/347.pdf (table 1, p18).

As such, I'd recommend simply removing the primitive from the library out of caution. I can open such PR if needed.

Nashtare:
As stated in https://eprint.iacr.org/2024/347.pdf (cf #252), the Griffin hash implementation should be removed for security concerns.

closes #252


irakliyk (approved):
All looks good! Thank you so much!

PR #255, opened 2024-02-29
Update logging of auxiliary trace building

Overcastan:
This small PR fixes the time calculation of auxiliary trace building.


irakliyk (approved):
Looks good! Thank you!

PR #259, opened 2024-03-13
Change fri remainder max degree to 127 for the example

adambor:
The example outlined in the README gives an error (FRI polynomial remainder degree must be one less than a power of two).

This is caused by using wrong value of 128 for the fri_remainder_max_degree, the fri remainder has to have a number of coefficients that is power of two (therefore polynomial degree has to be one less than that) so value 127 works well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,7 +326,7 @@ pub fn prove_work() -> (BaseElement, StarkProof) {
         0,  // grinding factor
         FieldExtension::None,
         8,   // FRI folding factor
-        128, // FRI max remainder length
+        127, // FRI max remainder length
     );
 
     // Instantiate the prover and generate the proof.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -326,7 +326,7 @@ pub fn prove_work() -> (BaseElement, StarkProof) {
         0,  // grinding factor
         FieldExtension::None,
         8,   // FRI folding factor
-        128, // FRI max remainder length
+        127, // FRI max remainder length
```
irakliyk, line 329: I think the comment is obsolete here as well. Should be "FRI remainder max degree".

adambor: Correct, changed it in the following commit!

irakliyk (approved):
Looks good! Thank you for fixing this!

PR #263, opened 2024-03-18
fix: re-add unintentionally removed re-exported liballoc macros

bitwalker:
With the changes in #262, the re-exports of `vec!` and `format!` macros was unintentionally removed. This is required for backwards compatibility until the deprecated `boxed`, `collections`, and `string` modules are removed.

/cc @irakliyk


bitwalker:
@irakliyk I'm going to open up a PR against `miden-crypto` as well since a similar fix is needed there.

irakliyk (approved):
Looks good! Thank you!

PR #267, opened 2024-03-27
Explicitly only support 1 auxiliary segment

plafer:
Removes the half-support for multiple auxiliary segment. We now explicitly support only one.


air/src/air/trace_info.rs
```diff
@@ -10,23 +10,19 @@ use utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serial
...
+/// one auxiliary segment. Metadata is just a vector of bytes and can store any values up to 64KB in
 /// size.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct TraceInfo {
     main_segment_width: usize,
-    aux_segment_widths: [usize; NUM_AUX_SEGMENTS],
-    aux_segment_rands: [usize; NUM_AUX_SEGMENTS],
-    num_aux_segments: usize,
+    aux_segment_width: usize,
+    aux_segment_rand: usize,
```
irakliyk, line 25: nit: we should probably name this `num_aux_segment_rands`.

prover/src/lib.rs
```diff
@@ -325,18 +324,18 @@ pub trait Prover {
...
-            aux_trace_segments.push(aux_segment);
+            aux_trace_segment = Some(aux_segment);
         }
 
         // make sure the specified trace (including auxiliary segments) is valid against the AIR.
         // This checks validity of both, assertions and state transitions. We do this in debug
         // mode only because this is a very expensive operation.
         #[cfg(debug_assertions)]
-        trace.validate(&air, &aux_trace_segments, &aux_trace_rand_elements);
+        trace.validate(&air, aux_trace_segment.as_ref(), &aux_trace_rand_elements);
```
irakliyk, line 334: Question: does just `&aux_trace_segment` not work here?

prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -207,7 +207,7 @@ where
             positions,
         )];
 
-        // build queries for auxiliary trace segments
+        // build queries for the auxiliary trace segment
         for (i, segment_tree) in self.aux_segment_trees.iter().enumerate() {
             let segment_lde = &self.aux_segment_ldes[i];
             result.push(build_segment_queries(segment_lde, segment_tree, positions));
```
irakliyk, line 213: Should this be just an if statement? Also, `Self.aux_segment_ldes` and `Self.aux_segment_trees` should probably be just singular objects rather than vectors.

prover/src/trace/poly_table.rs
```diff
@@ -102,8 +107,8 @@ impl<E: FieldElement> TracePolyTable<E> {
         self.main_segment_polys.columns()
     }
 
-    /// Returns an iterator over the polynomials of all auxiliary trace segments.
-    pub fn aux_trace_polys(&self) -> MultiColumnIter<E> {
+    /// Returns an iterator over the polynomials of the auxiliary trace segment.
+    pub fn aux_segment_polys(&self) -> MultiColumnIter<E> {
         MultiColumnIter::new(self.aux_segment_polys.as_slice())
     }
```
irakliyk, line 113: We can probably replace `MultiColumnIter` with just `ColumnIter` and maybe delete `MultiColumnIter` altogether.

prover/src/trace/poly_table.rs
```diff
@@ -43,15 +43,16 @@ impl<E: FieldElement> TracePolyTable<E> {
     /// Adds the provided auxiliary segment polynomials to this polynomial table.
     pub fn add_aux_segment(
         &mut self,
-        aux_segment_polys: ColMatrix<E>,
+        aux_segment_poly: ColMatrix<E>,
```
irakliyk, line 46: nit: I think this should still be plural as we can have multiple columns in a single aux segment.

irakliyk (changes requested):
Thank you! Looks good! I left some comments inline - most are pretty minor.

air/src/air/coefficients.rs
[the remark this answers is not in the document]
plafer: Ah yes oops - fixed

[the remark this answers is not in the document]
plafer: Also renamed `TraceInfo::get_aux_segment_rand_elements()` -> `TraceInfo::get_aux_segment_num_rand_elements()`

air/src/air/trace_info.rs
plafer: Done. Also renamed TraceInfo::get_aux_segment_rand_elements() -> TraceInfo::get_aux_segment_num_rand_elements()

prover/src/lib.rs
plafer: No, since `validate()` asks for `Option<&...>`, and we have `Option<...>` (which is what [`as_ref()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref) does).

Changing `validate()` to take `&Option<...>` doesn't improve on the current state either, because every time you match to get the item inside, you'd need to call `as_ref()` (otherwise you take ownership).

We could make `validate()` take ownership of `aux_trace_segment` too (i.e. signature would be `Option<ColMatrix<E>>`), since we drop it right after anyway?

prover/src/trace/trace_lde/default/mod.rs
plafer: Ah yes, I missed those. Fixed. I also renamed `TraceLde::add_aux_segment()` -> `TraceLde::set_aux_segment()`

prover/src/trace/poly_table.rs
plafer: fixed

prover/src/constraints/evaluator/default.rs
```diff
@@ -237,7 +237,7 @@ where
     ) {
         // initialize buffers to hold trace values and evaluation results at each step
         let mut main_frame = EvaluationFrame::new(trace.trace_info().main_trace_width());
-        let mut aux_frame = EvaluationFrame::new(trace.trace_info().aux_trace_width());
+        let mut aux_frame = EvaluationFrame::new(trace.trace_info().aux_segment_width());
```
irakliyk, line 240: Looking at things like this, I'm wondering if we should use `aux_trace` everywhere rather than `aux_segment`. This would be more consistent with how we name main trace segment. Basically, we have `main_trace` and `aux_trace`. And so things like `aux_segment_commitment` become `aux_trace_commitment` etc.

prover/src/trace/poly_table.rs
```diff
@@ -98,13 +100,16 @@ impl<E: FieldElement> TracePolyTable<E> {
...
-    pub fn main_trace_polys(&self) -> ColumnIter<E::BaseField> {
+    pub fn main_trace_polys(&self) -> impl Iterator<Item = &[E::BaseField]> {
         self.main_segment_polys.columns()
     }
 
-    /// Returns an iterator over the polynomials of all auxiliary trace segments.
-    pub fn aux_trace_polys(&self) -> MultiColumnIter<E> {
-        MultiColumnIter::new(self.aux_segment_polys.as_slice())
+    /// Returns an iterator over the polynomials of the auxiliary trace segment.
+    pub fn aux_segment_polys(&self) -> impl Iterator<Item = &[E]> {
```
irakliyk, line 108: Similar to the last comment: `aux_trace_polys` would be more consistent with `main_trace_polys`.

prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -118,7 +118,7 @@ where
     /// This function will panic if any of the following are true:
     /// - the number of rows in the provided `aux_trace` does not match the main trace.
     /// - this segment would exceed the number of segments specified by the trace layout.
-    fn add_aux_segment(
+    fn set_aux_segment(
```
irakliyk, line 121: I think the comment on line 120 needs to be updated.

irakliyk (approved):
Looks good! Thank you! I left a coupe of minor comments inline. After these are addressed, we can merge.

There is also a bigger comment about renaming. Currently we use `main_segment` and `main_trace` as well as `aux_segment` and `aux_trace` somewhat interchangeably. I think we should just use `main_trace` and `aux_trace` everywhere. But this renaming should probably be done in another PR.

prover/src/trace/mod.rs
```diff
@@ -51,19 +51,13 @@ pub trait Trace: Sized {
     /// Returns a reference to a [Matrix] describing the main segment of this trace.
     fn main_segment(&self) -> &ColMatrix<Self::BaseField>;
 
-    /// Builds and returns the next auxiliary trace segment. If there are no more segments to build
+    /// Builds and returns the auxiliary trace segment. If there are no more segments to build
     /// (i.e., the trace is complete), None is returned.
```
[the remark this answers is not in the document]
plafer: fixed

prover/src/trace/poly_table.rs
plafer: fixed

prover/src/trace/trace_lde/default/mod.rs
plafer: done

irakliyk (approved):
All looks good! Thank you!

PR #270, opened 2024-04-11
CI: check msrv

hackaugusto:
- [`actions-rs/toolchain`](https://github.com/actions-rs/toolchain) is no longer maintained and it is now an archived project
- there is no official action for rust, but now the GH action image include rustup natively, so the changes bellow uses rustup directly (this is the strategy from projects like [`rust-analyzer`](https://github.com/rust-lang/rust-analyzer/blob/99718d0c8bc5aadd993acdcabc1778fc7b5cc572/.github/workflows/ci.yaml#L63-L67))
- the main change here is adding the `cargo-hack` tool. Using that tool mainly because it is recommended in the [cargo book](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-rust-version)
    - This is basically a wrapper around `cargo check`, which checks every combination of features. It is fast enough that it should be done before the tests.
    - It runs using the compiler with the `rust-version`, which checks the setting is correct (it doesn't guarantee it is the lowest MSRV, but it does check that the setting is a valid MSRV)
- This fixes a typo in the CI trigger condition, it was `repoened`. Now it just uses the default which are [ref](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,114 +1,64 @@
 name: CI
 on:
+  workflow_dispatch:
+  pull_request:
   push:
     branches:
       - main
-  pull_request:
-    types: [opened, repoened, synchronize]
 
 jobs:
-  check:
-    name: Check with ${{matrix.features}}
+  light-checks:
+    name: Ligth checks
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        toolchain: [stable]
-        os: [ubuntu]
-        features: [--all-features, --no-default-features]
     steps:
       - uses: actions/checkout@main
-      - name: Install rust
-        uses: actions-rs/toolchain@v1
-        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -1,114 +1,64 @@
 name: CI
 on:
+  workflow_dispatch:
```
irakliyk, line 3: What does `workflow_dispatch` enable here?

irakliyk (approved):
Looks great! Thank you!

.github/workflows/ci.yml
hackaugusto: It [enable manual runs](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow)

PR #271, opened 2024-04-19
Generalize auxiliary trace building

### LINKED ISSUE #257: Generalize auxiliary trace building logic
With #240 almost done via #247 , it makes sense to discuss how we can make use of these recent updates in the specific context of LogUp GKR.

# Current
To recap, at a high level, the current workflow between the (STARK) prover and verifier can be described using the following simplified diagram
original_aux_trace

In the above
1.  $\mathsf{f_i}$ is a polynomial encoding the $i$-th main trace column and $\mathsf{commit(f_i)}$ is its commitment.
2. $\alpha := \left(\alpha_i\right)_i$ is a vector of random values used in reducing global constraints to local ones e.g., multi-set checks.
3.  $\mathsf{f_i^{'}}$ is a polynomial encoding the $i$-th auxiliary trace column and $\mathsf{commit(f_i^{'})}$ is its commitment.
4. $\beta := \left(\beta_i\right)_i$ is a vector of random values used in combining constraints in order to construct the constraint composition polynomial.
5.  $\mathsf{g_i}$ is a polynomial encoding the $i$-th segment polynomial of the constraint composition polynomial and $\mathsf{commit(g_i)}$ is its commitment.
6. $\mathsf{z}$ is the OOD evaluation point.
7. $\mathsf{v_i}$ are the evaluations of all the previous oracles (i.e., committed polynomials) at $\mathsf{z}$ and potentially their shifts.
8. $\gamma := \left(\gamma_i\right)_i$ is a vector of random values used in constructing the DEEP polynomial i.e., the input to FRI.
9. $\mathsf{h}$ is a polynomial encoding the DEEP polynomial and $\mathsf{commit(h)}$ is its commitment.

Note that FRI is invoked at the end in order to show low degree-ness as well as prove the openings. FRI is itself an IOP and thus is composed of multiple rounds of the prover sending commitments to polynomials and the verifier replying with randomness. The verifier concludes by doing a few checks on the messages it received from the prover and the randomness it sent and either outputs an Accept or Reject.

# Updated

We would like now to move to the following picture

proposal_aux_trace

The main difference now is that we allow the auxiliary trace building stage to be potentially more complex than just the verifier sending randomness and the prover answering with commitments to the auxiliary trace columns.
More specifically, in the case of GKR-LogUp, we could have the following simplified flow

gkr_builder_aux_trace

1. The verifier sends $\alpha := \left(\alpha_i\right)_i$ i.e., LogUp randomness.
2. Both the prover and verifier compute $C\left\(\alpha\right)$.
3. GKR is used for the statement $$\sum_{x\in\\{0,1\\}^{\nu}} \frac{p(x)}{q(x)} = C\left\(\alpha\right).$$ Note that we abuse the notation with regards to MLEs . Note also that $p$ and $q$ depend on $\alpha$ implicitly.
4. After a certain number of rounds making up the GKR IOP, the verifier is left with a GKR proof $\mathsf{\pi_{gkr}}$ and the prover is left with a vector of random values $\mathsf{r} := \left(\mathsf{r_i}\right)_i$.
5. The prover sends $\mathsf{y_i}$ it claims are the evaluations of the MLEs of $\mathsf{f_i}$.
6. The verifier sends randomness $\lambda := \left(\lambda_i\right)_i$ in order to batch the (claimed) evaluations into one evaluation $\mathsf{y} := \sum_i \mathsf{y_i}\cdot\lambda_i$.
7. The prover sends $\mathsf{commit(l)}$ and $\mathsf{commit(s)}$ where:  
           1. $\mathsf{l}$ is the Lagrange kernel at $\mathsf{r}$.
           2. $\mathsf{s}$ is the univariate sum-check column i.e., a column such that $\mathsf{s[row]} = \mathsf{s[row-1]} + \sum_i \lambda_i\mathsf{f_i(row)}\cdot\mathsf{l(row)}$[^1].
8. Both the prover and verifier update[^2] their AIR instance using $\mathsf{r}$, $\lambda$ and $\mathsf{y}$. This includes a boundary constraint against the last row of column $\mathsf{s}$ forcing it to be equal to $\mathsf{y}$ as well as the transition constraint in point 7.2 above. There are also the Lagrange kernel constraints that are defined using $\mathsf{r}$.

To summarize, with the new changes, the auxiliary trace building stage would culminate in a proof, an auxiliary segment composed of two columns and an updated AIR instance. 
As can be seen, there is a strong similarity between the above and the FRI stage i.e., both are IOPs with a non-constant number of rounds.
The task now is to discuss possible ways we can accommodate the above with the least amount of abstraction necessary i.e., offloading most of the work which happens at the auxiliary trace building stage to the "app" instead of the proof system.

[^1]: The sum might also include shifts of $\mathsf{f_i}$ as well as periodic columns. These are omitted in the analysis above for simplicity.
[^2]: We can also think of some of the constraint as being uninitialized and they get initialized using the appropriate randomness.

plafer:
Closes #257


prover/src/lib.rs
```diff
@@ -192,44 +194,73 @@ pub trait Prover {
...
+        let air = Self::Air::new(trace.info().clone(), pub_inputs, self.options().clone());
+
+        // create a channel which is used to simulate interaction between the prover and the
+        // verifier; the channel will be used to commit to values and to draw randomness that
+        // should come from the verifier.
+        let mut channel =
+            ProverChannel::<Self::Air, Self::BaseField, Self::HashFn, Self::RandomCoin>::new(
+                &air,
+                pub_inputs_elements,
+            );
```
plafer, line 226: We can't pull this out into a new function, since `channel` borrows `air`, and Rust won't let us return a value that is borrowed

plafer:
Note that we no longer use `TraceInfo::aux_segment_width` (it is instead passed to `AuxTraceBuilder` when needed) and `TraceInfo::num_aux_segment_rands`; we should probably remove them in this PR? It might require a bigger refactor to `TraceInfo`, since `TraceInfo` wouldn't really track information about the auxiliary trace anymore.

prover/src/lib.rs
```diff
@@ -192,44 +194,73 @@ pub trait Prover {
...
-                if !<CubeExtension<Self::BaseField>>::is_supported() {
-                    return Err(ProverError::UnsupportedFieldExtension(3));
-                }
-                self.generate_proof::<CubeExtension<Self::BaseField>>(trace)
-            }
-        }
-    }
+        // FIXME: the `self.options().field_extension` is now irrelevant and could be inconsistent
+        // with `E`.
+        assert!(!trace.info().is_multi_segment());
```
plafer, line 206: A big change in this PR is that the field extension is now known at compile-time. If the trace has an auxiliary segment, the caller must call `prove_with_aux_trace()`, which takes a generic `E`.

The `ProofOptions::field_extension` ultimately is limiting and [results in this "hack"](https://github.com/facebook/winterfell/blob/09525751727a283dfbf5e569a09909b82380e059/prover/src/lib.rs#L206-L219) (removed in this PR). Since completely removing `FieldExtension` is a larger change, we should do it in a subsequent cleanup PR.

plafer:
@irakliyk @Al-Kindi-0 ready for review :)

air/src/air/mod.rs
```diff
@@ -283,11 +286,20 @@ pub trait Air: Send + Sync {
...
     }
 
+    /// Returns the random elements used in constructing the Lagrange kernel column.
+    #[allow(unused_variables)]
+    fn get_lagrange_rand_elements<'a, E: Send + Sync>(
+        &self,
+        rand_eles: &'a Self::AuxRandElements<E>,
+    ) -> &'a [E] {
+        unimplemented!("getting the Lagrange random elements has not been implemented");
+    }
```
irakliyk, line 301: I'm assuming the number of random elements is pretty small. If so, we could probably change the return type to `Vec<E>` and not deal with lifetimes here.

Also minor nit: I don't think we abbreviate `elements` as `eles` in other places. We can probably just use `rand_elements` or maybe even `aux_rand_elements` here.

prover/src/channel.rs
```diff
@@ -68,6 +68,14 @@ where
         }
     }
 
+    // GETTER METHODS
+    // --------------------------------------------------------------------------------------------
+
+    /// Returns the inner public coin
+    pub fn public_coin(&mut self) -> &mut R {
+        &mut self.public_coin
+    }
```
irakliyk, line 77: minor nit: I would probably move this into the `PUBLIC COIN METHODS` section below.

prover/src/trace/mod.rs
```diff
@@ -51,17 +54,6 @@ pub trait Trace: Sized {
...
-    /// Builds and returns the auxiliary trace segment. If the trace does not require an auxiliary
-    /// segment, None is returned.
-    ///
-    /// The `rand_elements` slice contains the random elements to use to build the aux segment. If a
-    /// Lagrange kernel column is present, the `lagrange_kernel_rand_elements` should be used.
-    fn build_aux_segment<E: FieldElement<BaseField = Self::BaseField>>(
-        &mut self,
-        rand_elements: &[E],
-        lagrange_kernel_rand_elements: Option<&[E]>,
-    ) -> Option<ColMatrix<E>>;
```
irakliyk, line 63: What is the benefit of replacing this function with `AuxTraceBuilder` trait?

verifier/src/aux_verifier.rs
```diff
@@ -0,0 +1,53 @@
+use alloc::{string::ToString, vec::Vec};
+use crypto::{ElementHasher, RandomCoin, RandomCoinError};
+use math::FieldElement;
+
+/// A trait for generating the random elements required for constructing the auxiliary trace.
+pub trait AuxRandElementsGenerator {
```
irakliyk, line 6: Why is this trait defined in the verifier? Shouldn't both prover and verifier use the same logic for computing random elements?

prover/src/lib.rs
irakliyk: Sometimes we do want to pick field extension dynamically (e.g., degree 2 extension of the `f64` field can be used for ~100 bit security, while extension 3 is needed for 128 bit security).

How would we achieve this dynamic selection with this structure?

irakliyk (commented):
Thank you! This is not a full review (and I'm still going through and thinking over the changes), but I left some small nits inline and also added some questions which will help me with understanding of the general approach.

air/src/air/mod.rs
plafer: fixed

prover/src/channel.rs
plafer: fixed

prover/src/lib.rs
plafer: Could you give an example of a use case where this needs to be dynamic?

Could we not solve this by including type aliases for each field, e.g.

```rs
// f64 field

pub type ExtensionField100BitSecurity = QuadExtension<BaseElement>;
pub type ExtensionField128BitSecurity = CubeExtension<BaseElement>;
```

prover/src/trace/mod.rs
plafer: The main reason was as to not pollute the `Trace` trait, since now building the auxiliary trace requires 3 associated types: `AuxRandElements<E>`, `AuxParams`, and `AuxProof` that are only used with `build_aux_trace()`.

We could put it back in `Trace`, but then building a simple trace would require a lot of unnecessary dummy types:

```rs
impl Trace for SimpleTrace {
  type AuxRandElements<E> = PhantomData<E>;
  type AuxParams = ();
  type AuxProof = ();

  // ...
}
```

which could be okay too?

verifier/src/aux_verifier.rs
plafer: Right, I ended up reimplementing this logic in the prover by hand. Changed it so that `AuxTraceBuilder` now uses a `AuxRandElementsGenerator`

plafer: Correction: reverted those changes. I had forgotten why I had made a verifier-only trait, and @Al-Kindi-0 reminded me: this one needs to potentially verify the auxiliary proof as well.

Renamed `AuxTraceVerifier`, and now optionally verifies an `AuxProof`.

prover/src/lib.rs
```diff
@@ -192,44 +194,72 @@ pub trait Prover {
...
+    ///
+    /// The returned [StarkProof] attests that the specified `trace` is a valid execution trace of
+    /// the computation described by [Self::Air](Prover::Air) and generated using some set of
+    /// secret and public inputs. Public inputs must match the value returned from
+    /// [Self::get_pub_inputs()](Prover::get_pub_inputs) for the provided trace.
+    fn prove_with_aux_trace<E, ATB>(
+        &self,
+        trace: Self::Trace,
+        mut aux_trace_builder: ATB,
+    ) -> Result<(StarkProof, Option<AuxProof<ATB>>), ProverError>
```
irakliyk, line 256: A couple of thoughts here:

1. I think we should try to combine the `prove()` and `prove_with_aux_trace()` methods so that we have a single `prove()` method.
2. I think we should return a single proof object from this method (rather than a main proof and an optional auxiliary proof).

For the second point, we could make sure that `AuxProof` is serializable and deserializable and then just make it a part of the main proof (we could rename `StarkProof` into just `Proof` to make the naming consistent with the content).

For the first point, I'm not sure yet what the best solution would be. One potential approach is to provide `AuxTraceBuilder` as an optional parameter. The method then would look something like this:
```Rust
fn prove<E, ATB>(
    &self,
    trace: Self::Trace,
    aux_trace_builder: ATB,
) -> Result<Proof, ProverError>
where
    ATB: AuxTraceBuilder<AuxRandElements<E> = Self::AuxRandElements<E>>,
    Self::Air: Air<AuxRandElements<E> = AuxRandElements<ATB, E>>,
```
I think this could work, but there are two downsides to this:
1. We need to make `prove()` generic over `E` and thus, when deserializing `Proof` object, we need to know which extension filed was used to generate the proof. This is not a big deal (we handle hash function options like this already) - but it does add a bit more work.
2. I don't love making the function generic over `AxTraceBuilder` (and to some extent needing to pass a new `aux_trace_builder` on every call to `prove()`). A potential way around this is to make `AuxTraceBuilder` an associated type and then add a constructor method to the `Prove` trait to build it similar to how we handle `TraceLde` construction. But I understand that there may be challenges with this as well.

The other potential option is to get rid of `AuxTraceBuilder` trait and go back to the `Trace::build_aux_segment()` function. As you've pointed out, we'd need to add two extra associated types to the `Trace` trait:

```Rust
pub trait Trace: Sized {
    type BaseField: StarkField;
    type AuxRandElements<E: FieldElement<BaseField = Self::BaseField>>;
    type AuxProof: Serializable + Deserializable
```
The complication here could be related to enforcing that `Air::AuxRandElements` and `Trace::AuxRandElements` are the same type.

prover/src/lib.rs
```diff
@@ -125,6 +124,9 @@ pub trait Prover {
     /// Base field for the computation described by this prover.
     type BaseField: StarkField + ExtensibleField<2> + ExtensibleField<3>;
 
+    /// A type defining the random elements used in constructing the auxiliary trace segment.
+    type AuxRandElements<E: Send + Sync>;
```
irakliyk, line 128: Do we actually need this type? I think we should be able to refer to it via the `Air` trait like so: `<<Self as Prover>::Air as Air>::AuxRandElements<E>`.

plafer: I think I added it at some point to help write trait bounds, but it is no longer needed. Removed.

prover/src/lib.rs
plafer: - Merged both `prove_()` functions into a single non-generic `prove()` (as we previously had)
- `StarkProof` now contains a serialized `AuxProof`

The one quirk that I was not able to remove is that provers that prove a trace with no auxiliary segment will need to set `type AuxTraceBuilder<E> = EmptyAuxTraceBuilder`, where `EmptyAuxTraceBuilder` is a dummy type that implements the `AuxTraceBuilder` trait. This is ultimately a consequence of `prove()` not being generic:
- We can't write the `Self::AuxTraceBuilder<E>: AuxTraceBuilder<AuxRandElements<E> = AuxRandElementsProver<Self, E>, AuxProof = Self::AuxProof>` trait bound on `prove()` (since there is no `E` generic type)
- We also can't write it on `generate_proof<E>()`, since then `prove()` can't call it anymore (because `prove()` doesn't have the constraint)
- So we're forced to put that trait bound directly on the `Prover`'s associated type definition. Which means that implementors can't use `type AuxTraceBuilder = ()` anymore, since `()` doesn't implemement the `AuxTraceBuilder` trait.

Also, a consequence of putting the serialized `AuxProof` in `StarkProof` (now renamed `Proof`) is that the verifier needs to know the type that it will deserialize in. Hence, I wasn't able to avoid having 2 `verify()` methods (one that provides an `AuxTraceVerifier` which defines the `AuxProof` type, and on that doesn't).

prover/src/lib.rs
```diff
@@ -191,14 +199,28 @@ pub trait Prover {
     // PROVIDED METHODS
     // --------------------------------------------------------------------------------------------
 
+    /// Builds the auxiliary trace.
+    #[allow(unused_variables)]
+    fn build_aux_trace<E, Hasher>(
+        &self,
+        main_trace: &ColMatrix<E::BaseField>,
+        transcript: &mut impl RandomCoin<BaseField = E::BaseField, Hasher = Hasher>,
+    ) -> AuxTraceWithMetadata<E, AuxRandElementsProver<Self, E>, AuxProofProver<Self>>
```
irakliyk, line 208: Can we pass `main_trace` into this as `Self::Trace`?

Also, is there a reason why we pass `RandomCoin` as `impl` rather than as `Self::RandomCoin`?

Lastly, I think we should update the comment here to describe in more detail what this method is supposed to do in various cases.

prover/src/lib.rs
```diff
@@ -191,14 +199,28 @@ pub trait Prover {
...
+    where
+        E: FieldElement,
+        Hasher: ElementHasher<BaseField = E::BaseField>,
+    {
+        unimplemented!("`Prover::build_aux_trace` needs to be implemented when the trace has an auxiliary segment.")
+    }
+
     /// Returns a STARK proof attesting to a correct execution of a computation defined by the
-    /// provided trace.
+    /// provided trace. The trace must not contain an auxiliary trace segment.
```
irakliyk, line 217: I think the comment about auxiliary trace is no longer relevant. Also, maybe worth mentioning that `Proof` can now contain additional stuff.

prover/src/lib.rs
```diff
@@ -219,14 +241,11 @@ pub trait Prover {
         }
     }
 
-    // HELPER METHODS
-    // --------------------------------------------------------------------------------------------
```
irakliyk, line 223: nit: I would probably keep this separator here.

examples/src/fibonacci/fib2/air.rs
```diff
@@ -18,8 +18,10 @@ pub struct FibAir {
 }
 
 impl Air for FibAir {
+    type AuxRandElements<E: Send + Sync> = ();
     type BaseField = BaseElement;
     type PublicInputs = BaseElement;
+    type AuxProof = ();
```
irakliyk, line 24: minor nit: I'd probably order these as follows:
```Rust
type BaseField = BaseElement;
type PublicInputs = BaseElement;
type AuxRandElements<E: Send + Sync> = ();
type AuxProof = ();
```
(relevant for other example files as well)

examples/src/fibonacci/fib2/prover.rs
```diff
@@ -84,7 +84,7 @@ where
     fn new_evaluator<'a, E: FieldElement<BaseField = Self::BaseField>>(
         &self,
         air: &'a Self::Air,
-        aux_rand_elements: AuxTraceRandElements<E>,
+        aux_rand_elements: Option<AuxRandElementsProver<Self, E>>,
```
irakliyk, line 87: nit: could this be: `aux_rand_elements: Option<()>` since we set them to `()` in `Air`?

air/src/air/trace_info.rs
```diff
@@ -200,11 +200,6 @@ impl TraceInfo {
     pub fn get_aux_segment_width(&self) -> usize {
         self.aux_segment_width
     }
-
-    /// Returns the number of random elements required by the auxiliary trace segment.
-    pub fn get_num_aux_segment_rand_elements(&self) -> usize {
-        self.num_aux_segment_rands
-    }
```
irakliyk, line 207: Do we not need this any more? Seems a bit odd that we are removing the accessor but keeping the field (i.e., `self.num_aux_segment_rands).

air/src/air/mod.rs
```diff
@@ -188,6 +189,12 @@ pub trait Air: Send + Sync {
     /// This could be any type as long as it can be serialized into a sequence of field elements.
     type PublicInputs: ToElements<Self::BaseField>;
 
+    /// A type defining the random elements used in constructing the auxiliary trace segment.
+    type AuxRandElements<E: Send + Sync>: Send + Sync;
```
irakliyk, line 193: I would probably also impose a trait bound on `E: FieldElement<BaseField = Self::BaseField>`.

prover/src/lib.rs
```diff
@@ -487,6 +478,33 @@ pub trait Prover {
         Ok(proof)
     }
 
+    // HELPER METHODS
+    // --------------------------------------------------------------------------------------------
+
+    fn build_main_trace_and_commit<E>(
```
irakliyk, line 484: This method will appear in the public interface - so, we should document it (similar to the `build_constraint_commitment()` method below).

prover/src/lib.rs
```diff
@@ -323,19 +306,26 @@ pub trait Prover {
...
+
+            Some(aux_trace_with_metadata)
+        } else {
+            None
+        };
+
+        let (aux_trace, aux_rand_elements, aux_proof) = match aux_trace_with_metadata {
+            Some(atm) => (Some(atm.aux_trace), Some(atm.aux_rand_eles), atm.aux_proof),
+            None => (None, None, None),
+        };
```
irakliyk, line 318: nit: should this be merged with the previous statement block?

prover/src/lib.rs
```diff
@@ -99,6 +101,12 @@ pub mod tests;
 // this segment width seems to give the best performance for small fields (i.e., 64 bits)
 const DEFAULT_SEGMENT_WIDTH: usize = 8;
 
+/// Accesses the `AuxRandElements` type in a [`Prover`].
+pub type AuxRandElementsProver<P, E> = <<P as Prover>::Air as Air>::AuxRandElements<E>;
+
+/// Accesses the `AuxProof` type in a [`Prover`].
+pub type AuxProofProver<P> = <<P as Prover>::Air as Air>::AuxProof;
```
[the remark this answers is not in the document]
plafer: done

prover/src/lib.rs
plafer: This was an artifact of copy/pasting from the former `AuxTraceBuilder`. Now fixed.

I also documented what an auxiliary proof is in `AuxProofWithMetadata`, and `build_aux_trace()` now refers to it "for more information".

prover/src/lib.rs
plafer: Done. Docstring now briefly mentions the aux proof, refers to `Proof` for more information

prover/src/lib.rs
plafer: done

examples/src/fibonacci/fib2/air.rs
plafer: Fixed them all

examples/src/fibonacci/fib2/prover.rs
plafer: Fixed this one and all other relevant ones

air/src/air/trace_info.rs
plafer: Oops, I forgot to remove the field. Fixed.

We don't need it anymore because the `TraceInfo` doesn't know anymore how many random elements will be generated (e.g. in the case of GKR, that number is dynamic)

air/src/air/mod.rs
plafer: Fixed, but I think we should move away from adding trait bounds in type definitions, as [discussed on Discord](https://discord.com/channels/893151281848406016/1166735622455689268/1167139113024098364) (pointing to this [StackOverflow thread](https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl/66369912#66369912))

prover/src/lib.rs
plafer: I had added that helper function when we had 2 `prove()` methods, but it's no longer needed, so I removed it

prover/src/lib.rs
plafer: Left it separate as a consequence on `trace.validate()` now taking `aux_trace_with_metadata` ([comment](https://github.com/facebook/winterfell/pull/271#discussion_r1586754298))

air/src/proof/mod.rs
```diff
@@ -37,22 +37,26 @@ mod tests;
...
-/// A proof can be serialized into a sequence of bytes using [to_bytes()](StarkProof::to_bytes)
-/// function, and deserialized from a sequence of bytes using [from_bytes()](StarkProof::from_bytes)
-/// function.
+/// A proof can be serialized into a sequence of bytes using [to_bytes()](Proof::to_bytes) function,
+/// and deserialized from a sequence of bytes using [from_bytes()](Proof::from_bytes) function.
 ///
-/// To estimate soundness of a proof (in bits), [security_level()](StarkProof::security_level)
-/// function can be used.
+/// To estimate soundness of a proof (in bits), [security_level()](Proof::security_level) function
+/// can be used.
```
Al-Kindi-0, line 57: We should add an issue or a TODO to update the security estimator so that it takes into account the security of the auxiliary proof as well.

plafer: Created https://github.com/facebook/winterfell/issues/273

air/src/air/trace_info.rs
plafer: Note: I also removed `MAX_RAND_SEGMENT_ELEMENTS`, which I believe was only a `TraceInfo::to_elements()` limitation (it had to fit in a `u8`)

verifier/src/aux_verifier.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// verifies the auxiliary proof.
+    fn verify_aux_trace<E, Hasher>(
+        &self,
+        aux_proof: Option<Self::AuxProof>,
+        transcript: &mut impl RandomCoin<BaseField = E::BaseField, Hasher = Hasher>,
+    ) -> Result<Self::AuxRandElements<E>, Self::Error>
+    where
+        E: FieldElement,
+        Hasher: ElementHasher<BaseField = E::BaseField>;
+}
```
irakliyk, line 22: I would rename this trait into `AuxVerifier` and the `verify_aux_trace()` method into just `verify()`.

But also, it feels like we are trying to do too much with this trait. If I'm understanding things correctly, we have 3 types of auxiliary random values:

1. Random values needed for constructing user-defined auxiliary columns. The number of these random values was previously specified by `num_aux_segment_rands` (and which I think we might need to bring back).
2. Random values needed for constructing Lagrange kernel columns. The number of these values depends on the trace length.
3. Random values needed for constructing/verifying auxiliary proof.

Currently, this method (and its prover equivalent `build_aux_trace()`) is responsible for generating all 3 types of random values. But I'm wondering if that's the right approach.

An alternative could be to let the library generate the first two types of random values (since the data for these is known via the AIR definition and trace length) and have this method worry only about the 3rd type. Concretely, this could look like so:

`AuxRandValues` would become a struct which could look like:
```Rust
pub struct AuxRandElements<E> {
    user_defined: Vec<E>, // probably a better name is needed here
    lagrange: Vec<E>,
}
```

In the `Air` trait, we'd have:
```Rust
fn get_aux_rand_elements<E, R>(
     &self,
     public_coin: &mut R,
 ) -> Result<AuxRandElements, RandomCoinError>
```
This would generate random values both for user-defined columns and Lagrange kernel columns as needed. This would also eliminate the need for the `Air::get_lagrange_rand_elements()` method.

On the prover side, we'd change `build_aux_trace()` method to look as follows:
```Rust
fn build_aux_trace<E>(
    &self,
    main_trace: &Self::Trace,
    aux_rand_elements: &AuxRandElements<E>,
    public_coin: &mut Self::RandomCoin,
) -> (ColMatrix<E>, ProverAuxProof<Self>)
```

Or, if `AuxProof` needs to be generated before the randomness for Lagrange kernel columns, we could split this into two methods:
```Rust
/// Builds AuxProof object drawing randomness from the public coin as needed.
fn generate_aux_proof<E>(
    &self,
    main_trace: &Self::Trace,
    public_coin: &mut Self::RandomCoin,
) -> ProverAuxProof<Self>

/// Builds auxiliary trace columns.
fn build_aux_trace<E>(
    &self,
    main_trace: &Self::Trace,
    aux_rand_elements: &AuxRandElements<E>,
) -> ColMatrix<E>
```

One the verifier side, `AuxVerifier::verify` would look something like this:
```Rust
fn verify_aux_trace<E, Hasher>(
   &self,
   aux_proof: Self::AuxProof,
   public_coin: &mut impl RandomCoin<BaseField = E::BaseField, Hasher = Hasher>,
) -> Result<(), Self::Error>
```
And would be responsible only for verifying the `AuxProof`.

It is possible I've missed some crucial detail somewhere and this doesn't work for some reason. But if not, it feels to me like a cleaner approach.

irakliyk (changes requested):
Thank you! Looks good! I left some more comments inline - the main one is about potentially separating aux proof and aux trace handling. But I feel like this is the last major thing to align on.

air/src/air/mod.rs
irakliyk: Agreed - but we should probably do it more comprehensively in a follow-up PR.

air/src/air/trace_info.rs
irakliyk: I left a comment in the review below about this.

examples/src/rescue_raps/mod.rs
```diff
@@ -127,30 +130,85 @@ where
...
+            FieldExtension::Cubic => {
+                winterfell::verify_with_aux_trace::<
+                    CubeExtension<BaseElement>,
+                    RescueRapsAir,
+                    _,
+                    H,
+                    DefaultRandomCoin<H>,
+                >(proof, aux_verifier, pub_inputs, &acceptable_options)
+            }
+        }
```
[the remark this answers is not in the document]
plafer: This is `verify_with_aux_trace()` though, so we need to decide what is the type of `E`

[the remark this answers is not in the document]
irakliyk: Ah - this is because `AuxTraceVerifier` needs to know `E`, right?

examples/src/rescue_raps/prover.rs
```diff
@@ -132,9 +136,67 @@ where
...
+    {
+        let main_trace = main_trace.main_segment();
+        let rand_elements = {
+            let mut rand_elements = Vec::with_capacity(self.aux_segment_width);
+            for _ in 0..self.aux_segment_width {
+                rand_elements.push(transcript.draw().unwrap());
+            }
+
+            rand_elements
+        };
```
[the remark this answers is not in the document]
plafer: Oops, that's just a naming error on my end. It should just be `num_rand_elements`

air/src/air/trace_info.rs
```diff
@@ -80,7 +77,6 @@ impl TraceInfo {
     pub fn new_multi_segment(
         main_segment_width: usize,
         aux_segment_width: usize,
-        num_aux_segment_rands: usize,
         trace_length: usize,
         trace_meta: Vec<u8>,
     ) -> Self {
```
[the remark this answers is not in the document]
plafer: The struct that implements `Prover` would be responsible for that. As noted in [this comment](https://github.com/facebook/winterfell/pull/271#discussion_r1589000554), the `RescueRapsProver` is an example of that: it has a variable `num_rand_elements` (renamed from `aux_segment_width`).

However the `Air` no longer knows about it, so the `AuxTraceVerifier` also needs to "hardcode" the same constant... Will look into it more

verifier/src/aux_verifier.rs
plafer: I agree that `build/verify_aux_trace()` needs to be improved, but came to a different conclusion.

> But also, it feels like we are trying to do too much with this trait. If I'm understanding things correctly, we have 3 types of auxiliary random values:

This isn't quite right. At the lowest level, the `Air` can't generate Lagrange kernel random values separately from building the GKR proof, as they are generated at some point in that process. We need to think of all Lagrange-related things in the repo as GKR-related things. Or specifically, the GKR instantiation of `AuxProof`. So concretely speaking, your random values of type 2 (Lagrange-kernel random values) are a subset of type 3 ("aux proof"-random values), but only in the case where the `AuxProof` is GKR.

I think this hints at a larger issue: we started explicitly supporting GKR in winterfell by adding the Lagrange kernel things, but then stopped halfway and made the `AuxProof` machinery generic over any auxiliary proof. I believe this explains most of the awkwardness of the current solution you're trying to fix here. If we moved the GKR implementation in winterfell, then it would enable us to make things a lot cleaner:

- We would get rid of `AuxProof`
  - The only `AuxProof` supported would be GKR
- Your suggestion for the `AuxRandElements` struct that separates the random values between "user" and `lagrange` would work well. The library would only ever need to know about `lagrange` random values (as an option), and nothing else, since we don't support any other type of auxiliary proof.
  - So we'd get rid of the `AuxRandElements` associated types too
- Get rid of `AuxTraceVerifier`
  - The `Air` would know whether or not GKR is being used, and would be either apply the GKR `AuxTraceVerifier`, or the default one that only generates `num_aux_rand_elements` "user" random values.
- etc.

It seems to me like we'd be able to drastically reduce the complexity of the implementation, and still be in line with our initial decision to explicitly supporting accelerating LogUp with GKR (with our introducing the Lagrange-specific machinery). It isn't clear to me anyway that the library is currently general/flexible enough to allow for other types of `AuxProof` out of the box. The only example we currently have (GKR) wasn't, since it needed the Lagrange kernel-specific machinery to work.

### My thoughts on next steps
- I would leave the implementation mostly as-is, since it works, and we'll be able to test out the GKR acceleration in the VM more quickly,
- If we like the results, and decide to switch the VM to using GKR for LogUp, then we move the GKR implementation in winterfell, and simplify the implementation as explained above.

prover/src/trace/mod.rs
```diff
@@ -89,8 +88,9 @@ pub trait Trace: Sized {
     fn validate<A, E>(
         &self,
         air: &A,
-        aux_segment: Option<&ColMatrix<E>>,
-        aux_rand_elements: &AuxTraceRandElements<E>,
+        aux_trace_with_metadata: Option<
+            &AuxTraceWithMetadata<E, <A as Air>::AuxRandElements<E>, <A as Air>::AuxProof>,
+        >,
```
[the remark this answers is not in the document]
plafer: Introduced an `AirAuxTraceWithMetadata` type

prover/src/lib.rs
```diff
@@ -262,7 +288,7 @@ pub trait Prover {
         // commit to the main trace segment
         let (mut trace_lde, mut trace_polys) = {
             // extend the main execution trace and build a Merkle tree from the extended trace
-            let span = info_span!("commit_to_main_trace_segment").entered();
+            let _span = info_span!("commit_to_main_trace_segment").entered();
```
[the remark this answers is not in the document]
plafer: Oh really? I don't understand why that would be, will have to look into it.

Re-added the `drop(span)` back in.

air/src/air/trace_info.rs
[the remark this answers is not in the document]
plafer: As a temporary solution (see [comment](https://github.com/facebook/winterfell/pull/271#discussion_r1589078479)), I think this shouldn't be a problem, since the number of random elements should always be known by both the prover and verifier as part of the problem definition? I believe we'd be able to clean this up properly if we brought the GKR implementation to winterfell.

verifier/src/aux_verifier.rs
irakliyk: > This isn't quite right. At the lowest level, the `Air` can't generate Lagrange kernel random values separately from building the GKR proof, as they are generated at some point in that process.

Ah! I see - so, I was incorrect here. My current understanding of the process is as follows:

1. After the main trace is built, we build the auxiliary proof.
2. In the process of building this proof, we generate random values needed for Lagrange kernel column.
3. We may also have some "user defined" aux rand values, which we can draw from the random coin after the auxiliary proof is generated (we could do it before as well, I think - as long as we remain consistent between prover and verifier)
4. We then use these random values to build the auxiliary trace (which would currently include building Largange kernel column "by hand" if needed).
5. We also use the randomness from step 2 and 3 during constraint evaluation.

> I think this hints at a larger issue: we started explicitly supporting GKR in winterfell by adding the Lagrange kernel things, but then stopped halfway and made the `AuxProof` machinery generic over any auxiliary proof. I believe this explains most of the awkwardness of the current solution you're trying to fix here.

I agree with this assessment. And thinking about it now, it seems to me that we can make some additional assumptions which would make the current implementation less general but at the same time, will be less complex and a smaller "diff" with the existing code.

Here is what I'm thinking:

First, I still think we should have two separate methods: `generate_aux_proof()` and `build_aux_trace()`. `generate_aux_proof()` could look like so:
```Rust
fn generate_aux_proof<E>(
    &self,
    main_trace: &Self::Trace,
    public_coin: &mut Self::RandomCoin,
) -> (ProverAuxProof<Self>, Vec<E>)
```
The returned `Vec<E>` would be explicitly interpreted as random values needed for Lagrange kernel column (we could also wrap them in a newtype - e.g., `LagrangeRandElements`).

After this function returns, we would build `AuxRandElements` struct which would be similar to the one in my previous comment.
```Rust
pub struct AuxRandElements<E> {
    user_defined: Vec<E>, // probably a better name is needed here
    lagrange: Vec<E>,
}
```
To build this struct, we could use `Air::build_aux_rand_elements()` which could look like so:
```Rust
fn get_aux_rand_elements<E, R>(
    &self,
    public_coin: &mut R,
    lagrange_rand_elements: Vec<E>,
) -> Result<Vec<E>, RandomCoinError>
```
This would work similarly to how `get_aux_trace_segment_random_elements()` works in the existing code and would also eliminate the need for a separate `Air::get_lagrange_rand_elements()` introduced in this PR.

Then, these `AuxRandElements` would be passed to the auxiliary trace constructor. The function could look as:
```Rust
fn build_aux_trace<E>(
    &self,
    main_trace: &Self::Trace,
    aux_rand_elements: &AuxRandElements<E>,
) -> ColMatrix<E>
```
We could actually move this function back into the `Trace` trait - but I'm also fine with keeping it in the `Prover` trait.

**Assumption**: one assumption I'm making here is that we are not duplicating a lot of work between `generate_aux_proof()` and `build_aux_trace()`. If we are, then it may make sense to rethink this - but also, this could be an optimization to make later.

One the verifier side, we would make analogous changes:

We'd have `AuxVerifier::verify()` function which could look like so:
```Rust
fn verify<E, Hasher>(
   &self,
   aux_proof: Self::AuxProof,
   public_coin: &mut impl RandomCoin<BaseField = E::BaseField, Hasher = Hasher>,
) -> Result<Vec<E>, Self::Error>
```
Where the returned `Vec<E>` would again be assumed to explicitly contain random values for the Lagrange column.

Then we'd use the returned values to construct `AuxRandElements` via `Air:: get_aux_rand_elements()` and the rest of the process would work as it does now.

I would also probably put the `AuxVerifier` as an associated type on the `Air` trait. So, basically, the two associated types that we'd need to add to the `Air` trait would be as follows:

```Rust
/// An auxiliary (i.e. non-STARK) proof object. If not needed, set to `()`.
type AuxProof: Serializable + Deserializable;

/// An auxiliary (i.e., non-STARK) verifier for verifying auxiliary proofs. If not needed, set to `()`.
type AuxVerifier: AuxVerifier<AuxProof = Self::AuxProof>;
```
I think this should allow us to go back to a single `verify()` function in the verifier.

If the above works, my intuition is that it will reduce the size of this PR somewhat, will position us better to transition to integrating GRK proofs into Winterfell, and will also introduce fewer API changes for this transient state of the library.

Al-Kindi-0: > We may also have some "user defined" aux rand values, which we can draw from the random coin after the auxiliary proof is generated (we could do it before as well, I think - as long as we remain consistent between prover and verifier)

I think "after" makes more sense, as this will cover:

1. The case when there is a Lagrange kernel: the subsequent drawn randomness will be used to construct the second auxiliary column alongside the Lagrange kernel column.
2. The case when there is no Lagrange kernel: the `generate_aux_proof` will be trivial and the subsequent drawn randomness will default to the auxiliary trace randomness as it currently exists.

air/src/air/aux.rs
```diff
@@ -0,0 +1,87 @@
...
+/// A trait for verifying a GKR proof.
+pub trait GkrVerifier {
+    /// The GKR proof.
+    type GkrProof: Deserializable;
+    /// The error that can occur during GKR proof verification.
+    type Error: ToString;
+
+    /// Verifies the GKR proof, and returns the random elements that will be used in building
+    /// the Lagrange kernel auxiliary column.
+    fn verify<E, Hasher>(
```
irakliyk, line 65: nit: should the comment say "were used in building" since the verifier doesn't build the column?

air/src/air/mod.rs
```diff
@@ -188,6 +192,12 @@ pub trait Air: Send + Sync {
     /// This could be any type as long as it can be serialized into a sequence of field elements.
     type PublicInputs: ToElements<Self::BaseField>;
 
+    /// An auxiliary (i.e. non-STARK) proof object. If not needed, set to `()`.
+    type GkrProof: Serializable + Deserializable;
+
+    /// An auxiliary (i.e., non-STARK) verifier for verifying GKR proofs. If not needed, set to [`DefaultGkrVerifier`].
+    type GkrVerifier: GkrVerifier<GkrProof = Self::GkrProof>;
```
irakliyk, line 199: minor nit: should the comments now mention GKR specifically? Also, we probably don't need a reference to the `DefaultGkrVerifier`.

air/src/air/mod.rs
```diff
@@ -283,20 +296,50 @@ pub trait Air: Send + Sync {
...
+    // --------------------------------------------------------------------------------------------
+
+    /// Returns the [`GkrVerifier`] to be used to verify the GKR proof.
+    ///
+    /// Leave unimplemented if the `Air` doesn't use a GKR proof.
+    fn get_auxiliary_proof_verifier<E: FieldElement<BaseField = Self::BaseField>>(
+        &self,
+    ) -> Self::GkrVerifier {
+        unimplemented!("`get_auxiliary_proof_verifier()` must be implemented when the proof contains a GKR proof");
+    }
```
irakliyk, line 314: Looking at this now, I wonder if we need the `GkrVerifier` trait at all. Maybe a single function here - e.g., `verify_gkr_proof()` would have sufficed.

air/src/air/mod.rs
```diff
@@ -283,20 +296,50 @@ pub trait Air: Send + Sync {
...
 
+    /// Returns a vector of field elements required for construction of the auxiliary trace segment.
+    ///
+    /// The elements are drawn uniformly at random from the provided public coin.
+    ///
+    /// TODOP: Document how these exclude the Lagrange ones, and how they're put in `AuxRandElements::rand_elements`.
+    fn get_aux_rand_elements<E, R>(&self, public_coin: &mut R) -> Result<Vec<E>, RandomCoinError>
+    where
+        E: FieldElement<BaseField = Self::BaseField>,
+        R: RandomCoin<BaseField = Self::BaseField>,
```
irakliyk, line 326: Do we not want to pass `Option<LagrangeRandElement>` into this and return `AuxRandElements` from here?

prover/src/trace/mod.rs
```diff
@@ -89,8 +92,7 @@ pub trait Trace: Sized {
     fn validate<A, E>(
         &self,
         air: &A,
-        aux_segment: Option<&ColMatrix<E>>,
-        aux_rand_elements: &AuxTraceRandElements<E>,
+        aux_trace_with_metadata: Option<&AirAuxTraceWithMetadata<A, E>>,
```
irakliyk, line 95: Should we also try to verify the auxiliary proof somewhere in this function?

verifier/src/errors.rs
```diff
@@ -13,6 +13,9 @@ use core::fmt;
 /// Represents an error returned by the verifier during an execution of the protocol.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum VerifierError {
+    /// This error occurs when the process of generating the random elements for the auxiliary trace
+    /// building failed.
+    AuxTraceVerificationFailed(String),
```
irakliyk, line 18: Should this say `AuxProofVerificationFailed`?

irakliyk (approved):
Looks good! Thank you! I left a few comments inline. Most of them are minor nits and the one about replacing `GkrVerifier` with a function can be addressed in a subsequent PR.

air/src/air/aux.rs
plafer: fixed

air/src/air/mod.rs
plafer: Oops, forgot those. Fixed

air/src/air/mod.rs
plafer: Agreed. IMO we should wait for the next PR, as this will change anyways

air/src/air/mod.rs
plafer: I preferred to relieve the user of this boilerplate; we know the user will never need to use the `Option<LagrangeRandElement>`, other than pass it in `AuxRandElements` at the end. I found it cleaner for us to be doing that, and let the user just worry about the "user-defined" aux rand elements

verifier/src/errors.rs
plafer: Renamed to `GkrProofVerificationFailed`, to be in line with the rest of the codebase (that now mentions `GkrProof`)

prover/src/trace/mod.rs
plafer: Created https://github.com/facebook/winterfell/issues/278

irakliyk (approved):
All looks good! Thank you!

PR #272, opened 2024-04-22
felt u64: add conversion to/from usize

hackaugusto:
This commit also:

- Reorders conversion definitions so that integer types are in
  increasing order
- Changes the conversions to use `.into()` instead of `as` casts, when
  the conversion is infallible. Instead of the cast like `as u128` which
  at first sight doesn't guarantee no truncation


irakliyk (approved):
Looks good! Thank you!

PR #275, opened 2024-05-02
Introduce `AsyncProver`

plafer:
Based on #271.

Introduced an asynchronous version of `Prover` (under a new `"async"` feature). A few notes:

- Uses the new [return-position `impl Trait` in traits](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html) of Rust 1.75. I chose not to use the `async` syntactic sugar, as it doesn't allow us to specify `+ Send + Sync` on the return `Future`s (as explained in the linked blog post).
  - This is a pragmatic solution, since the compiler team currently recommends using the [`trait-variant` crate](https://crates.io/crates/trait-variant), but the `trait_variant` macro didn't like our `E: FieldElement<BaseField = Self::BaseField>` bound in `generate_proof()` (with a difficult to decipher error message).
  - Also, I'm not sure that I like generating another trait just to get `+ Send` bounds. So I decided to only support the `+ Send + Sync` variant, so that it would work with `tokio`'s `spawn()`.
- We should fix the `tracing` spans in a few spots if we decide to go down this route
- After #271 is merged, I would split `Prover::generate_proof()` into many sub- functions, and call the appropriate ones in `AsyncProver::generate_proof()` so as to remove the amount of copy/paste code
- Personal TODO: make sure the CI will properly build/check the new `"async"` feature

I'm happy to revisit any of these assumptions 🙂.


plafer:
Closing as we will go in a different direction

Issue #277, opened 2024-05-04
`group_vector_elements` panics during account code compilation


JohnDonavon:
[group_vector_elements](https://github.com/facebook/winterfell/blob/09525751727a283dfbf5e569a09909b82380e059/utils/core/src/lib.rs#L119)
panics during account compilation for wasm built with rust nightly compiler or rust stable version 1.78.0

A simple example can be seen here: https://github.com/JohnDonavon/AssemblyWasm/blob/john-test/rust_lib/src/lib.rs

If you set the rust compiler to version 1.77.x, build the wasm (via wasm-pack or however you wish), and run the `complile` function from the browser it works. However, if you build this wasm with version 1.78.0 or nightly, the TransactionKernel assembler fails to compile the module and panics at the above mentioned function.

I've rewritten the function locally and the account code compiles successfully after. Here is the rewrite if it ends up being helpful:

```Rust
pub fn group_vector_elements_2<T, const N: usize>(source: Vec<T>) -> Vec<[T; N]> {
    assert_eq!(
        source.len() % N,
        0,
        "source length must be divisible by {}, but was {}",
        N,
        source.len()
    );

    // Capacity of the new vector
    let capacity = source.len() / N;
    let mut result = Vec::with_capacity(capacity);

    // SAFETY: We ensure the source length is exactly divisible by N
    // Convert Vec<T> into Vec<[T; N]>
    unsafe {
        // Convert source into a raw pointer for further manipulation
        let mut source_ptr = source.as_ptr();
        std::mem::forget(source); // Avoid destructor being called

        for _ in 0..capacity {
            // Point to an array of N elements. We advance by N each iteration.
            let array_ptr = source_ptr as *const [T; N];
            // Copy the array. This avoids aliasing issues by not modifying source_ptr directly.
            let array = std::ptr::read(array_ptr);
            result.push(array);
            source_ptr = source_ptr.add(N); // Move the pointer by N elements
        }
    }

    result
}
```

irakliyk:
Very interesting! Thank you for reporting this! I'm curious if the following modification would also fix the issue:
```Rust
pub fn group_vector_elements<T, const N: usize>(mut source: Vec<T>) -> Vec<[T; N]> {
    assert_eq!(
        source.len() % N,
        0,
        "source length must be divisible by {}, but was {}",
        N,
        source.len()
    );

    let len = source.len() / N;
    let cap = source.capacity() / N;

    // reduce the capacity of the vector to make sure it is a multiple of N
[...]

PR #282, opened 2024-05-09
Remove un-needed vector utilities

irakliyk:
This PR addresses https://github.com/facebook/winterfell/issues/277 and in the process removes `group_vector_elements()` and `FieldElement::zeroed_vector()` functions.

After doing some benchmarking, it seems like the impact of removing these functions is negligible (or maybe even non-existent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.9.0 (2024-05-09)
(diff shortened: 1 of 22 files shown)


bitwalker (commented):
LGTM. Seems like they've improved optimizations on `vec!` a bit, nice that we can just lean on that here!

PR #283, opened 2024-06-10
Swapped `maybe-async` for `maybe-async-await`

phklive:
In this PR I propose the swap from the `maybe-async` crate to `maybe-async-await` crate. 

Instead of relying on `maybe-async-rs` and `async-trait` we have implemented our own conditional `async/await` macros. 

These `#[maybe-async]` and `maybe_await!()` macros enable easy conditional asynchronicity and can be used as a standalone crate.


utils/maybe_async/Cargo.toml
```diff
@@ -1,26 +1,22 @@
 [package]
 name = "winter-maybe-async"
-version = "0.9.0"
+version = "0.1.0"
```
irakliyk, line 3: Let's change the version to `0.10.0`.

utils/maybe_async/Cargo.toml
```diff
@@ -1,26 +1,22 @@
...
 description = "sync/async macro for winterfell"
 authors = ["winterfell contributors"]
 readme = "README.md"
 license = "MIT"
 repository = "https://github.com/facebook/winterfell"
-documentation = "https://docs.rs/winter-maybe-async/0.9.0"
+documentation = ""
```
irakliyk, line 9: Documentation link should be `https://docs.rs/winter-maybe-async/0.10.0`

utils/maybe_async/src/lib.rs
irakliyk, line 1: We should add the copyright header for this file (as it was before) and also would be great to add basic doc comments to the macros.

utils/maybe_async/src/lib.rs
```diff
@@ -1,70 +1,43 @@
...
-            AsyncAwaitRemoval.remove_async_await(quote!(#item))
-        },
-        Item::Trait(item) => {
-            for inner in &mut item.items {
-                if let TraitItem::Fn(ref mut method) = inner {
-                    if method.sig.asyncness.is_some() {
-                        method.sig.asyncness = None;
-                    }
-                }
+pub fn maybe_async(_attr: TokenStream, input: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
phklive: Here `#[maybe_async]` is an attribute procedural macro attribute. Hence it can take some arguments e. g.

```rust
#[derive(Serialize, Debug)]
```

here `derive` is the macro and `Serialize, Debug` are the `attributes` (`_attrs`). 

In our macro we expect no arguments hence we omit them.

air/src/air/assertions/mod.rs
irakliyk, line 1: We can probably revert changes to this file.

air/src/air/context.rs
irakliyk, line 1: We can probably revert changes to this file.

utils/maybe_async/README.md
```diff
@@ -1,91 +1,47 @@
...
-
-    fn get_hello(&self) -> String {
-        "hello".into()
-    }
+// Adding `maybe_async` to regular functions
+#[maybe_async]
+fn hello_world() {
+    // ...
 }
 ```
```
irakliyk, line 24: I would probably add how the output would look like if `async` feature is on. Something like:

When the `async` feature is enabled, the above code will be transformed into:
```Rust
async trait ExampleTrait {
    async fn say_hello(&self);

    async fn get_hello(&self) -> String;
}

async fn hello_world() {
    // ...
}
```

utils/maybe_async/README.md
```diff
@@ -1,91 +1,47 @@
...
-Finally, `#[maybe_async]` can be used on `fn` items, which works in an analogous way to the previous examples.
-
-```rs
 #[maybe_async]
-async fn say_hello() {
-    // ...
+fn world() -> String {
+    "world".to_string()
 }
 ```
```
irakliyk, line 43: Similar to the last comment, I'd add. something like this after this:

When the `async` feature is enabled, the above code will be transformed into:
```Rust
async fn hello_world() {
    let w = world().await;

    println!("hello {}", w);
}

async fn world() -> String {
    "world".to_string()
}
```

air/src/air/context.rs
phklive: Clippy was not passing because of those comments.

air/src/air/assertions/mod.rs
phklive: Clippy was not passing because of those comments.

phklive: What should I do?

irakliyk: ah! then let's keep them.

air/src/air/context.rs
irakliyk: Ah! Then let's keep them.

utils/maybe_async/src/lib.rs
```diff
@@ -4,67 +4,53 @@
...
-fn convert_sync(input: &mut Item) -> TokenStream2 {
-    match input {
-        Item::Impl(item) => {
-            for inner in &mut item.items {
-                if let ImplItem::Fn(ref mut method) = inner {
-                    if method.sig.asyncness.is_some() {
-                        method.sig.asyncness = None;
-                    }
-                }
+pub fn maybe_async(_attr: TokenStream, input: TokenStream) -> TokenStream {
```
irakliyk, line 15: I would get rid of line 10 and add an example (same as in readme):

```Rust
/// Parses a function (regular or trait) and conditionally adds the `async` keyword
/// depending on the `async` feature flag being enabled.
///
/// For example:
/// ```ignore
/// trait ExampleTrait {
///     #[maybe_async]
///     fn say_hello(&self);
/// 
///     #[maybe_async]
///     fn get_hello(&self) -> String;
/// }
///
/// 
/// #[maybe_async]
/// fn hello_world() {
///     // ...
/// }
/// ```
/// When the `async` feature is enabled, will be transformed into:
/// ```ignore
/// trait ExampleTrait {
///     async fn say_hello(&self);
/// 
///     async fn get_hello(&self) -> String;
/// }
///
/// 
/// async fn hello_world() {
///     // ...
/// }
/// ```
```

utils/maybe_async/src/lib.rs
```diff
@@ -4,67 +4,53 @@
...
-        Item::Impl(item) => quote!(#[async_trait::async_trait]#item),
-        Item::Fn(item) => quote!(#item),
-        Item::Static(item) => quote!(#item),
-    }
+/// maybe_await procedural macro
+///
+/// Parses an expression and conditionally adds the `.await` keyword at the end of it
+/// depending on the `async` feature flag being enabled.
+#[proc_macro]
+pub fn maybe_await(input: TokenStream) -> TokenStream {
```
irakliyk, line 46: Similar comment as the one above re adding an example and removing line 41.

irakliyk (approved):
Looks good! Thank you! I'll make some of the remaining changes directly.

PR #287, opened 2024-06-21
Fix GKR-LogUp API

### LINKED ISSUE #286: GKR-LogUp: additional required API changes
The API changes in #271 to support GKR-LogUp were insufficient. As described in #257, GKR-LogUp relies on 2 auxiliary columns `l` (the Lagrange kernel) and `s`. As of #271, the prover currently doesn't have enough information to build the `s` column, and the Air doesn't have enough information to evaluate the `s` constraints.

## GKR-LogUp auxiliary columns recap

The GKR circuit proof leaves the following claims to be proven by the Air: $\tilde{f_0}(r), ..., \widetilde{f_{m-1}}(r)$, where $\tilde{f_i}$ is the MLE of the i'th main trace column $f_i$, and $r$ is a random point (referred to as "Lagrange kernel random elements" in the code). At the time of writing, in the Miden VM's LogUp GKR implementation, these claims correspond to the [`FinalOpeningClaim.openings`](https://github.com/0xPolygonMiden/miden-vm/blob/8b7b5c6739f240281a9abfd077d1ccb55e4f4f67/processor/src/trace/virtual_bus/sum_check/mod.rs#L70) field.

To prove these $m$ claims, we first combine them into a single claim by taking a random linear combination. Hence, the claim to prove becomes:

$$
\lambda_0 \tilde{f_0}(r) + \dots + \lambda_1 \widetilde{f_{m-1}}(r) = \sum_{i=0}^{m-1} \lambda_i \tilde{f_i}(r)
$$

This expression can be rearranged:

$$
\begin{align}
\sum_{i=0}^{m-1} \lambda_i \tilde{f_i}(r) &= \sum_{i=0}^{m-1} \lambda_i \sum_{x=0}^{n-1} \mathrm{eq}(x, r) f_i(x) \\
&= \sum_{x=0}^{n-1} \mathrm{eq}(x, r) \sum_{i=0}^{m-1} \lambda_i f_i(x)
\end{align}
$$

Notice that this is an inner product between $$\mathrm{eq}(x, r)$$ and $$\sum_{i=0}^{m-1} \lambda_i f_i(x).$$ The columns $l$ and $s$ will implement this inner product. Specifically,

- The $l$ column stores the Lagrange kernel. That is, the value at row $x$ is $\mathrm{eq}(x, r)$ (where $x$ is the bit decomposition of the row index), and $r$ is the above mentioned Lagrange kernel randomness.
- The $s$ column is a running sum where row $x$ stores the inner product from row $0$ to row $x$. Specifically, $$s(x) = s(x-1) + l(x) \sum_{i=0}^{m-1} \lambda_i f_i(x)$$

The constraints for $l$ were discussed and implemented in #271. The constraints for $s$ will tie everything together, ultimately ensuring that the prover claims for $\tilde{f_i}(r)$ indeed correspond to the MLE of $f_i$ evaluated at $r$. Specifically, the boundary constraints are

$$
\begin{align}
s[0] &= l[0] \sum_{i=0}^{m-1} \lambda_i f_i[0], \\
s[last] &= \sum_{i=0}^{m-1} \lambda_i \tilde{f_i}(r),
\end{align}
$$

where $\tilde{f_i}(r)$ are the above-mentioned prover claims left from the GKR proof. The transition constraint is:

$$
s[x+1] = s[x] + l[x+1] \sum_{i=0}^{m-1} \lambda_i f_i[x+1], \quad \forall x \in \{0, \dots, n-2\},
$$

where $n$ is the trace length.

## Required API changes

There are currently 2 issues with the current API.

1. the prover is currently missing access to the $\lambda_i$ values to build the aux trace, while 
2. the Air is currently missing access to both the $\lambda_i$ values and the $\tilde{f_i}(r)$ openings to evaluate the constraints.

Below are my suggestions on how to concretely achieve this (specific naming choices subject to change).

Define a new type `GkrRandElements` as

```rs
struct GkrRandElements<E> {
  lagrange: LagrangeRandElements,
  lambdas: Vec<E>
}
```

Consequently, `AuxRandElements` is modified to include this new type:

```rs
struct AuxRandElements<E> {
    rand_elements: Vec<E>,
    gkr: Option<GkrRandElements<E>>,
}
```

Then, `Prover::build_aux_trace(main_trace, aux_rand_elements)` now has all the information it needs to build the auxiliary trace, solving issue 1.

Then, `Prover::generate_gkr_proof()` now returns `GkrRandElements` instead of only `LagrangeKernelRandElements`. Consequently, `Prover::generate_gkr_proof()` the prover will also need to commit to commit to openings $\tilde{f_i}(r)$, and sample the $\lambda_i$ values. Analogously, `GkrVerifier::verify()` also needs to return `GkrRandElements`.

As for the changes in `Air`, I suggest we add 2 methods analogous to `Air::evaluate_aux_transition()` and `Air::get_aux_assertions()`, that are only called when a `GkrProof` is present, and pass in the $\lambda_i$ values, as well as the $\tilde{f_i}(r)$ openings (boundary constraints only). This solves issue 2.

Note that unlike the Lagrange kernel constraints, the constraints for the `s` column can be defined outside of winterfell, as they only require the typical `EvaluationFrame`.

plafer:
Closes #286 

I opted to expand the scope of the current `Air::evaluate_aux_transition()` and `Air::get_aux_assertions()` instead of creating new methods (as suggested in the issue), since it ended up being much cleaner.

### DIFF
--- a/air/src/air/aux.rs
+++ b/air/src/air/aux.rs
@@ -6,11 +6,11 @@ use utils::Deserializable;
 
 use super::lagrange::LagrangeKernelRandElements;
 
-// TODOP: fix all docs and naming
 /// Holds the randomly generated elements necessary to build the auxiliary trace.
 ///
-/// Specifically, [`AuxRandElements`] currently supports 2 types of random elements:
+/// Specifically, [`AuxRandElements`] currently supports 3 types of random elements:
 /// - the ones needed to build the Lagrange kernel column (when using GKR to accelerate LogUp),
+/// - the ones needed to build the "s" auxiliary column (when using GKR to accelerate LogUp),
 /// - the ones needed to build all the other auxiliary columns
 #[derive(Debug, Clone)]
 pub struct AuxRandElements<E> {
@@ -25,13 +25,13 @@ impl<E> AuxRandElements<E> {
         Self { rand_elements, gkr: None }
     }
 
-    /// Creates a new [`AuxRandElements`], where the auxiliary trace contains a Lagrange kernel
-    /// column.
+    /// Creates a new [`AuxRandElements`], where the auxiliary trace contains columns needed when
+    /// using GKR to accelerate LogUp (i.e. a Lagrange kernel column and the "s" column).
     pub fn new_with_gkr(rand_elements: Vec<E>, gkr: Option<GkrRandElements<E>>) -> Self {
         Self { rand_elements, gkr }
     }
 
-    /// Returns the random elements needed to build all columns other than the Lagrange kernel one.
+    /// Returns the random elements needed to build all columns other than the two GKR-related ones.
     pub fn rand_elements(&self) -> &[E] {
         &self.rand_elements
     }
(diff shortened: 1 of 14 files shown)


air/src/air/aux.rs
```diff
@@ -8,39 +8,72 @@ use super::lagrange::LagrangeKernelRandElements;
...
-    pub fn new_with_lagrange(
-        rand_elements: Vec<E>,
-        lagrange: Option<LagrangeKernelRandElements<E>>,
-    ) -> Self {
-        Self { rand_elements, lagrange }
+    /// Creates a new [`AuxRandElements`], where the auxiliary trace contains columns needed when
+    /// using GKR to accelerate LogUp (i.e. a Lagrange kernel column and the "s" column).
+    pub fn new_with_gkr(rand_elements: Vec<E>, gkr: Option<GkrRandElements<E>>) -> Self {
+        Self { rand_elements, gkr }
     }
```
irakliyk, line 32: Should the signature of this constructor be:
```rust
pub fn new_with_gkr(rand_elements: Vec<E>, gkr: GkrRandElements<E>) -> Self
```
That is, we assume that this constructor is used when GRK is enabled. This may require minor refactoring elsewhere, but I think it would be cleaner (and would also match to doc comments).

air/src/air/aux.rs
```diff
@@ -8,39 +8,72 @@ use super::lagrange::LagrangeKernelRandElements;
...
+    ///
+    /// These correspond to the lambdas in our documentation.
+    pub fn gkr_openings_combining_randomness(&self) -> Option<&[E]> {
+        self.gkr.as_ref().map(|gkr| gkr.openings_combining_randomness.as_ref())
+    }
+}
+
+/// Holds all the random elements needed when using GKR to accelerate LogUp.
+#[derive(Clone, Debug)]
+pub struct GkrRandElements<E> {
```
irakliyk, line 54: Let's expand the doc comment a bit more to explain that this consists of Lagrange kernel randomness and the "opening combining" randomness (i.e., something similar to 12 - 13 above - or maybe with a bit more detail).

air/src/air/aux.rs
```diff
@@ -8,39 +8,72 @@ use super::lagrange::LagrangeKernelRandElements;
...
+    lagrange: LagrangeKernelRandElements<E>,
+    openings_combining_randomness: Vec<E>,
+}
+
+impl<E> GkrRandElements<E> {
+    pub fn new(
```
irakliyk, line 60: nit: since this is a member of public API, let's add a doc comment here.

irakliyk (approved):
Looks good! Thank you! I left a few small comments inline. After these are addressed, we can merge.

I do have some bigger comments which go beyond this PR - I'll write them up in the issue.

air/src/air/aux.rs
plafer: Good point; fixed

air/src/air/aux.rs
plafer: done

air/src/air/aux.rs
plafer: Done

irakliyk (approved):
All looks good! Thank you!

PR #288, opened 2024-06-25
Final LogUp-GKR fixes

plafer:
Final changes needed in https://github.com/0xPolygonMiden/miden-vm/pull/1364

### DIFF
--- a/prover/src/constraints/evaluator/lagrange.rs
+++ b/prover/src/constraints/evaluator/lagrange.rs
@@ -10,6 +10,7 @@ use air::{
     LagrangeKernelEvaluationFrame, LagrangeKernelRandElements,
 };
 use math::{batch_inversion, FieldElement};
+use tracing::instrument;
 
 use crate::{StarkDomain, TraceLde};
 
@@ -49,6 +50,7 @@ impl<E: FieldElement> LagrangeKernelConstraintsBatchEvaluator<E> {
     ///
     /// Writes the evaluations in `combined_evaluations_acc` at the corresponding (constraint
     /// evaluation) domain index.
+    #[instrument(skip_all)]
     pub fn evaluate_constraints<T>(
         &self,
         trace: &T,
--- a/prover/src/trace/poly_table.rs
+++ b/prover/src/trace/poly_table.rs
@@ -6,7 +6,7 @@
 use alloc::vec::Vec;
 
 use air::{proof::TraceOodFrame, LagrangeKernelEvaluationFrame};
-use math::{FieldElement, StarkField};
+use math::{polynom, FieldElement, StarkField};
 
 use crate::{matrix::ColumnIter, ColMatrix};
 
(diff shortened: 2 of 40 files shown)


irakliyk (commented):
Looks good! Thank you! Some tests don't seem to be passing though.

Also question: is the main reason for passing GKR proof to constraint evaluator is to enable boundary constraint computation for the `s` column? Or is it something else?

prover/src/trace/poly_table.rs
```diff
@@ -74,6 +74,10 @@ impl<E: FieldElement> TracePolyTable<E> {
         for aux_polys in self.aux_trace_polys.iter() {
             result.append(&mut aux_polys.evaluate_columns_at(x));
         }
+
+        if let Some(lagrange_kernel_poly) = &self.lagrange_kernel_poly {
+            result.push(polynom::eval(lagrange_kernel_poly, x));
+        }
```
plafer, line 80: Note: we also need to pass this here so that this column is also available in the OOD frame, needed by the "s" column's transition constraint.

plafer:
> Also question: is the main reason for passing GKR proof to constraint evaluator is to enable boundary constraint computation for the s column? Or is it something else?

Yes exactly. Specifically the constraint for the last row, implemented [here](https://github.com/0xPolygonMiden/miden-vm/blob/35484c4fcd0c1f63c473c1e71b42f8971c81f864/air/src/constraints/logup.rs#L72-L73) in the VM.

irakliyk (approved):
Looks good! Thank you!

irakliyk:
Could we also add instrumentation for the Lagrange kernel column constraints evaluation (i.e., [here](https://github.com/facebook/winterfell/blob/5db0310fd7a5e24a3f56eedafb976119b28fbe97/prover/src/constraints/evaluator/lagrange.rs#L52))?

plafer:
> Could we also add instrumentation for the Lagrange kernel column constraints evaluation (i.e., [here](https://github.com/facebook/winterfell/blob/5db0310fd7a5e24a3f56eedafb976119b28fbe97/prover/src/constraints/evaluator/lagrange.rs#L52))?

Done

prover/src/constraints/evaluator/lagrange.rs
```diff
@@ -49,6 +50,7 @@ impl<E: FieldElement> LagrangeKernelConstraintsBatchEvaluator<E> {
     ///
     /// Writes the evaluations in `combined_evaluations_acc` at the corresponding (constraint
     /// evaluation) domain index.
+    #[instrument(skip_all)]
     pub fn evaluate_constraints<T>(
```
irakliyk, line 54: Is there a way to customize how this function appears in the logs? For example, right now it appears in the logs as `evaluate_constraints` but would be great if it appeared as `evaluate_lagrange_kernel_constraints` or something similar.

plafer:
Closing as stale

PR #294, opened 2024-08-05
Add multi-linear polynomial support

Al-Kindi-0:
This PR adds a few structs, methods and helper functions necessary for the sum-check protocol (to be included in a separate PR).

### DIFF
--- a/math/Cargo.toml
+++ b/math/Cargo.toml
@@ -32,13 +32,15 @@ name = "poly_div_lagrange"
 harness = false
 
 [features]
-concurrent = ["utils/concurrent", "std"]
+concurrent = ["utils/concurrent", "rayon", "std"]
 default = ["std"]
 std = ["utils/std"]
 
 [dependencies]
 serde = { version = "1.0", features = [ "derive" ], optional = true, default-features = false }
+thiserror = { version = "1.0", git = "https://github.com/bitwalker/thiserror", branch = "no-std", default-features = false }
 utils = { version = "0.9", path = "../utils/core", package = "winter-utils", default-features = false }
+rayon = { version = "1.8", optional = true }
 
 [dev-dependencies]
 criterion = "0.5"
--- a/math/src/polynom/multi_linear.rs
+++ b/math/src/polynom/multi_linear.rs
@@ -1,0 +1,244 @@
+use alloc::vec::Vec;
+use core::ops::Index;
+
+#[cfg(feature = "concurrent")]
+pub use rayon::prelude::*;
+use utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable};
+
+use crate::FieldElement;
+
+// MULTI-LINEAR POLYNOMIAL
+// ================================================================================================
+
+/// Represents a multi-linear polynomial.
+///
+/// The representation stores the evaluations of the polynomial over the boolean hyper-cube
+/// ${0 , 1}^ν$.
+#[derive(Clone, Debug, PartialEq)]
+pub struct MultiLinearPoly<E: FieldElement> {
+    evaluations: Vec<E>,
+}
+
+impl<E: FieldElement> MultiLinearPoly<E> {
+    /// Constructs a [`MultiLinearPoly`] from its evaluations over the boolean hyper-cube ${0 , 1}^ν$.
+    pub fn from_evaluations(evaluations: Vec<E>) -> Self {
+        if !evaluations.len().is_power_of_two() {
+            panic!("A multi-linear polynomial should have a power of 2 number of evaluations over the Boolean hyper-cube")
+        }
+        Self { evaluations }
+    }
+
+    /// Returns the number of variables of the multi-linear polynomial.
+    pub fn num_variables(&self) -> usize {
+        self.evaluations.len().trailing_zeros() as usize
+    }
+
+    /// Returns the evaluations over the boolean hyper-cube.
+    pub fn evaluations(&self) -> &[E] {
+        &self.evaluations
+    }
+
+    /// Returns the number of evaluations. This is equal to the size of the boolean hyper-cube.
+    pub fn num_evaluations(&self) -> usize {
+        self.evaluations.len()
+    }
+
+    /// Evaluate the multi-linear at some query $(r_0, ..., r_{ν - 1}) ∈ 𝔽^ν$.
+    ///
+    /// It first computes the evaluations of the Lagrange basis polynomials over the interpolating
+    /// set ${0 , 1}^ν$ at $(r_0, ..., r_{ν - 1})$ i.e., the Lagrange kernel at $(r_0, ..., r_{ν - 1})$.
+    /// The evaluation then is the inner product, indexed by ${0 , 1}^ν$, of the vector of
+    /// evaluations times the Lagrange kernel.
+    pub fn evaluate(&self, query: &[E]) -> E {
+        let tensored_query = compute_lagrange_basis_evals_at(query);
+        inner_product(&self.evaluations, &tensored_query)
+    }
+
+    /// Similar to [`Self::evaluate`], except that the query was already turned into the Lagrange
+    /// kernel (i.e. the [`lagrange_ker::EqFunction`] evaluated at every point in the set
+    /// `${0 , 1}^ν$`).
+    ///
+    /// This is more efficient than [`Self::evaluate`] when multiple different [`MultiLinearPoly`]
+    /// need to be evaluated at the same query point.
+    pub fn evaluate_with_lagrange_kernel(&self, lagrange_kernel: &[E]) -> E {
+        inner_product(&self.evaluations, lagrange_kernel)
+    }
+
+    /// Computes $f(r_0, y_1, ..., y_{ν - 1})$ using the linear interpolation formula
+    /// $(1 - r_0) * f(0, y_1, ..., y_{ν - 1}) + r_0 * f(1, y_1, ..., y_{ν - 1})$ and assigns
+    /// the resulting multi-linear, defined over a domain of half the size, to `self`.
+    pub fn bind_least_significant_variable(&mut self, round_challenge: E) {
+        let num_evals = self.evaluations.len() >> 1;
+        for i in 0..num_evals {
+            self.evaluations[i] = self.evaluations[i << 1]
+                + round_challenge * (self.evaluations[(i << 1) + 1] - self.evaluations[i << 1]);
+        }
+        self.evaluations.truncate(num_evals)
+    }
+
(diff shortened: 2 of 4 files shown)


math/src/polynom/univariate.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Clone, Debug)]
+pub struct CompressedUnivariatePolyCoef<E: FieldElement>(Vec<E>);
+
+impl<E: FieldElement> CompressedUnivariatePolyCoef<E> {
+    /// Evaluates a polynomial at a challenge point using a round claim.
+    ///
+    /// The round claim is used to recover the coefficient of the linear term using the relation
+    /// 2 * c0 + c1 + ... c_{n - 1} = claim. Using the complete list of coefficients, the polynomial
+    /// is then evaluated using Horner's method.
+    pub fn evaluate_using_claim(&self, claim: &E, challenge: &E) -> E {
```
irakliyk, line 24: nit: should we rename `challenge` to `x` as we usually use `x` for variables at which we evaluate polynomials?

math/src/polynom/univariate.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+impl<E> Deserializable for CompressedUnivariatePolyCoef<E>
+where
+    E: FieldElement,
+{
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        Ok(Self(Deserializable::read_from(source)?))
+    }
+}
```
irakliyk, line 50: I would probably defer adding these until we actually need them as we may end up using a different way to serialize things.

math/src/polynom/univariate.rs
```diff
@@ -0,0 +1,263 @@
...
+
+/// The evaluations of a univariate polynomial of degree n at 0, 1, ..., n with the evaluation at 0
+/// omitted.
+///
+/// This compressed representation is useful during the sum-check protocol as the full uncompressed
+/// representation can be recovered from the compressed one and the current sum-check round claim.
+#[derive(Clone, Debug)]
+pub struct CompressedUnivariatePolyEvals<E>(Vec<E>);
```
irakliyk, line 58: Similar to one of the previous comments: I wonder if we should use `SmallVec` here instead of `Vec`.

math/src/polynom/multi_linear.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct MultiLinearPoly<E: FieldElement> {
+    evaluations: Vec<E>,
+}
+
+impl<E: FieldElement> MultiLinearPoly<E> {
+    /// Constructs a [`MultiLinearPoly`] from its evaluations over the boolean hyper-cube ${0 , 1}^ν$.
+    pub fn from_evaluations(evaluations: Vec<E>) -> Self {
+        if !evaluations.len().is_power_of_two() {
+            panic!("A multi-linear polynomial should have a power of 2 number of evaluations over the Boolean hyper-cube")
+        }
```
irakliyk, line 27: nit: I would write this as `assert!(evaluations.len().is_power_of_two(), "A multi-linear...");`.

math/src/polynom/multi_linear.rs
```diff
@@ -0,0 +1,244 @@
...
+impl<E> Deserializable for MultiLinearPoly<E>
+where
+    E: FieldElement,
+{
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        Ok(Self {
+            evaluations: Deserializable::read_from(source)?,
+        })
+    }
+}
```
irakliyk, line 120: Similar to one other comment: I'd probably defer adding these to later.

math/src/polynom/multi_linear.rs
```diff
@@ -0,0 +1,244 @@
...
+    #[cfg(not(feature = "concurrent"))]
+    return x.iter().zip(y.iter()).fold(E::ZERO, |acc, (x_i, y_i)| acc + *x_i * *y_i);
+
+    #[cfg(feature = "concurrent")]
+    return x
+        .par_iter()
+        .zip(y.par_iter())
+        .map(|(x_i, y_i)| *x_i * *y_i)
+        .reduce(|| E::ZERO, |a, b| a + b);
+}
```
irakliyk, line 230: Nice!

I would mention in the doc comments that this function is uses multiple threads when `concurrent` feature is enabled.

math/src/polynom/multi_linear.rs
```diff
@@ -0,0 +1,244 @@
...
+        MultiLinearPoly::from_evaluations(eq_evals)
+    }
+}
+
+// HELPER
+// ================================================================================================
+
+/// Computes the evaluations of the Lagrange basis polynomials over the interpolating
+/// set ${0 , 1}^ν$ at $(r_0, ..., r_{ν - 1})$ i.e., the Lagrange kernel at $(r_0, ..., r_{ν - 1})$.
+fn compute_lagrange_basis_evals_at<E: FieldElement>(query: &[E]) -> Vec<E> {
```
irakliyk, line 201: If there is no straightforward way to parallelize this right now, let's add a `TODO` comment to make sure we know that this would be an important function to parallelize.

math/src/polynom/multi_linear.rs
```diff
@@ -0,0 +1,244 @@
...
+///
+/// $((y_0, ..., y_{ν - 1}) ↦ EQ((r_0, ..., r_{ν - 1}), (y_0, ..., y_{ν - 1})))$
+///
+/// and contains a method to generate the Lagrange kernel for defining evaluations of multi-linear
+/// extensions of arbitrary functions $(f: {0 , 1}^ν ⇾ 𝔽)$ at a given point $(r_0, ..., r_{ν - 1})$
+/// as well as a method to evaluate $EQ^{~}((r_0, ..., r_{ν - 1}), (t_0, ..., t_{ν - 1})))$ for
+/// $(t_0, ..., t_{ν - 1}) ∈ 𝔽^ν$.
+pub struct EqFunction<E> {
+    r: Vec<E>,
+}
```
irakliyk, line 163: Is `r` expected to be a relatively small vector here? If so, maybe worth using `SmallVec` here as well.

math/src/polynom/multi_linear.rs
```diff
@@ -0,0 +1,244 @@
...
+    #[cfg(feature = "concurrent")]
+    return x
+        .par_iter()
+        .zip(y.par_iter())
+        .map(|(x_i, y_i)| *x_i * *y_i)
+        .reduce(|| E::ZERO, |a, b| a + b);
+}
+
+// TESTS
+// ================================================================================================
```
irakliyk, line 233: We should probably add more tests to make sure everything works as expected.

irakliyk (commented):
Looks good! Thank you! I left some comments inline.

More generally, it seems to me that structs introduced in this PR are relevant specifically for the sumcheck protocol. So, maybe it makes sense to create a `winter-sumcheck` crate and move them there?

This crate would contain the two types of sumcheck that are relevant specifically for LogUp-GKR and we can make this explicit in the crate description etc. (i.e., we are not trying to build a generic implementation of sumcheck - but need only what is relevant for our current use case).

What do you think?

math/Cargo.toml
```diff
@@ -32,13 +32,15 @@ name = "poly_div_lagrange"
...
 
 [features]
-concurrent = ["utils/concurrent", "std"]
+concurrent = ["utils/concurrent", "rayon", "std"]
 default = ["std"]
 std = ["utils/std"]
 
 [dependencies]
 serde = { version = "1.0", features = [ "derive" ], optional = true, default-features = false }
+thiserror = { version = "1.0", git = "https://github.com/bitwalker/thiserror", branch = "no-std", default-features = false }
```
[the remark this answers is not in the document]
Al-Kindi-0: No, but for sumcheck yes

math/Cargo.toml
```diff
@@ -32,13 +32,15 @@ name = "poly_div_lagrange"
 harness = false
 
 [features]
-concurrent = ["utils/concurrent", "std"]
+concurrent = ["utils/concurrent", "rayon", "std"]
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

math/src/polynom/univariate.rs
```diff
@@ -0,0 +1,263 @@
...
+
+impl<E: FieldElement> CompressedUnivariatePolyEvals<E> {
+    /// Gives the coefficient representation of a polynomial represented in evaluation form.
+    ///
+    /// Since the evaluation at 0 is omitted, we need to use the round claim to recover
+    /// the evaluation at 0 using the identity p(0) + p(1) = claim.
+    /// Now, we have that for any polynomial p(x) = c0 + c1 * x + ... + c_{n-1} * x^{n - 1}:
+    ///
+    /// 1. p(0) = c0.
+    /// 2. p(x) = c0 + x * q(x) where q(x) = c1 + ... + c_{n-1} * x^{n - 2}.
```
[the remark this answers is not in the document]
Al-Kindi-0: Thanks! I have changed it here and elsewhere to the proposal

math/src/polynom/univariate.rs
```diff
@@ -0,0 +1,263 @@
...
+// COMPRESSED UNIVARIATE POLYNOMIAL
+// ================================================================================================
+
+/// The coefficients of a univariate polynomial of degree n with the linear term coefficient
+/// omitted.
+///
+/// This compressed representation is useful during the sum-check protocol as the full uncompressed
+/// representation can be recovered from the compressed one and the current sum-check round claim.
+#[derive(Clone, Debug)]
+pub struct CompressedUnivariatePolyCoef<E: FieldElement>(Vec<E>);
```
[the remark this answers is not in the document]
Al-Kindi-0: Makes sense. Changed

math/src/polynom/univariate.rs
Al-Kindi-0: Changed

math/src/polynom/multi_linear.rs
Al-Kindi-0: Changed

math/src/polynom/multi_linear.rs
Al-Kindi-0: Added

math/src/polynom/multi_linear.rs
Al-Kindi-0: Changed

math/src/polynom/multi_linear.rs
Al-Kindi-0: Added more tests

math/src/polynom/univariate.rs
Al-Kindi-0: Removed here but will introduce them in the next PR

PR #295, opened 2024-08-07
Sumcheck IOP for LogUp-GKR

Al-Kindi-0:
Adds a `sumcheck` crate containing the necessary tools for the LogUp-GKR protocol.

PR #296, opened 2024-08-09
GKR backend for LogUp-GKR

Al-Kindi-0:
Builds on #295 and implements the GKR backend for the LogUp-GKR argument.


air/src/air/mod.rs
```diff
@@ -303,15 +305,13 @@ pub trait Air: Send + Sync {
...
-    /// Returns the [`GkrVerifier`] to be used to verify the GKR proof.
-    ///
-    /// Leave unimplemented if the `Air` doesn't use a GKR proof.
-    fn get_gkr_proof_verifier<E: FieldElement<BaseField = Self::BaseField>>(
+    /// Returns the object needed for the LogUp-GKR argument.
+    fn get_logup_gkr_evaluator<E: FieldElement<BaseField = Self::BaseField>>(
         &self,
-    ) -> Self::GkrVerifier {
+    ) -> Self::LogUpGkrEvaluator {
         unimplemented!("`get_auxiliary_proof_verifier()` must be implemented when the proof contains a GKR proof");
```
plafer, line 315: This message needs to be updated with the new method name

air/src/air/context.rs
```diff
@@ -232,12 +241,16 @@ impl<B: StarkField> AirContext<B> {
...
+        } else {
+            None
+        }
     }
 
-    /// Returns true if the auxiliary trace segment contains a Lagrange kernel column
-    pub fn has_lagrange_kernel_aux_column(&self) -> bool {
-        self.lagrange_kernel_aux_column_idx().is_some()
+    /// Returns true if LogUp-GKR is enabled.
+    pub fn is_with_logup_gkr(&self) -> bool {
```
plafer, line 252:
```suggestion
    pub fn uses_logup_gkr(&self) -> bool {
```

air/src/air/logup_gkr.rs
```diff
@@ -0,0 +1,138 @@
...
+    PeriodicValue(Vec<B>),
+}
+
+impl LogUpGkrEvaluator for () {
+    type BaseField = BaseElement;
+
+    type PublicInputs = ();
+
+    fn get_oracles(&self) -> Vec<LogUpGkrOracle<Self::BaseField>> {
+        unimplemented!()
```
plafer, line 97: nit: I would instead `panic` with an informative error message such as `LogUpGkrEvaluator method X called, but LogUpGkr is not enabled` or something like that

air/src/air/logup_gkr.rs
```diff
@@ -0,0 +1,138 @@
...
+        E::ZERO
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, PartialOrd, Eq, Ord)]
+pub enum LogUpGkrOracle<B: StarkField> {
+    CurrentRow(usize),
+    NextRow(usize),
+    PeriodicValue(Vec<B>),
+}
```
plafer, line 89: nit: can we define this before or after the `impl LogUpGkrEvaluator for ()` block, such that all the `LogUpGkrEvaluator` code is in 1 place?

air/src/air/logup_gkr.rs
```diff
@@ -0,0 +1,138 @@
...
+        E: FieldElement<BaseField = Self::BaseField>,
+    {
+        E::ZERO
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, PartialOrd, Eq, Ord)]
+pub enum LogUpGkrOracle<B: StarkField> {
+    CurrentRow(usize),
+    NextRow(usize),
```
plafer, line 87:
```suggestion
    CurrentRow { col_idx: usize },
    NextRow { col_idx: usize },
```
This is a bit more descriptive than just a `usize`, where I had to go look at how it is used to infer that it stores the column index

air/src/air/aux.rs
```diff
@@ -60,70 +62,66 @@ impl<E> AuxRandElements<E> {
...
 /// openings combining randomness.
 #[derive(Clone, Debug)]
-pub struct GkrRandElements<E> {
-    lagrange: LagrangeKernelRandElements<E>,
-    openings_combining_randomness: Vec<E>,
+pub struct GkrData<E: FieldElement> {
+    pub lagrange_kernel_eval_point: LagrangeKernelRandElements<E>,
+    pub openings_combining_randomness: Vec<E>,
+    pub openings: Vec<E>,
+    pub oracles: Vec<LogUpGkrOracle<E::BaseField>>,
```
plafer, line 69: I don't think it is necessary to store the oracles here

air/src/air/logup_gkr.rs
```diff
@@ -0,0 +1,138 @@
...
+pub trait LogUpGkrEvaluator: Clone + Sync {
+    /// Defines the base field of the evaluator.
+    type BaseField: StarkField;
+
+    /// Public inputs need to compute the final claim.
+    type PublicInputs: ToElements<Self::BaseField> + Send;
+
+    /// Gets a list of all oracles involved in LogUp-GKR; this is intended to be used in construction of
+    /// MLEs.
+    fn get_oracles(&self) -> Vec<LogUpGkrOracle<Self::BaseField>>;
```
plafer, line 28:
```suggestion
    fn get_oracles(&self) -> &[LogUpGkrOracle<Self::BaseField>] {
```

This is called in many different places, which results in a new allocation everytime. I think it should work to only create the oracles once, and return a slice with `get_oracles()`

prover/src/logup_gkr/prover.rs
```diff
@@ -0,0 +1,256 @@
...
+        log_up_randomness,
+        multi_linear_ext_polys,
+        transcript,
+    )?;
+
+    Ok(FinalLayerProof::new(proof))
+}
+
+// TODO: Make the multi-linears over the base field and define an operation of folding with a challenge
+// in an extension field.
```
plafer, line 124: Can we create an issue for this and remove this `TODO`? It'll be easier to track

plafer (commented):
not a full review yet

prover/src/logup_gkr/prover.rs
```diff
@@ -0,0 +1,256 @@
...
+fn build_mls_from_main_trace_segment<E: FieldElement>(
+    oracles: Vec<air::LogUpGkrOracle<E::BaseField>>,
+    main_trace: &ColMatrix<<E as FieldElement>::BaseField>,
+) -> Result<Vec<MultiLinearPoly<E>>, GkrProverError> {
+    let mut mls = vec![];
+
+    for oracle in oracles {
+        match oracle {
+            air::LogUpGkrOracle::CurrentRow(index) => {
```
plafer, line 133:
```suggestion
            LogUpGkrOracle::CurrentRow(index) => {
```

Let's import `LogUpGkrOracle`

air/src/air/mod.rs
Al-Kindi-0: fixed

air/src/air/context.rs
Al-Kindi-0: fixed

air/src/air/logup_gkr.rs
Al-Kindi-0: fixed

air/src/air/logup_gkr.rs
Al-Kindi-0: Added a comment to add more context

air/src/air/logup_gkr.rs
Al-Kindi-0: Added a comment to add more context

air/src/air/aux.rs
Al-Kindi-0: I think it is needed for the s-column, which is done in another PR building on this one

air/src/air/logup_gkr.rs
Al-Kindi-0: changed

prover/src/logup_gkr/prover.rs
Al-Kindi-0: done

prover/src/logup_gkr/prover.rs
Al-Kindi-0: Created

prover/src/logup_gkr/prover.rs
```diff
@@ -0,0 +1,256 @@
...
+///
+/// \sum_{y ∈ {0,1}^μ} EQ(ρ', y) * g_{[y]}(f_0(x_0, ... , x_{ν - 1}), ... , f_{κ - 1}(x_0, ... ,
+/// x_{ν - 1}))
+///
+/// where ρ' is the randomness sampled during the first sum-check protocol.
+///
+/// As part of the final sum-check protocol, the openings {f_j(ρ)} are provided as part of a
+/// [`FinalOpeningClaim`]. This latter claim will be proven by the STARK prover later on using the
+/// auxiliary trace.
+pub fn prove_gkr<E: FieldElement>(
```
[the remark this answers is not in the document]
Al-Kindi-0: I see your point. On the other hand, some of the structs (e.g., `GkrCircuitProof`) as well many of the comments fit better with the current name.
No strong preference though

prover/src/logup_gkr/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+        let mut query = vec![E::BaseField::ZERO; evaluator.get_oracles().len()];
+        let mut numerators = vec![E::ZERO; num_fractions];
+        let mut denominators = vec![E::ZERO; num_fractions];
+        for i in 0..main_trace.main_segment().num_rows() {
+            let wires_from_trace_row = {
+                main_trace.read_main_frame(i, &mut main_frame);
+
+                evaluator.build_query(&main_frame, &[], &mut query);
+
+                evaluator.evaluate_query(
```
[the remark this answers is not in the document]
Al-Kindi-0: `build_query` is useful for building the input layer and the OOD evaluation. For the first one we can merge the two but for the second one the `evaluate_query` logic is not needed.
One the other hand, `evaluate_query` is used without the `build_query` logic in the high degree sum-check.

sumcheck/src/lib.rs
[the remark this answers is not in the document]
Al-Kindi-0: Fixed, thank you!

prover/src/logup_gkr/mod.rs
```diff
@@ -0,0 +1,441 @@
...
+    ) -> CircuitLayer<E> {
+        let num_fractions = evaluator.get_num_fractions();
+        let mut input_layer_wires =
+            Vec::with_capacity(main_trace.main_segment().num_rows() * num_fractions);
+        let mut main_frame = EvaluationFrame::new(main_trace.main_segment().num_cols());
+
+        let mut query = vec![E::BaseField::ZERO; evaluator.get_oracles().len()];
+        let mut numerators = vec![E::ZERO; num_fractions];
+        let mut denominators = vec![E::ZERO; num_fractions];
+        for i in 0..main_trace.main_segment().num_rows() {
```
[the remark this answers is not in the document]
Al-Kindi-0: My thinking was that, at the "application" level, the selectors will be designed so as to be zero for such edge cases. Thus I didn't give much thought to a choice we can make with respect to `NextRow` oracles, namely either we choose a circular shift, as we do now, or we use a plain shift where we will get zero at the last row as a result.

prover/src/logup_gkr/mod.rs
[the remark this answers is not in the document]
plafer: Good point, I missed that.

prover/src/logup_gkr/mod.rs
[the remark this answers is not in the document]
plafer: > namely either we choose a circular shift, as we do now, or we use a plain shift where we will get zero at the last row as a result.

Right, we currently do a circular shift. But we actually want a plain shift, and I don't see how this would be possible with the current API. For example, when `x` represents the last row, how does `evaluate_query()` know that ultimately it should output 0s for the numerators and 1s for the denominator (since all it gets is the evaluations of the oracles at `x`)?

I'm currently not seeing it, but if you're confident that it's possible, then we can move on.

examples/src/fibonacci/fib2/air.rs
```diff
@@ -69,3 +68,57 @@ impl Air for FibAir {
         ]
     }
 }
+
+#[derive(Clone, Default)]
+pub struct PlainLogUpGkrEval<B: FieldElement> {
```
plafer, line 73: We should export a `DummyLogUpGkrEval<BaseField, PublicInputs>` can be used when LogUp-GKR is disabled (instead of reimplementing it everytime)

plafer (commented):
LGTM!

I'm still not sure if the `LogUpGkrEvaluator` API is flexible enough to properly support the use of the `NextRow` `GkrOracle`, but if it isn't we can always fix it in a subsequent PR

prover/src/logup_gkr/mod.rs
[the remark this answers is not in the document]
Al-Kindi-0: For example, the numerators could be an expression involving selectors which are zero once a `HALT` instruction has been initiated. For this row, and all subsequent rows assuming that a `HALT` can only be followed by a `HALT`, the numerators will be zero. The denominators are not important as far as I can see, of course as long as they are not zero, and this is guaranteed with high probability by the LogUp randomness.
I might be missing something extra but we can, as you suggested, fix it in a subsequent targeted PR.

examples/src/fibonacci/fib2/air.rs
Al-Kindi-0: Good idea!
Implemented and exported

air/src/air/trace_info.rs
```diff
@@ -177,7 +180,7 @@ impl TraceInfo {
 
     /// Returns the number of columns in the auxiliary segment of an execution trace.
     pub fn aux_segment_width(&self) -> usize {
-        self.aux_segment_width
+        self.aux_segment_width + 2 * self.logup_gkr as usize
     }
```
irakliyk, line 184: I would mention in the doc comment that when LogUp-GKR is enabled, two extra columns are added to the auxiliary trace.

Also, should we rename this to `aux_trace_width()`? To keep it parallel with `main_trace_width()`.

air/src/air/trace_info.rs
```diff
@@ -198,9 +201,8 @@ impl TraceInfo {
         }
     }
 
-    /// Returns the number of columns in the auxiliary trace segment.
-    pub fn get_aux_segment_width(&self) -> usize {
-        self.aux_segment_width
+    pub fn is_with_logup_gkr(&self) -> bool {
+        self.logup_gkr
     }
```
irakliyk, line 206: Let's add a doc comment to this method.

Also, maybe something like `logup_gkr_enabled()` could be a better method name? Not a strong preference though - so, feel free to ignore.

air/src/air/logup_gkr.rs
```diff
@@ -81,9 +82,72 @@ pub trait LogUpGkrEvaluator: Clone + Sync {
     }
 }
 
+#[derive(Clone, Default)]
+pub struct DummyLogUpGkrEval<B: StarkField, P: Clone + Send + Sync + ToElements<B>> {
+    _field: PhantomData<B>,
+    _public_inputs: PhantomData<P>,
+}
```
irakliyk, line 89: I don't really love "Dummy" prefix here since we use it to indicate that we are not using LogUp-GKR. Maybe something like "Void", "Null", or "Phantom" would be better?

But also, if we can get rid of the associated type as I mentioned in one of the comments, maybe we won't need this at all.

air/src/air/logup_gkr.rs
```diff
@@ -81,9 +82,72 @@ pub trait LogUpGkrEvaluator: Clone + Sync {
...
+
 #[derive(Clone, Debug, PartialEq, PartialOrd, Eq, Ord)]
 pub enum LogUpGkrOracle<B: StarkField> {
+    // a column with a given index in the main trace segment
     CurrentRow(usize),
+    // a column with a given index in the main trace segment but shifted upwards
     NextRow(usize),
+    // a virtual periodic column defined by its values in a given cycle. Note that the cycle length
+    // must be a power of 2.
     PeriodicValue(Vec<B>),
```
irakliyk, line 152: nit: we should use doc comments here - i.e., `///`.

air/src/air/context.rs
```diff
@@ -232,12 +241,16 @@ impl<B: StarkField> AirContext<B> {
...
+        }
     }
 
-    /// Returns true if the auxiliary trace segment contains a Lagrange kernel column
-    pub fn has_lagrange_kernel_aux_column(&self) -> bool {
-        self.lagrange_kernel_aux_column_idx().is_some()
+    /// Returns true if LogUp-GKR is enabled.
+    pub fn uses_logup_gkr(&self) -> bool {
+        self.logup_gkr
     }
```
irakliyk, line 254: nit (feel free to ignore): maybe `logup_gkr_enabled()` would be a better name?

air/src/air/coefficients.rs
```diff
@@ -32,6 +32,7 @@ pub struct ConstraintCompositionCoefficients<E: FieldElement> {
     pub transition: Vec<E>,
     pub boundary: Vec<E>,
     pub lagrange: Option<LagrangeConstraintsCompositionCoefficients<E>>,
+    pub s_col: Option<E>,
```
irakliyk, line 35: Should we also update doc comments for this struct?

air/src/air/coefficients.rs
```diff
@@ -109,4 +110,6 @@ pub struct DeepCompositionCoefficients<E: FieldElement> {
     pub constraints: Vec<E>,
     /// Lagrange kernel trace polynomial composition coefficient $\gamma$.
     pub lagrange: Option<E>,
+    /// TODO
+    pub s_col: Option<E>,
```
irakliyk, line 114: Similar comment re doc-comments as above.

winterfell/src/lib.rs
```diff
@@ -594,15 +594,14 @@
 #[cfg(test)]
 extern crate std;
 
-pub use air::{AuxRandElements, GkrVerifier};
+pub use air::AuxRandElements;
```
irakliyk, line 597: Should we re-export `LogUpGkrEvaluator` from here?

verifier/src/lib.rs
```diff
@@ -341,6 +344,41 @@ where
         .map_err(VerifierError::FriVerificationFailed)
 }
 
+fn verify_gkr<E: FieldElement, H: ElementHasher<BaseField = E::BaseField>>(
```
irakliyk, line 347: I would move this into its own file - e.g., `src/logup_gkr.rs`.

irakliyk (commented):
Thank you! Looks good! Not a full review yet - but I left some comments/questions inline.

air/src/air/mod.rs
```diff
@@ -306,16 +305,14 @@ pub trait Air: Send + Sync {
...
-    /// Leave unimplemented if the `Air` doesn't use a GKR proof.
-    fn get_gkr_proof_verifier<E: FieldElement<BaseField = Self::BaseField>>(
+    /// Returns the object needed for the LogUp-GKR argument.
+    fn get_logup_gkr_evaluator<E: FieldElement<BaseField = Self::BaseField>>(
         &self,
-    ) -> Self::GkrVerifier {
-        unimplemented!("`get_auxiliary_proof_verifier()` must be implemented when the proof contains a GKR proof");
+    ) -> Self::LogUpGkrEvaluator {
+        unimplemented!("`get_logup_gkr_evaluator()` must be implemented when LogUp-GKR is enabled");
     }
```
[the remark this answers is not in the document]
Al-Kindi-0: Indeed, but should the signature then be something like:

```rust
fn get_logup_gkr_evaluator<E: FieldElement<BaseField = Self::BaseField>>(
        &self,
    ) -> impl LogUpGkrEvaluator<BaseField = Self::BaseField, PublicInputs = Self::PublicInputs> {
        DummyLogUpGkrEval::default()
    }
```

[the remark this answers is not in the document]
irakliyk: I'm hoping we can git rid of `DummyLogUpGkrEval` entirely. If `AirContext::uses_logup_gkr()` return false, this method should never be called and so panicking here should be fine. And if it returns true, then this should return a non-dummy evaluator.

Also, why do we need the type parameter `E`? It seems like specifying the base field is sufficient for the `LogUpGkrEvaluator` - no?

verifier/src/lib.rs
Al-Kindi-0: Moved!
I have also refactored it in order to mimic the prover but it seems that there is the logic for generating the randomness post-GKR which is duplicated in both the verifier and prover crates. The best solution would be to move this logic to the AIR crate but then we have to deconstruct `FinalOpeningClaim` before passing it.
Curious if you see a cleaner way.

winterfell/src/lib.rs
Al-Kindi-0: Done

air/src/air/coefficients.rs
Al-Kindi-0: Updated

air/src/air/coefficients.rs
Al-Kindi-0: Updated

air/src/air/context.rs
Al-Kindi-0: I like this name better, updated!

air/src/air/logup_gkr.rs
Al-Kindi-0: Changed

air/src/air/logup_gkr.rs
Al-Kindi-0: Changed to `Phantom`

air/src/air/trace_info.rs
Al-Kindi-0: Changed

air/src/air/trace_info.rs
Al-Kindi-0: I think it is easier to change `main_trace_width` to `main_segment_width` given the rest of the code-base. I think the rational was that there is only one execution trace but, potentially, two segments; a main and an auxiliary one.

air/src/air/aux.rs
```diff
@@ -65,70 +67,66 @@ impl<E> AuxRandElements<E> {
 /// the Lagrange kernel random elements. Those claims are (linearly) combined into one using the
 /// openings combining randomness.
 #[derive(Clone, Debug)]
-pub struct GkrRandElements<E> {
-    lagrange: LagrangeKernelRandElements<E>,
-    openings_combining_randomness: Vec<E>,
+pub struct GkrData<E: FieldElement> {
```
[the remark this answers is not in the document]
Al-Kindi-0: Updated! Thanks

sumcheck/Cargo.toml
```diff
@@ -40,7 +40,7 @@ air = { version = "0.9", path = "../air", package = "winter-air", default-featur
 crypto = { version = "0.9", path = "../crypto", package = "winter-crypto", default-features = false }
 math = { version = "0.9", path = "../math", package = "winter-math", default-features = false }
 utils = { version = "0.9", path = "../utils/core", package = "winter-utils", default-features = false }
-rayon = { version = "1.8", optional = true }
+rayon = { version = "1.8", optional = true}
```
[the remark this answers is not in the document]
Al-Kindi-0: Reverted

air/src/air/mod.rs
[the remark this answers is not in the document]
Al-Kindi-0: If we don't provide a default implementation in the trait then we would have to implement it each time we implement `Air`. The method returns an `impl LogUpGkrEvaluator<BaseField = Self::BaseField, PublicInputs = Self::PublicInputs>` and hence I am not how we get over the fact that `()` doesn't implement the trait.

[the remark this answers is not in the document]
Al-Kindi-0: > Also, why do we need the type parameter E? It seems like specifying the base field is sufficient for the LogUpGkrEvaluator - no?

Correct, it just is less cumbersome to write `::<E>` than `::<E::BaseField>`. Changed it in a separate commit though

prover/src/lib.rs
```diff
@@ -201,32 +201,49 @@ pub trait Prover {
...
+                &evaluator,
+                &lagrange_col,
+            );
+
+            aux_trace.merge_column(s_col);
+            aux_trace.merge_column(lagrange_col);
+        }
+
+        aux_trace
     }
```
irakliyk, line 247: Why do we need to have this wrapper? Could the following work?

- We have `build_aux_trace()` function (as we do now) and this function gets called when appropriate directly from  the `generate_proof()` function.
- We have a separate function called something like `build_gkr_trace_columns()` and this function either returns two columns or appends the columns to a passed in auxiliary trace.
  - I also wonder if maybe this function can be somehow combined with `prove_gkr()` function.

irakliyk (commented):
Looks good! Thank you! Still not a full review (though I think I covered 80% of the code by now) - but I left some more comments inline.

air/src/air/mod.rs
```diff
@@ -306,16 +299,15 @@ pub trait Air: Send + Sync {
...
-    fn get_gkr_proof_verifier<E: FieldElement<BaseField = Self::BaseField>>(
+    /// Returns the object needed for the LogUp-GKR argument.
+    fn get_logup_gkr_evaluator<B: StarkField>(
         &self,
-    ) -> Self::GkrVerifier {
-        unimplemented!("`get_auxiliary_proof_verifier()` must be implemented when the proof contains a GKR proof");
+    ) -> impl LogUpGkrEvaluator<BaseField = Self::BaseField, PublicInputs = Self::PublicInputs>
+    {
+        PhantomLogUpGkrEval::new()
     }
```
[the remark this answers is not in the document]
irakliyk: Ah - I guess we need to return something from here because we can't implement `LogUpGkrEvaluator` for `()` right?

verifier/src/lib.rs
irakliyk: It does feel like `generate_gkr_randomness()` should be a method of `LogUpGkrEvaluator`. And I guess the signature there could look something like this:
```rust
pub fn generate_gkr_randomness<E: FieldElement, H: ElementHasher<BaseField = E::BaseField>>(
    oracle_openings: Vec<E>,
    lagrange_kernel_evel_point: Vec<E>,
    public_coin: &mut impl RandomCoin<Hasher = H, BaseField = E::BaseField>,
) -> GkrData<E>
```

I think this look OK - though I think the name `generate_gkr_randomness()` feels a bit weird because: (1) not all GKR-related randomness is generated here (e.g., `lagrange_kernel_evel_point` is generated somewhere else) and (2) this returns more than just random values (e.g., `GkrData` includes oracles and their openings).

But I don't have a better suggestion yet.

air/src/air/mod.rs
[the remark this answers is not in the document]
Al-Kindi-0: Exactly!

air/src/air/aux.rs
```diff
@@ -52,83 +50,92 @@ impl<E> AuxRandElements<E> {
     pub fn gkr_openings_combining_randomness(&self) -> Option<&[E]> {
         self.gkr.as_ref().map(|gkr| gkr.openings_combining_randomness.as_ref())
     }
+
+    pub fn gkr_data(&self) -> Option<GkrData<E>> {
+        self.gkr.clone()
+    }
```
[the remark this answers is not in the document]
Al-Kindi-0: Added

air/src/air/aux.rs
```diff
@@ -52,83 +50,92 @@ impl<E> AuxRandElements<E> {
...
 }
 
-/// Holds all the random elements needed when using GKR to accelerate LogUp.
+/// Holds all the data needed when using LogUp-GKR in order to build the two extra auxiliary columns
+/// for running the univariate IOP for multi-linear evaluations of [1].
```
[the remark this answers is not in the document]
Al-Kindi-0: Updated

examples/Cargo.toml
```diff
@@ -26,6 +26,7 @@ default = ["std"]
 std = ["core-utils/std", "hex/std", "rand-utils", "winterfell/std"]
 
 [dependencies]
+air = { version = "0.9", path = "../air", package = "winter-air", default-features = false }
```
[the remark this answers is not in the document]
Al-Kindi-0: Removed

verifier/src/lib.rs
```diff
@@ -175,19 +183,28 @@ where
...
             let gkr_proof = {
-                let gkr_proof_serialized = channel
-                    .read_gkr_proof()
-                    .expect("Expected an a GKR proof because trace has lagrange kernel column");
+                let gkr_proof_serialized =
+                    channel.read_gkr_proof().expect("Expected a GKR proof but there was none");
 
-                Deserializable::read_from_bytes(gkr_proof_serialized)
+                GkrCircuitProof::read_from_bytes(gkr_proof_serialized)
                     .map_err(|err| VerifierError::ProofDeserializationError(err.to_string()))?
```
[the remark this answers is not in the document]
Al-Kindi-0: Fixed

prover/src/lib.rs
Al-Kindi-0: Implemented your suggestion

verifier/src/lib.rs
Al-Kindi-0: I have implemented the refactoring and came up with a better, although more verbose, name.

prover/src/lib.rs
```diff
@@ -333,8 +333,10 @@ pub trait Prover {
 
                 (None, AuxRandElements::new(rand_elements))
             };
+            let mut aux_trace = maybe_await!(self.build_aux_trace(&trace, &aux_rand_elements));
```
irakliyk, line 336: Is there a reason to build the non-LogUp-GKR part of the aux trace here rather than before we build the GKR proof (i.e., do this on line 302 above).

If there is no particular reason, then I think it would be a bit more natural to first deal with non-LogUp-GKR part, then generate the GKR proof, and then build the LogUp-GKR-related columns.

irakliyk (approved):
Looks good! Thank you! I left some more comments inline. Most of them are pretty minor and we can address them either here or in the subsequent PRs.

prover/src/lib.rs
Al-Kindi-0: Yes, I agree. Have changed it now and simplified a few things on the way

verifier/src/lib.rs
```diff
@@ -175,19 +179,24 @@ where
...
             let gkr_rand_elements = air
-                .get_gkr_proof_verifier::<E>()
-                .verify::<E, _>(gkr_proof, &mut public_coin)
-                .map_err(|err| VerifierError::GkrProofVerificationFailed(err.to_string()))?;
+                .get_logup_gkr_evaluator::<E::BaseField>()
+                .generate_univariate_iop_for_multi_linear_opening_data(
+                    openings,
+                    eval_point,
+                    &mut public_coin,
+                );
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

verifier/src/channel.rs
```diff
@@ -172,9 +174,12 @@ where
         self.pow_nonce
     }
 
-    /// Returns the serialized GKR proof, if any.
-    pub fn read_gkr_proof(&self) -> Option<&Vec<u8>> {
-        self.gkr_proof.as_ref()
+    /// Returns the GKR proof, if any.
+    pub fn read_gkr_proof(&self) -> Result<GkrCircuitProof<E>, VerifierError> {
+        GkrCircuitProof::read_from_bytes(
+            self.gkr_proof.as_ref().expect("Expected a GKR proof but there was none"),
```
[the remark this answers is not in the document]
Al-Kindi-0: If the deserialization fails then this should return an error but probably you are thinking about something else

air/src/air/logup_gkr.rs
```diff
@@ -79,11 +81,122 @@ pub trait LogUpGkrEvaluator: Clone + Sync {
...
+            self.get_oracles().to_vec(),
+        )
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct PhantomLogUpGkrEval<B: StarkField, P: Clone + Send + Sync + ToElements<B>> {
+    _field: PhantomData<B>,
+    _public_inputs: PhantomData<P>,
+}
```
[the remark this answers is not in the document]
Al-Kindi-0: changed

air/src/air/mod.rs
```diff
@@ -195,13 +194,7 @@ pub trait Air: Send + Sync {
...
     /// A type defining shape of public inputs for the computation described by this protocol.
     /// This could be any type as long as it can be serialized into a sequence of field elements.
-    type PublicInputs: ToElements<Self::BaseField> + Send;
-
-    /// An GKR proof object. If not needed, set to `()`.
-    type GkrProof: Serializable + Deserializable + Send;
-
-    /// A verifier for verifying GKR proofs. If not needed, set to `()`.
-    type GkrVerifier: GkrVerifier<GkrProof = Self::GkrProof>;
+    type PublicInputs: Clone + Sync + ToElements<Self::BaseField> + Send;
```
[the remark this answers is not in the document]
Al-Kindi-0: Changed

PR #304, opened 2024-08-23
Add support for periodic columns in LogUp-GKR

Al-Kindi-0:
Builds on #296


Al-Kindi-0:
Opened a new PR instead

PR #305, opened 2024-08-28
Update constraint degree checks

Al-Kindi-0:
This PR relaxes the constraint degree checks so that they become degree bounds checks instead of degree equality checks.

### DIFF
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -280,13 +280,6 @@ impl<B: StarkField> AirContext<B> {
             self.trace_len() / 2 + 1,
             n
         );
-        // make sure the composition polynomial can be computed correctly with the specified
-        // number of exemptions.
-        // The `ce_blowup` factor puts a ceiling on the maximal degree of a constraint composition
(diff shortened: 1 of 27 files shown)


air/src/air/context.rs
```diff
@@ -325,26 +280,6 @@ impl<B: StarkField> AirContext<B> {
             self.trace_len() / 2 + 1,
             n
         );
```
Al-Kindi-0, line 282: exemption points should probably be removed as well (i.e., fixed to always be 1)

Al-Kindi-0:
I have removed now constraint degree from the code base. I personally don't like the result as it is not straightforward to know if the max degree of the (constraints of the) AIR is chosen correctly. Moreover, even in debug mode, I am not sure we can efficiently test this and hence the only way to know that this is the case is to wait for verification to not pass.
Thus I am not in favor of removing the constraint degrees unless we solve the above.

Al-Kindi-0:
Closing this in favor of #311 since it seems that having degree constraints is better than not having them.

PR #307, opened 2024-09-03
Add support for periodic columns in LogUp-GKR

Al-Kindi-0:


air/src/air/logup_gkr/mod.rs
```diff
@@ -229,3 +250,76 @@ pub enum LogUpGkrOracle<B: StarkField> {
...
+        self.table.len()
+    }
+
+    pub fn table(&self) -> &[Vec<E>] {
+        &self.table
+    }
+
+    pub fn get_periodic_values_at(&self, row: usize) -> Vec<E> {
+        self.table.iter().map(|col| col[row % col.len()]).collect()
+    }
```
irakliyk, line 300: I think this is used during constraint evaluation in the prover. If so, recurring a vector here may be expensive. Maybe we should return the values by writing them into a passed in mutable buffer?

air/src/air/logup_gkr/mod.rs
```diff
@@ -229,3 +250,76 @@ pub enum LogUpGkrOracle<B: StarkField> {
...
+    where
+        E: FieldElement + ExtensionOf<B>,
+        B: StarkField,
+    {
+        let mut result = vec![];
+        for col in table.iter() {
+            let mut res = vec![];
+            for v in col {
+                res.push(E::from(*v))
+            }
```
irakliyk, line 283: I think this can be replaced with `embed_extension()` call.

sumcheck/src/verifier/mod.rs
```diff
@@ -147,3 +149,23 @@ pub enum SumCheckVerifierError {
...
 }
+
+// HELPER
+// =================================================================================================
+
+/// Evaluates periodic columns as multi-linear extensions.
+fn evaluate_periodic_columns_at<E: FieldElement>(
+    periodic_columns: PeriodicTable<E>,
+    eval_point: &[E],
+) -> Vec<E> {
```
irakliyk, line 160: Should this be a method on the `PeriodicTable`?

sumcheck/src/verifier/mod.rs
```diff
@@ -86,12 +86,14 @@ pub fn verify_sum_check_input_layer<E: FieldElement, H: ElementHasher<BaseField
...
-        &mut numerators,
-        &mut denominators,
-    );
+
+    let periodic_columns = evaluator.build_periodic_values::<E, E>();
+    let periodic_columns_evaluations =
+        evaluate_periodic_columns_at(periodic_columns, &proof.0.openings_claim.eval_point);
+
+    let mut query = proof.0.openings_claim.openings.clone();
+    query.extend_from_slice(&periodic_columns_evaluations);
```
irakliyk, line 95: Here we are assuming that periodic column values are always at the end of the query, right? But when we define oracles, I don't think this is enforced. Could this cause issues?

sumcheck/src/prover/high_degree.rs
```diff
@@ -398,6 +433,21 @@ fn sumcheck_round<E: FieldElement>(
...
+                evals_zero
+                    .iter_mut()
+                    .skip(num_mls)
+                    .enumerate()
+                    .for_each(|(i, ev)| *ev = periodic_at_zero[i]);
+                evals_one
+                    .iter_mut()
+                    .skip(num_mls)
+                    .enumerate()
+                    .for_each(|(i, ev)| *ev = periodic_at_one[i]);
```
irakliyk, line 449: If we change `get_periodic_values_at()` to write to a buffer, we may be able to do this as:
```rust
periodic_table.get_periodic_values_at(2 * i, &mut evals_zero[num_mls + 1..]);
periodic_table.get_periodic_values_at(2 * i + 1, &mut evals_one[num_mls + 1..]);
```

sumcheck/src/prover/high_degree.rs
```diff
@@ -311,6 +331,21 @@ fn sumcheck_round<E: FieldElement>(
...
```
irakliyk, line 347: Same comment as above.

irakliyk (commented):
Thank you! Looks good! I left some comments inline - the main one is about assuming that periodic oracles being always at the end of the list of oracles. But not sure if that's a big issue.

sumcheck/src/verifier/mod.rs
Al-Kindi-0: Indeed we are, and this will break things if that is not the case. Should we add checks and panic throughout to enforce this?

sumcheck/Cargo.toml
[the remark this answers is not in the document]
Al-Kindi-0: Done

winterfell/src/tests/logup_gkr_simple.rs
```diff
@@ -1,8 +1,3 @@
-// Copyright (c) Facebook, Inc. and its affiliates.
-//
-// This source code is licensed under the MIT license found in the
-// LICENSE file in the root directory of this source tree.
-
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

winterfell/src/tests/mod.rs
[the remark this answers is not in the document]
Al-Kindi-0: Done

air/src/air/logup_gkr/mod.rs
Al-Kindi-0: Makes sense, implemented

air/src/air/logup_gkr/mod.rs
Al-Kindi-0: Yes, fixed

sumcheck/src/verifier/mod.rs
Al-Kindi-0: That was the initial plan but then the method depends on MultiLinear which is part of the sum-check crate. The later we can import from the air crate

sumcheck/src/prover/high_degree.rs
Al-Kindi-0: Done

sumcheck/src/prover/high_degree.rs
Al-Kindi-0: Done

winterfell/src/tests/logup_gkr_periodic.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        for i in 0..trace_len / 8 {
+            values_2[8 * i] = BaseElement::from(8 * i as u32);
+        }
+
+        let mut periodic: Vec<BaseElement> = (0..trace_len).map(|_idx| BaseElement::ZERO).collect();
+
+        for i in 0..trace_len / 8 {
+            periodic[8 * i] = BaseElement::ONE;
+        }
```
plafer, line 82: `periodic` is never used. Was this intended?

sumcheck/src/prover/high_degree.rs
```diff
@@ -311,6 +331,10 @@ fn sumcheck_round<E: FieldElement>(
                 let eq_at_zero = eq_ml.evaluations()[2 * i];
                 let eq_at_one = eq_ml.evaluations()[2 * i + 1];
 
+                // add evaluation of periodic columns
+                periodic_table.get_periodic_values_at(2 * i, &mut evals_zero[num_mls..]);
+                periodic_table.get_periodic_values_at(2 * i + 1, &mut evals_one[num_mls..]);
```
plafer, line 336: I don't think we can assume that periodic values will always be at the end of a `query`, right?

prover/src/logup_gkr/mod.rs
```diff
@@ -109,16 +109,21 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
...
+        let mut periodic_values_row = vec![E::BaseField::ZERO; periodic_values.num_columns()];
         let mut numerators = vec![E::ZERO; num_fractions];
         let mut denominators = vec![E::ZERO; num_fractions];
         for i in 0..main_trace.main_segment().num_rows() {
             let wires_from_trace_row = {
                 main_trace.read_main_frame(i, &mut main_frame);
+                periodic_values.get_periodic_values_at(i, &mut periodic_values_row);
+                evaluator.build_query(&main_frame, &periodic_values_row, &mut query);
 
                 evaluator.build_query(&main_frame, &[], &mut query);
```
plafer, line 128: We should remove that second call to `build_query()`, right?

sumcheck/src/verifier/mod.rs
plafer: If we always expect it to be at the end of the `query` buffer, can't we simply not pass it to `build_query()`, and append it ourselves?

IIUC, there is no other correct way to do it, so there's no point in asking (and hoping for) the user to do it.

plafer (commented):
Looks good! Left a few nits.

air/src/air/logup_gkr/mod.rs
```diff
@@ -229,3 +250,72 @@ pub enum LogUpGkrOracle<B: StarkField> {
...
+    pub fn table(&self) -> &[Vec<E>] {
+        &self.table
+    }
+
+    pub fn get_periodic_values_at(&self, row: usize, values: &mut [E]) {
+        self.table
+            .iter()
+            .zip(values.iter_mut())
+            .for_each(|(col, value)| *value = col[row % col.len()])
+    }
```
irakliyk, line 296: nit: let's rename this to `fill_periodic_values_at()`.

sumcheck/Cargo.toml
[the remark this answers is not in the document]
irakliyk: Seems like the changes are still there?

sumcheck/src/verifier/mod.rs
irakliyk: Agreed. I think we should try to derive oracles for periodic columns so that the user does not need to make sure that oracles are ordered correctly.

Also, maybe we define query something like this:
```rust
pub struct Query<E: FieldElement> {
    pub committed_values: Vec<E>,
    pub periodic_value: Vec<E>,
}
```
This way, there is no ambiguity about where the periodic values should go (though, it does require 2 vectors instead of 1).

Al-Kindi-0: > If we always expect it to be at the end of the `query` buffer, can't we simply not pass it to `build_query()`, and append it ourselves?
> 

I am not sure how `build_query()` fits into the code snippet above. `build_query()` is specifically for building from the main trace segment.

sumcheck/Cargo.toml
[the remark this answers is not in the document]
Al-Kindi-0: Not sure why it didn't revert, now it should be fine

sumcheck/src/verifier/mod.rs
Al-Kindi-0: > Agreed. I think we should try to derive oracles for periodic columns so that the user does not need to make sure that oracles are ordered correctly.
> 

I am not sure I understand fully but the way I see solving the issue raised is by separating the the oracles into committed and periodic. 
My assumption was that, if we are going to test the consistency between `get_oracles()` and `build_query()` anyway then we might as well impose that periodic columns are at the end. The test will check for this any and imposing that periodic columns are towards the end is implicit in requiring consistency between `get_oracles()` and `build_query()`.

air/src/air/logup_gkr/mod.rs
```diff
@@ -145,6 +145,27 @@ pub trait LogUpGkrEvaluator: Clone + Sync {
...
+
+        let oracles = self.get_oracles();
+
+        for oracle in oracles {
+            if let LogUpGkrOracle::PeriodicValue(values) = oracle {
+                let values = embed_in_extension(values.to_vec());
+                table.push(values)
+            }
+        }
+        PeriodicTable { table }
```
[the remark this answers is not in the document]
Al-Kindi-0: Thanks for the suggestion, switched to it
Indeed, the generic is redundant now, it is a leftover from a previous iteration. Removed now

air/src/air/logup_gkr/mod.rs
```diff
@@ -229,3 +250,72 @@ pub enum LogUpGkrOracle<B: StarkField> {
...
+}
+
+impl<E> PeriodicTable<E>
+where
+    E: FieldElement,
+{
+    pub fn new(table: Vec<Vec<E::BaseField>>) -> Self {
+        let mut result = vec![];
+        for col in table.iter() {
+            let res = embed_in_extension(col.to_vec());
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

prover/src/logup_gkr/prover.rs
```diff
@@ -97,6 +104,7 @@ fn prove_input_layer<
     evaluator: &impl LogUpGkrEvaluator<BaseField = E::BaseField>,
     log_up_randomness: Vec<E>,
     multi_linear_ext_polys: Vec<MultiLinearPoly<E>>,
+    periodic_table: &mut PeriodicTable<E>,
     claim: GkrClaim<E>,
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

sumcheck/src/prover/high_degree.rs
```diff
@@ -160,6 +160,7 @@ pub fn sum_check_prove_higher_degree<
     r_sum_check: E,
     log_up_randomness: Vec<E>,
     mut mls: Vec<MultiLinearPoly<E>>,
+    periodic_table: &mut PeriodicTable<E>,
     coin: &mut impl RandomCoin<Hasher = H, BaseField = E::BaseField>,
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

winterfell/src/tests/logup_gkr_periodic.rs
```diff
@@ -0,0 +1,364 @@
...
+}
+
+impl LogUpGkrPeriodic {
+    fn new(trace_len: usize, aux_segment_width: usize) -> Self {
+        assert!(trace_len < u32::MAX.try_into().unwrap());
+
+        let table: Vec<BaseElement> =
+            (0..trace_len).map(|idx| BaseElement::from(idx as u32)).collect();
+        let mut multiplicity: Vec<BaseElement> =
+            (0..trace_len).map(|_idx| BaseElement::ZERO).collect();
```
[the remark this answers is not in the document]
Al-Kindi-0: Done

winterfell/src/tests/logup_gkr_periodic.rs
Al-Kindi-0: Yep, it was a leftover from some consistency checks I had before. Removed

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Yep, remnants from the merge conflicts

air/src/air/logup_gkr/mod.rs
Al-Kindi-0: Changed

sumcheck/src/verifier/mod.rs
plafer: > I am not sure how build_query() fits into the code snippet above. build_query() is specifically for building from the main trace segment.

We are currently making an assumption about the `query` argument to `evaluate_query()` - namely that the periodic columns need to be appended at the end of the buffer. Hence, this assumption needs to be upheld by `build_query()` as well.

My point was: why give the user the possibly to screw it up and place the periodic columns somewhere other than at the end of the `query` buffer? We could instead change `build_query()` to be

```rs
    fn build_query<E>(&self, frame: &EvaluationFrame<E>, query: &mut [E])
    where
    { ... }
```

and append periodic values manually after each call to `build_query()`, for example [here](https://github.com/al-kindi-0/winterfell/blob/873345af6777668f88e245bcabcb995ccc28585c/prover/src/logup_gkr/mod.rs#L125-L126):

```rs
evaluator.build_query(&main_frame, &mut query);
query.extend(periodic_values_row);
```

i.e. we are enforcing that assumption ourselves, and not leaving the user any possibility for mistake.

IIUC, @irakliyk built on this and suggested we make the structure explicit by representing the periodic values separately (instead of appended at the end up the buffer). So the previous example would look like

```rs
evaluator.build_query(&main_frame, &mut query);
let query_full = Query::new(query, periodic_values_row);
```

Does that make sense? Maybe I'm missing something too.

Al-Kindi-0: Yep, it does make sense. One thing that is different with respect to the new proposal is with respect to new allocations. In the current way, we just write to a buffer but with the current proposal I am not sure if we are incurring some kind of overhead when creating `Query`.

irakliyk: I think `Query::new()` should be fine here - i.e., it would not allocate anything as it would leave on the stack.

irakliyk (commented):
Overall, looks good! Thank you! I do wonder if we should change things slight based on the discussion in https://github.com/facebook/winterfell/pull/307#discussion_r1742783220. Basically, it seems like adding periodic values to the list of other oracles imposes some implicit assumptions. So, maybe we should just treat them separately (and we kind of do in a few places already).

One way to do this could be as follows:

First, we remove `PeriodicValue` variant from the `LogUpGkrOracle` enum.

Then, we add a dedicated method to the `LogUpGkrEvaluator` to get periodic columns. Something like:
```rust
pub trait LogUpGkrEvaluator {
    fn get_periodic_column_values(&self) -> Vec<Vec<Self::BaseField>>;
}
```

Then, we restrict `LogUpGkrEvaluator::build_query()` to work only with committed columns. So, it would look something like this:
```rust
pub trait LogUpGkrEvaluator {
    fn build_query<E>(&self, frame: &EvaluationFrame<E>, query: &mut [E])
    where
        E: FieldElement<BaseField = Self::BaseField>;
}
```
Finally, we pass periodic values to the `LogUpGkrEvaluator::evaluate_query()` explicitly like so:
```rust
pub trait LogUpGkrEvaluator {
    fn evaluate_query<F, E>(
        &self,
        query: &[F],
        periodic_values: &[F],
        logup_randomness: &[E],
        numerators: &mut [E],
        denominators: &mut [E],
    ) where
        F: FieldElement<BaseField = Self::BaseField>,
        E: FieldElement<BaseField = Self::BaseField> + ExtensionOf<F>;
}
```

This way, we don't make any implicit assumptions about where periodic values go in the `query` - they are always treated separately.

Lastly, the implementation of `LogUpGkrEvaluator ::build_periodic_values()` would be simplified as well.

irakliyk:
One other question: with the suggestions I made above, it looks like handling of periodic values in `LogUpGkrEvaluator` is almost the same as in the main `Air` trait. Should we just keep the periodic columns defined in the `Air` trait? (i.e., basically the user defines all periodic column in the `Air` trait and they can be used both for regular constraints and for LogUp-GKR.

I guess the answer to this depends on how much overhead this will introduce as we may end up with some periodic values passed to `LogUpGkrEvaluator` which it doesn't really need.

irakliyk (approved):
Looks good! Thank you! I left one small comment inline (but we can also merge w/o it and fix it later).

sumcheck/src/prover/high_degree.rs
```diff
@@ -280,21 +298,28 @@ fn sumcheck_round<E: FieldElement>(
     evaluator: &impl LogUpGkrEvaluator<BaseField = <E as FieldElement>::BaseField>,
     eq_ml: &MultiLinearPoly<E>,
     mls: &[MultiLinearPoly<E>],
+    periodic_table: &mut PeriodicTable<E>,
```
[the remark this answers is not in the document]
Al-Kindi-0: Good point! Fixed

PR #309, opened 2024-09-06
fix: incorrect read_slice impl for ReadAdapter

### LINKED ISSUE #308: ReadAdapter is broken
It appears that the implementation of ReadAdapter, at least when applied to a `File`, is broken. I haven't yet looked to pin down the specific problem, but it appears to me that we may have an off-by-one error of some kind, as we get a panic attempting to read beyond the end of a buffer by 1 byte (at least in the case I have where this comes up).

I'll try to take a look at this tomorrow, but wanted to at least get this bug on our radar, as it currently affects various bits of Miden functionality that builds on `winter-utils`.

bitwalker:
This commit fixes a bug that is present in the implementation of `ByteReader::read_slice` for ReadAdapter. Specifically, it does not consume any input, so calling `read_slice` and then attempting to read the next value in the input, will read the same bytes again. The documented behavior of this function is that any of the `read_*` methods consume the corresponding amount of input.

To catch this, and to prevent future regressions, a new test was added that serializes some data to a file using one approach, and deserializes it using the ReadAdapter. This ensures that we don't accidentally make choices in the writer that aren't matched by the reader, and vice versa.

Closes #308

### DIFF
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -230,9 +230,11 @@ impl Deserializable for OodFrame {
 // ================================================================================================
 
 /// Stores the trace evaluations at `z` and `gz`, where `z` is a random Field element in
-/// `current_row` and `next_row`, respectively. If the Air contains a Lagrange kernel auxiliary
-/// column, then that column interpolated polynomial will be evaluated at `z`, `gz`, `g^2 z`, ...
-/// `g^(2^(v-1)) z`, where `v == log(trace_len)`, and stored in `lagrange_kernel_frame`.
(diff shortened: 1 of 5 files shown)


bitwalker:
@irakliyk This fixes #308, and also addresses the issue in Miden when deserializing `Library` or `Program` from files.

irakliyk (approved):
Looks good! Thank you for fixing this!

Since this is a non-breaking change, could you change the base to `main`? This way, I could make a patch release which would fix things immediately downstream.

bitwalker:
@irakliyk Done!

PR #311, opened 2024-09-11
Update constraint degree checks

Al-Kindi-0:
Same as #305 

This PR relaxes the constraint degree checks so that they become degree bounds checks instead of degree equality checks.


irakliyk (approved):
Looks good! Thank you!

This was only an issue during proof generation, right? During verification the moved away from strict equality checks a while back, I think.

Al-Kindi-0:
> Looks good! Thank you!
> 
> This was only an issue during proof generation, right? During verification the moved away from strict equality checks a while back, I think.

That is indeed the case.

PR #314, opened 2024-09-12
Cleanup AirContext

Al-Kindi-0:
Proposes a minor cleanup of AirContext methods

### DIFF
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -106,7 +106,7 @@ impl<B: StarkField, P> AirContext<B, P> {
         );
         assert!(num_main_assertions > 0, "at least one assertion must be specified");
 
-        if trace_info.is_multi_segment() && !trace_info.logup_gkr_enabled() {
+        if !trace_info.logup_gkr_enabled() {
             assert!(
              !aux_transition_constraint_degrees.is_empty(),
             "at least one transition constraint degree must be specified for the auxiliary trace segment"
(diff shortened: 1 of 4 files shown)


air/src/air/context.rs
```diff
@@ -115,15 +115,6 @@ impl<B: StarkField, P> AirContext<B, P> {
...
             );
-        } else {
-            assert!(
-                aux_transition_constraint_degrees.is_empty(),
-                "auxiliary transition constraint degrees specified for a single-segment trace"
-            );
-            assert!(
-                num_aux_assertions == 0,
-                "auxiliary assertions specified for a single-segment trace"
-            );
```
irakliyk, line 126: I think we still may need this check - just under a slightly different set of conditions. Basically, it could be like so:
```rust
if trace_info.is_multi_segment() {
    if !trace_info.logup_gkr_enabled() {
        // make sure at least one aux constraint/assertion is specified
    }
} else {
    // make sure there are no aux constraints/assertions specified
}
```

Al-Kindi-0: Indeed, fixed it now

air/src/air/context.rs
```diff
@@ -163,32 +166,10 @@ impl<B: StarkField, P> AirContext<B, P> {
             trace_domain_generator: B::get_root_of_unity(trace_length.ilog2()),
             lde_domain_generator: B::get_root_of_unity(lde_domain_size.ilog2()),
             num_transition_exemptions: 1,
-            logup_gkr: false,
+            logup_gkr,
```
irakliyk, line 169: One more question: do we actually need `logup_gkr` field in `AirContext`. Seems like we should be able to get it from `trace_info` which is available in `AirContext` already.

Al-Kindi-0: Agreed, removed it now

irakliyk (approved):
All looks good! Thank you!

PR #315, opened 2024-09-13
Add LogUp-GKR benchmark

Al-Kindi-0:
Adds a simple end-to-end LogUp-GKR benchmark. Simplifies some of the benchmarks in the `sum-check` crate.

### DIFF
--- a/prover/benches/logup_gkr.rs
+++ b/prover/benches/logup_gkr.rs
@@ -20,16 +20,16 @@ use winter_prover::{
     DefaultConstraintEvaluator, DefaultTraceLde, Prover, StarkDomain, Trace, TracePolyTable,
 };
 
-const TRACE_LENS: [usize; 3] = [2_usize.pow(16), 2_usize.pow(18), 2_usize.pow(20)];
+const TRACE_LENS: [usize; 2] = [2_usize.pow(20), 2_usize.pow(21)];
 const AUX_TRACE_WIDTH: usize = 2;
 
 fn prove_with_lagrange_kernel(c: &mut Criterion) {
-    let mut group = c.benchmark_group("Lagrange kernel column");
+    let mut group = c.benchmark_group("prove with Lagrange kernel column");
     group.sample_size(10);
     group.measurement_time(Duration::from_secs(20));
 
     for &trace_len in TRACE_LENS.iter() {
-        group.bench_function(BenchmarkId::new("prover", trace_len), |b| {
+        group.bench_function(BenchmarkId::new("", trace_len), |b| {
             let trace = LogUpGkrSimple::new(trace_len, AUX_TRACE_WIDTH);
             let prover = LogUpGkrSimpleProver::new(AUX_TRACE_WIDTH);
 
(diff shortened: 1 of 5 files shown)


prover/benches/logup_gkr.rs
```diff
@@ -0,0 +1,351 @@
...
+}
+
+criterion_group!(lagrange_kernel_group, prove_with_lagrange_kernel);
+criterion_main!(lagrange_kernel_group);
+
+// LogUpGkrSimple
+// =================================================================================================
+
+#[derive(Clone, Debug)]
+struct LogUpGkrSimple {
```
irakliyk, line 52: nit: seems like this is a mock execution trace, right? If so, maybe it should be called `LogUpGkrSimpleTrace`?

prover/benches/logup_gkr.rs
```diff
@@ -0,0 +1,351 @@
...
+
+        for i in 0..4 {
+            values_1[i + 4] = BaseElement::ONE;
+        }
+
+        let mut values_2: Vec<BaseElement> = (0..trace_len).map(|_idx| BaseElement::ZERO).collect();
+
+        for i in 0..4 {
+            values_2[i + 4] = BaseElement::ONE;
+        }
```
irakliyk, line 85: Could we add a short explanation of what is happening below? (i.e., what do the 5 columns in the table are meant to contain).

prover/benches/logup_gkr.rs
```diff
@@ -0,0 +1,351 @@
...
+    crypto::{hashers::Blake3_256, DefaultRandomCoin},
+    math::{fields::f64::BaseElement, ExtensionOf, FieldElement},
+    matrix::ColMatrix,
+    DefaultConstraintEvaluator, DefaultTraceLde, Prover, StarkDomain, Trace, TracePolyTable,
+};
+
+const TRACE_LENS: [usize; 2] = [2_usize.pow(20), 2_usize.pow(21)];
+const AUX_TRACE_WIDTH: usize = 2;
+
+fn prove_with_lagrange_kernel(c: &mut Criterion) {
```
irakliyk, line 26: Could we add comments to this function to describe what it is primarily benchmarking? I believe it is the full end-to-end proving process for a trace table which has 5 columns and some very simple constraints. But it does use LogUp-GKR (though with a relatively simple input layer).

irakliyk (approved):
Looks good! Thank you! I left a few doc-related comments inline.

prover/benches/logup_gkr.rs
Al-Kindi-0: Changed

prover/benches/logup_gkr.rs
Al-Kindi-0: Added

prover/benches/logup_gkr.rs
Al-Kindi-0: Added

irakliyk (approved):
All looks good! Thank you!

PR #316, opened 2024-09-13
Parallelize Lagrange constraints evaluation

Al-Kindi-0:
Improves Lagrange constraints evaluation. This improves the end-to-end LogUp-GKR benchmark by 17-18% on my machine.

### DIFF
--- a/prover/src/constraints/evaluator/logup_gkr.rs
+++ b/prover/src/constraints/evaluator/logup_gkr.rs
@@ -10,6 +10,8 @@ use air::{
     LagrangeKernelConstraints, LagrangeKernelEvaluationFrame, LogUpGkrEvaluator,
 };
 use math::{batch_inversion, FieldElement};
+#[cfg(feature = "concurrent")]
+pub use utils::rayon::prelude::*;
 
 use crate::{StarkDomain, TraceLde};
 
@@ -68,8 +70,6 @@ where
         );
         let boundary_divisors_inv = self.compute_boundary_divisors_inv(domain);
 
-        let mut lagrange_frame = LagrangeKernelEvaluationFrame::new_empty();
-
         let evaluator = self.air.get_logup_gkr_evaluator();
         let s_col_constraint_divisor = compute_s_col_divisor::<E>(domain, self.air.trace_length());
         let s_col_idx = trace.trace_info().s_column_idx().expect("S-column should be present");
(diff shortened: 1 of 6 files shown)


irakliyk:
Thank you! Not a review yet, but a couple of questions:

> Improves Lagrange constraints evaluation by 17-18% on my machine.

What can I ran to test the effect on my machine?

Also, this seems quite a bit less than what we get by parallelizing general constraint evaluation (I think it is something like 4x or 5x on your machine). Are there big parts still left to address for Lagrange kernel constraints? Or is there something inherently different here?

Al-Kindi-0:
The command I used are:

1. `cargo bench --bench logup_gkr`
2. `cargo bench --features concurrent --bench logup_gkr`

The improvement is for the overall benchmark, which is end-to-end. As constraint evaluation is a small percentage, I think the improvement to Lagrange constraint evaluation is probably on the order you gave.

prover/src/constraints/evaluator/logup_gkr.rs
```diff
@@ -68,90 +70,184 @@ where
...
-                (rhs - lhs) * self.s_col_composition_coefficient.mul_base(divisor_at_step)
-            };
-
-            combined_evaluations_acc[step] +=
-                lagrange_combined_evaluations + s_col_combined_evaluation;
+        #[cfg(feature = "concurrent")]
+        combined_evaluations_acc
+            .par_iter_mut()
+            .enumerate()
+            .fold(
```
irakliyk, line 88: One potential issue with this approach is poor memory locality resulting of basically random assignment of data to various threads. An alternative would be to do something similar to what we with regular constraint evaluation:

1. First, we split up the trace into adjacent fragments. The number of fragments would be equal to the number of threads.
2. Then we evaluate each fragment in a different thread (e.g., see [here](https://github.com/facebook/winterfell/blob/main/prover/src/constraints/evaluator/default.rs#L105)).

I'm not 100% sure this will yield a significant improvement, but given that we are seeing only 1.5x overall improvement, I think this may be worth a try.

prover/src/constraints/evaluator/logup_gkr.rs
```diff
@@ -68,90 +70,184 @@ where
...
+                        vec![E::BaseField::ZERO; evaluator.get_oracles().len()],
+                    )
+                },
+                |(mut lagrange_frame, mut main_frame, mut aux_frame, mut query), (step, acc)| {
+                    // compute Lagrange kernel frame
+                    trace.read_lagrange_kernel_frame_into(
+                        step << lde_shift,
+                        l_col_idx,
+                        &mut lagrange_frame,
+                    );
```
irakliyk, line 103: One potential thing to check is whether using `Vec::push()` inside this method is slowing things down for some reason. It shouldn't but also we usually use direct element assignment for things like this.

Al-Kindi-0:
Subsumed by #317

Issue #306, opened 2024-09-02
Move s-column constraint evaluation logic into the regular constraint evaluator


Al-Kindi-0:
Currently during constraint evaluation and when LogUp-GKR is enabled, we are reading both main and aux frames twice from the trace LDE: the first time during regular constraint evaluation and the second time during S-column constraint evaluation.

An alternative approach could be to move S column evaluation logic into the regular constraint evaluator.

_Originally posted by @irakliyk in https://github.com/facebook/winterfell/pull/297#discussion_r1740408424_

Al-Kindi-0:
Done as part of #317

PR #317, opened 2024-09-16
Parallelize Lagrange constraints evaluation

Al-Kindi-0:
Builds on #315 and supersedes #316 .
The PR proposes a new constraint evaluator that is specific to when LogUp-GKR is enabled.
This has also the advantage of closing #306 .

### DIFF
--- a/air/src/air/logup_gkr/lagrange/frame.rs
+++ b/air/src/air/logup_gkr/lagrange/frame.rs
@@ -29,10 +29,14 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
         Self { frame }
     }
 
+    pub fn inner_mut(&mut self) -> &mut [E] {
+        &mut self.frame
+    }
+
     /// Constructs an empty Lagrange kernel evaluation frame from the raw column polynomial
     /// evaluations. The frame can subsequently be filled using [`Self::frame_mut`].
-    pub fn new_empty() -> Self {
-        Self { frame: Vec::new() }
+    pub fn new_empty(size: usize) -> Self {
+        Self { frame: vec![E::ZERO; size] }
     }
 
     /// Constructs the frame from the Lagrange kernel column trace polynomial coefficients for an
--- a/air/src/air/logup_gkr/lagrange/frame.rs
+++ b/air/src/air/logup_gkr/lagrange/frame.rs
@@ -26,13 +26,13 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
     // --------------------------------------------------------------------------------------------
 
     /// Constructs a Lagrange kernel evaluation frame from the raw column polynomial evaluations.
-    pub fn new(frame: Vec<E>) -> Self {
+    pub fn with_values(frame: Vec<E>) -> Self {
         Self { frame }
     }
 
     /// Constructs an empty Lagrange kernel evaluation frame from the raw column polynomial
     /// evaluations. The frame can subsequently be filled using [`Self::frame_mut`].
-    pub fn new_empty(trace_len: usize) -> Self {
+    pub fn new(trace_len: usize) -> Self {
         let frame_length = trace_len.ilog2() as usize + 1;
         Self { frame: vec![E::ZERO; frame_length] }
     }
--- a/air/src/air/logup_gkr/lagrange/frame.rs
+++ b/air/src/air/logup_gkr/lagrange/frame.rs
@@ -4,6 +4,7 @@
 // LICENSE file in the root directory of this source tree.
 
 use alloc::vec::Vec;
+use core::ops::{Index, IndexMut};
 
 use math::{polynom, FieldElement, StarkField};
 
@@ -29,14 +30,11 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
         Self { frame }
     }
 
-    pub fn inner_mut(&mut self) -> &mut [E] {
-        &mut self.frame
-    }
-
     /// Constructs an empty Lagrange kernel evaluation frame from the raw column polynomial
     /// evaluations. The frame can subsequently be filled using [`Self::frame_mut`].
-    pub fn new_empty(size: usize) -> Self {
-        Self { frame: vec![E::ZERO; size] }
+    pub fn new_empty(trace_len: usize) -> Self {
+        let frame_length = trace_len.ilog2() as usize + 1;
+        Self { frame: vec![E::ZERO; frame_length] }
     }
 
     /// Constructs the frame from the Lagrange kernel column trace polynomial coefficients for an
@@ -65,14 +63,6 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
         Self { frame }
     }
 
-    // MUTATORS
-    // --------------------------------------------------------------------------------------------
-
-    /// Returns a mutable reference to the inner frame.
-    pub fn frame_mut(&mut self) -> &mut Vec<E> {
-        &mut self.frame
-    }
-
     // ACCESSORS
     // --------------------------------------------------------------------------------------------
 
@@ -86,5 +76,19 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
     /// This is equal to `log(trace_length) + 1`.
     pub fn num_rows(&self) -> usize {
         self.frame.len()
+    }
+}
+
+impl<E: FieldElement> Index<usize> for LagrangeKernelEvaluationFrame<E> {
+    type Output = E;
+
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.frame[index]
+    }
+}
+
+impl<E: FieldElement> IndexMut<usize> for LagrangeKernelEvaluationFrame<E> {
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.frame[index]
     }
 }
--- a/prover/src/constraints/evaluator/logup_gkr.rs
+++ b/prover/src/constraints/evaluator/logup_gkr.rs
@@ -6,19 +6,18 @@
 use alloc::vec::Vec;
 
 use air::{
-    Air, EvaluationFrame, GkrData, LagrangeConstraintsCompositionCoefficients,
-    LagrangeKernelConstraints, LagrangeKernelEvaluationFrame, LogUpGkrEvaluator,
+    Air, GkrData, LagrangeConstraintsCompositionCoefficients,
+    LagrangeKernelConstraints, LogUpGkrEvaluator,
 };
 use math::{batch_inversion, FieldElement};
 #[cfg(feature = "concurrent")]
 pub use utils::rayon::prelude::*;
 
-use crate::{StarkDomain, TraceLde};
+use crate::StarkDomain;
 
 /// Contains a specific strategy for evaluating the Lagrange kernel and s-column boundary and
 /// transition constraints.
-pub struct LogUpGkrConstraintsEvaluator<'a, E: FieldElement, A: Air<BaseField = E::BaseField>> {
-    air: &'a A,
+pub struct LogUpGkrConstraintsEvaluator<E: FieldElement> {
     pub(crate) lagrange_kernel_constraints: LagrangeKernelConstraints<E>,
     pub(crate) gkr_data: GkrData<E>,
     pub(crate) s_col_composition_coefficient: E,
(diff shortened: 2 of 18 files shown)


prover/src/constraints/evaluator/logup_gkr_evaluator.rs
```diff
@@ -0,0 +1,461 @@
...
+        evaluation
+    }
+
+    /// Computes the transition and boundary constraints for the Lagrange kernel.
+    fn evaluate_lagrange_transition(
+        &self,
+        lagrange_frame: &LagrangeKernelEvaluationFrame<E>,
+        step: usize,
+        trans_constraints_divisors: &LagrangeKernelTransitionConstraintsDivisor<E>,
+        boundary_divisors_inv: &[E],
```
irakliyk, line 370: Would it make sense to combine these divisors together? Meaning, instead of `LagrangeKernelTransitionConstraintsDivisor` we have something like `LagrangeKernelConstraintDivisors` which contains both boundary and transition constraints for the Lagrange kernel. It could look something like this:
```rust
impl<E: FieldElement> LagrangeKernelConstraintDivisors<E> {
    pub fn get_transition_divisor_inv(&self, constraint_idx: usize, row_idx: usize) -> E {
        ...
    }

    pub fn get_boundary_divisor_inv(&self, row_idx: usize) -> E {
        ...
    }
}
```

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    /// Computes the transition constraints for the s-column.
+    ///
+    /// The s-column implements the cohomological sum-check argument of [1] and
+    /// the constraint we enfore is exactly Eq (4) in Lemma 1 in [1].
+    ///
+    /// [1]: https://eprint.iacr.org/2021/930
+    fn evaluate_s_column_transition(
+        &self,
+        evaluator: &impl LogUpGkrEvaluator<BaseField = E::BaseField>,
```
irakliyk, line 420: Not for this PR: I think the only reason why we need to pass the evaluator here (instead of getting it from `self.logup_gkr_constraints_evaluator`) is that the evaluator does not have a concrete type, and thus, cannot be a field inside a struct. If so, we may want to re-add the associated type to the `Air` trait for `LogUpGkrEvaluator`. But this would be for a follow-up PR.

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
```diff
@@ -0,0 +1,461 @@
...
+        main_frame: &EvaluationFrame<E::BaseField>,
+        aux_frame: &EvaluationFrame<E>,
+        query: &mut [E::BaseField],
+        divisor_at_step: E::BaseField,
```
irakliyk, line 424: Should we put the inverse divisors into `self.logup_gkr_constraints_evaluator`? It seems like this might be a natural place for them (and similar would apply to Lagrange kernel constraint divisors).

Also, one thing I realized: we tend to keep track of inverse divisors and composition coefficients separately. But I'm not sure there is a good reason for this. Instead, let's say we a "constraint modifier" $m$ (open to better names) as follows:

$$
m = \frac{c}{d_i}
$$

Where $c$ is the composition coefficient and $d_i$ is the divisor at row $I$. Then we could just multiply the numerator by this modifier - e.g., lines 441 - 445 below would become:
```rust

(rhs - lhs)
    * self
        .logup_gkr_constraints_evaluator
        .get_s_column_constraint_modifier(step)
```
I think we should be able to apply the same to the Lagrange kernel constraints.

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
```diff
@@ -0,0 +1,461 @@
...
+    /// and all auxiliary segments).
+    fn evaluate_fragment_full<T: TraceLde<E>>(
+        &self,
+        trace: &T,
+        domain: &StarkDomain<A::BaseField>,
+        fragment: &mut EvaluationTableFragment<E>,
+        trans_constraints_divisors: &LagrangeKernelTransitionConstraintsDivisor<E>,
+        boundary_divisors_inv: &[E],
+        s_col_constraint_divisor: &[E::BaseField],
+    ) {
```
irakliyk, line 218: Related to the above 2 comments, I'm wondering if it would be better to move `trans_constraints_divisors`, `boundary_divisors_inv`, and `s_col_constraint_divisor` into `self.logup_gkr_constraints_evaluator`. This way, we won't need to pass them as parameters here and it would also encapsulate LogUp-GKR-related code better.

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
```diff
@@ -0,0 +1,461 @@
...
+        // initialize buffers to hold trace values and evaluation results at each step
+        let mut main_frame = EvaluationFrame::new(trace.trace_info().main_segment_width());
+        let mut aux_frame = EvaluationFrame::new(trace.trace_info().aux_segment_width());
+        let mut tm_evaluations = vec![E::BaseField::ZERO; self.num_main_transition_constraints()];
+        let mut ta_evaluations = vec![E::ZERO; self.num_aux_transition_constraints()];
+        let mut evaluations = vec![E::ZERO; fragment.num_columns()];
+        let frame_length = trace.trace_info().length().ilog2() as usize + 1;
+        let mut lagrange_frame = LagrangeKernelEvaluationFrame::new_empty(frame_length);
```
irakliyk, line 226: I wonder if we should move the logic for determining `frame_length` into `LagrangeKernelEvaluationFrame::new_empty()`. Basically, it could be:
```rust
let mut lagrange_frame = LagrangeKernelEvaluationFrame::new_empty(trace_length);
```

air/src/air/logup_gkr/lagrange/frame.rs
```diff
@@ -29,10 +29,14 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
...
+    pub fn inner_mut(&mut self) -> &mut [E] {
+        &mut self.frame
+    }
+
     /// Constructs an empty Lagrange kernel evaluation frame from the raw column polynomial
     /// evaluations. The frame can subsequently be filled using [`Self::frame_mut`].
-    pub fn new_empty() -> Self {
-        Self { frame: Vec::new() }
+    pub fn new_empty(size: usize) -> Self {
+        Self { frame: vec![E::ZERO; size] }
```
irakliyk, line 39: nit: the naming feels a bit weird here. I would probably rename the current `new()` constructor into `with_values()` and rename this constructor into just `new()`.

irakliyk (commented):
Great job! Thank you! I left some comments and improvement suggestions inline.

Overall, I'm not sure I like creating a separate constraint evaluator specifically for LogUp-GKR, but I think once the inline comments are addressed, we can merge this structure to see how it impacts the overall proof generation performance. Then, we can think of a way to refactor it.

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
Al-Kindi-0: I agree. In the current implementation, it would make sense to have `LogUpGkrConstraintsDivisors` be part of 
of `LogUpGkrConstraintEvaluator`. The only issue with making this work is that requires a `StarkDomain` for its creation and I couldn't come up with a clean way to make it work.

prover/src/logup_gkr/mod.rs
```diff
@@ -4,6 +4,8 @@ use core::ops::Add;
 use air::{EvaluationFrame, GkrData, LogUpGkrEvaluator};
 use math::FieldElement;
 use sumcheck::{EqFunction, MultiLinearPoly, SumCheckProverError};
+#[cfg(feature = "concurrent")]
+pub use utils::rayon::prelude::*;
```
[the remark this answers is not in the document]
Al-Kindi-0: We don't, it was from a previous implementation.
Removed

prover/src/constraints/evaluator/logup_gkr.rs
```diff
@@ -6,152 +6,57 @@
...
 
 use air::{
-    Air, EvaluationFrame, GkrData, LagrangeConstraintsCompositionCoefficients,
-    LagrangeKernelConstraints, LagrangeKernelEvaluationFrame, LogUpGkrEvaluator,
+    Air, GkrData, LagrangeConstraintsCompositionCoefficients, LagrangeKernelConstraints,
+    LogUpGkrEvaluator,
 };
 use math::{batch_inversion, FieldElement};
+#[cfg(feature = "concurrent")]
+pub use utils::rayon::prelude::*;
```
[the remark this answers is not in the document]
Al-Kindi-0: Removed

air/src/air/logup_gkr/lagrange/frame.rs
```diff
@@ -29,10 +29,14 @@ impl<E: FieldElement> LagrangeKernelEvaluationFrame<E> {
         Self { frame }
     }
 
+    pub fn inner_mut(&mut self) -> &mut [E] {
```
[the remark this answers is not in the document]
Al-Kindi-0: Add, but kept `inner()` as it still makes things cleaner with than without

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
Al-Kindi-0: Created an issue

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
Al-Kindi-0: Makes sense!
I have now included multiplication with `c` in the computation of the divisors for the s-column as well as for the Lagrange kernel constraints.

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
Al-Kindi-0: Moved

air/src/air/logup_gkr/lagrange/frame.rs
Al-Kindi-0: Changed

prover/src/constraints/evaluator/logup_gkr.rs
```diff
@@ -192,8 +75,18 @@ where
...
+///
+/// Note that instead of storing `1 / div` for Lagrange and s-column transition constraints, we
+/// store instead `c / div` where `c` is the constraint composition coefficient associated to
+/// divisor `div`. This reduces the number of multiplications during the computation of
+/// evaluations of the constraint composition polynomial. We call `c / div` constraint evaluation
+/// multipliers or just constraint multipliers.
+pub(crate) struct LogUpGkrConstraintsDivisors<E: FieldElement> {
+    lagrange_transition_multipliers: Vec<E>,
+
+    lagrange_boundary_divisors_inv: Vec<E::BaseField>,
```
irakliyk, line 87: Why can't we store `lagrange_boundary_multipliers` here as we do for transition constraints? If this results in too much refactoring for this PR, let's create an issue for this.

irakliyk (approved):
Looks good! Thank you! I left two small comments inline.

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
irakliyk: This is because when we create a new evaluator, we don't pass the domain into the constructor, right? I think we can probably refactor this. Let's create an issue for this.

prover/src/constraints/evaluator/logup_gkr.rs
```diff
@@ -342,21 +280,22 @@ impl<E: FieldElement> TransitionDivisorEvaluator<E> {
     }
 }
 
-/// Computes the evaluations of the s-column divisor.
+/// Computes the evaluations of the s-column multipliers.
 ///
 /// The divisor for the s-column is $X^n - 1$ where $n$ is the trace length. This means that
 /// we need only compute `ce_blowup` many values and thus only that many exponentiations.
-fn compute_s_col_divisor<E: FieldElement>(
+pub(crate) fn compute_s_col_multipliers<E: FieldElement>(
```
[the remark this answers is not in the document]
Al-Kindi-0: Not anymore, changed

prover/src/constraints/evaluator/logup_gkr.rs
Al-Kindi-0: Changed to the same approach with multipliers

prover/src/constraints/evaluator/logup_gkr_evaluator.rs
Al-Kindi-0: Done

irakliyk (approved):
Looks good! Thank you!

PR #321, opened 2024-09-18
Add instrumentation

Al-Kindi-0:


prover/src/logup_gkr/mod.rs
```diff
@@ -56,6 +57,7 @@ pub struct EvaluatedCircuit<E: FieldElement> {
 impl<E: FieldElement> EvaluatedCircuit<E> {
     /// Creates a new [`EvaluatedCircuit`] by evaluating the circuit where the input layer is
     /// defined from the main trace columns.
+    #[instrument(skip_all)]
     pub fn new(
```
irakliyk, line 61: Here, it might be good to add a custom name (similar to how we do it [here](https://github.com/facebook/winterfell/blob/c388e16acc866af6ca9477273b9c24dd201fc5ff/prover/src/constraints/evaluator/default.rs#L53)).

irakliyk (approved):
Looks good! Thank you! Should we also add instrumentation to method like `evaluate_main_transition()`, `evaluate_aux_transistion()`, `evaluate_lagranage_transition()` (e.g., [here](https://github.com/facebook/winterfell/blob/c388e16acc866af6ca9477273b9c24dd201fc5ff/prover/src/constraints/evaluator/logup_gkr_evaluator.rs#L278)).

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Added!

Al-Kindi-0:
> Looks good! Thank you! Should we also add instrumentation to method like `evaluate_main_transition()`, `evaluate_aux_transistion()`, `evaluate_lagranage_transition()` (e.g., [here](https://github.com/facebook/winterfell/blob/c388e16acc866af6ca9477273b9c24dd201fc5ff/prover/src/constraints/evaluator/logup_gkr_evaluator.rs#L278)).

That would log the function each time it is called, which is not what we want. Probably there is a better way to get the overall time instead.

irakliyk (approved):
Looks good! Than you!

PR #324, opened 2024-09-19
Parallelize input layer generation

Al-Kindi-0:
Leads to about to 20-25% improvement in input layer generation.

### DIFF
--- a/prover/Cargo.toml
+++ b/prover/Cargo.toml
@@ -29,7 +29,7 @@ harness = false
 
 [features]
 async = ["maybe_async/async"]
-concurrent = ["crypto/concurrent", "math/concurrent", "fri/concurrent", "utils/concurrent", "dep:rayon", "std"]
+concurrent = ["crypto/concurrent", "math/concurrent", "fri/concurrent", "utils/concurrent", "std"]
 default = ["std"]
 std = ["air/std", "crypto/std", "fri/std", "math/std", "utils/std"]
 
@@ -39,7 +39,6 @@ crypto = { version = "0.9", path = "../crypto", package = "winter-crypto", defau
 fri = { version = "0.9", path = '../fri', package = "winter-fri", default-features = false }
 math = { version = "0.9", path = "../math", package = "winter-math", default-features = false }
 maybe_async = { path = "../utils/maybe_async" , package = "winter-maybe-async" }
-rayon = { version = "1.8", optional = true }
 sumcheck = { version = "0.1", path = "../sumcheck", package = "winter-sumcheck", default-features = false }
 thiserror = { version = "1.0", git = "https://github.com/bitwalker/thiserror", branch = "no-std", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["attributes"]}
--- a/prover/Cargo.toml
+++ b/prover/Cargo.toml
@@ -29,7 +29,7 @@ harness = false
 
 [features]
 async = ["maybe_async/async"]
-concurrent = ["crypto/concurrent", "math/concurrent", "fri/concurrent", "utils/concurrent", "std"]
+concurrent = ["crypto/concurrent", "math/concurrent", "fri/concurrent", "utils/concurrent", "dep:rayon", "std"]
 default = ["std"]
 std = ["air/std", "crypto/std", "fri/std", "math/std", "utils/std"]
 
@@ -39,6 +39,7 @@ crypto = { version = "0.9", path = "../crypto", package = "winter-crypto", defau
 fri = { version = "0.9", path = '../fri', package = "winter-fri", default-features = false }
 math = { version = "0.9", path = "../math", package = "winter-math", default-features = false }
 maybe_async = { path = "../utils/maybe_async" , package = "winter-maybe-async" }
+rayon = { version = "1.8", optional = true }
 sumcheck = { version = "0.1", path = "../sumcheck", package = "winter-sumcheck", default-features = false }
 thiserror = { version = "1.0", git = "https://github.com/bitwalker/thiserror", branch = "no-std", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["attributes"]}
--- a/prover/src/logup_gkr/mod.rs
+++ b/prover/src/logup_gkr/mod.rs
@@ -6,8 +6,8 @@ use math::FieldElement;
 use sumcheck::{EqFunction, MultiLinearPoly, SumCheckProverError};
 use tracing::instrument;
 use utils::{
-    batch_iter_mut, chunks, ByteReader, ByteWriter, Deserializable, DeserializationError,
-    Serializable,
+    batch_iter_mut, chunks, uninit_vector, ByteReader, ByteWriter, Deserializable,
+    DeserializationError, Serializable,
 };
 
 use crate::Trace;
@@ -111,7 +111,7 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
     /// Generates the input layer of the circuit from the main trace columns and some randomness
     /// provided by the verifier.
     fn generate_input_layer(
-        main_trace: &impl Trace<BaseField = E::BaseField>,
+        trace: &impl Trace<BaseField = E::BaseField>,
         evaluator: &impl LogUpGkrEvaluator<BaseField = E::BaseField>,
         log_up_randomness: &[E],
     ) -> CircuitLayer<E> {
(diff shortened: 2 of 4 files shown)


prover/src/logup_gkr/mod.rs
```diff
@@ -113,46 +115,105 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
...
-                let input_gates_values: Vec<CircuitWire<E>> = numerators
-                    .iter()
-                    .zip(denominators.iter())
-                    .map(|(numerator, denominator)| CircuitWire::new(*numerator, *denominator))
-                    .collect();
-                input_gates_values
+        #[cfg(feature = "concurrent")]
+        let input_layer_wires = {
+            let mut input_layer_wires = unsafe {
+                utils::uninit_vector(main_trace.main_segment().num_rows() * num_fractions)
```
irakliyk, line 121: I wonder if we could use the `batch_iter_mut!` macro here. Maybe something like this:
```rust
batch_iter_mut!(&mut input_layer_wires, 1024, |batch: &mut [E], batch_offset: usize| {
    // run a loop here similar to the one on lines 185 - 207 but taking the offset into account
}
```
If this works, it may speed things up a bit and should also allow us to use the same code for concurrent and serial versions.

Al-Kindi-0: Implemented!
The performance is the same as far as I am seeing, but the removing code duplication makes this proposal better.

prover/src/logup_gkr/mod.rs
```diff
@@ -114,45 +119,72 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
...
-                    .iter()
-                    .zip(denominators.iter())
-                    .map(|(numerator, denominator)| CircuitWire::new(*numerator, *denominator))
-                    .collect();
-                input_gates_values
-            };
-
-            input_layer_wires.extend(wires_from_trace_row);
-        }
+            unsafe { utils::uninit_vector(main_trace.main_segment().num_rows() * num_fractions) };
```
irakliyk, line 122: A couple of nits:

- We should probably call the variable just `trace`; `main_trace.main_segment()` looks a bit weird.
- We can import `uninit_vector()` from utils at the top of the file.

prover/src/logup_gkr/mod.rs
```diff
@@ -114,45 +119,72 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
...
+            |batch: &mut [CircuitWire<E>], batch_offset: usize| {
+                let mut main_frame = EvaluationFrame::new(num_cols);
+
+                let mut query = vec![E::BaseField::ZERO; num_oracles];
+                let mut periodic_values_row = vec![E::BaseField::ZERO; num_periodic_cols];
+                let mut numerators = vec![E::ZERO; num_fractions];
+                let mut denominators = vec![E::ZERO; num_fractions];
+                let num_rows_to_process = batch.len() / numerators.len();
+                let row_offset = batch_offset / numerators.len();
+                for i in (0..main_trace.main_segment().num_rows())
```
irakliyk, line 139: A couple of nits:

- `num_rows_to_process` does not depend on the batch, right? If so, I'd move it out of this code block. Also, maybe a better name would be `num_rows_per_batch`?
- Let's add a blank line before the `for` loop.

prover/src/logup_gkr/mod.rs
```diff
@@ -114,45 +119,72 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
...
+                        // We have that `(i - row_offset) * num_fractions <= (num_rows - 1) * num_fractions`
+                        // by virtue of `i - row_offset <= num_rows - 1`.
+                        // Since `j <= num_fractions - 1`, we get that
+                        // `(i - row_offset) * num_fractions + j <= num_rows * num_fractions - 1`.
+                        unsafe {
+                            let batch_entry =
+                                batch.get_unchecked_mut((i - row_offset) * num_fractions + j);
+                            *batch_entry = *value;
+                        }
+                    }
```
irakliyk, line 177: Would `.copy_from_slice()` not work better here? For example, something like:
```rust
batch[row_offset * num_fractions..].copy_from_slice(&wires_from_trace_row);
```

prover/src/logup_gkr/mod.rs
Al-Kindi-0: But the size of the batch is computed based on the number of threads, so we would have get that outside of the code block, right?

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Done

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Works! Only that it is `(i - row_offset)` instead of `row_offset`

prover/src/logup_gkr/mod.rs
irakliyk: Ah yes! Please ignore.

prover/Cargo.toml
```diff
@@ -39,6 +39,7 @@ crypto = { version = "0.9", path = "../crypto", package = "winter-crypto", defau
 fri = { version = "0.9", path = '../fri', package = "winter-fri", default-features = false }
 math = { version = "0.9", path = "../math", package = "winter-math", default-features = false }
 maybe_async = { path = "../utils/maybe_async" , package = "winter-maybe-async" }
+rayon = { version = "1.8", optional = true }
```
irakliyk, line 42: Do we still need this dependency?

prover/src/logup_gkr/mod.rs
irakliyk: I would still import `uninit_vector` on line 10 above and not use `utils::` here.

Al-Kindi-0: I haven't pushed yet :)

Al-Kindi-0: Now it should be fixed

prover/Cargo.toml
Al-Kindi-0: Removed

prover/src/logup_gkr/mod.rs
```diff
@@ -106,53 +111,72 @@ impl<E: FieldElement> EvaluatedCircuit<E> {
...
+                            &mut denominators,
+                        );
+
+                        let input_gates_values: Vec<CircuitWire<E>> = numerators
+                            .iter()
+                            .zip(denominators.iter())
+                            .map(|(numerator, denominator)| {
+                                CircuitWire::new(*numerator, *denominator)
+                            })
+                            .collect();
```
irakliyk, line 163: Allocating vectors here may add up as we do it for every row in the trace. As far as I can tell, we do this only so that we can copy the values to the `batch` immediately after we build the vector. Could we do it here? e.g., something like this:
```rust
let n = (i - row_offset) * num_fractions;
for (j, (numerator, denominator)) in
    numerators.iter().zip(denominators.iter()).enumerate()
{
    batch[n + j] = CircuitWire::new(*numerator, *denominator);
}
```
If this works, then we don't need to do `copy_from_slice()` at the of the iteration.

irakliyk (commented):
Looks good! Thank you! I left a couple more comments inline.

prover/src/logup_gkr/mod.rs
irakliyk: Or maybe even like this:
```rust
let n = (i - row_offset) * num_fractions;
for ((wire, numerator), denominator) in
    batch[n..].iter_mut().zip(numerators.iter()).zip(denominators.iter())
{
    *wire = CircuitWire::new(*numerator, *denominator);
}
```

irakliyk: Another option is to change `evaluate_query()` to something like:
```rust
fn evaluate_query<F, E>(
    &self,
    query: &[F],
    periodic_values: &[F],
    logup_randomness: &[E],
    wires: &mut [CircuitWire],
) where
    F: FieldElement<BaseField = Self::BaseField>,
    E: FieldElement<BaseField = Self::BaseField> + ExtensionOf<F>;
```
But maybe that's for another PR.

Al-Kindi-0: Great idea! Implemented

irakliyk (approved):
Looks good! Thank you! Did the last change affect the runtime at all?

Also, let's create an issue for changing the `evaluate_query()` interface

Al-Kindi-0:
> Looks good! Thank you! Did the last change affect the runtime at all?
Nothing I could notice

Issue #303, opened 2024-08-21
Optimize construction of multi-linear polynomials for final sum-check


Al-Kindi-0:
Before running the final sum-check of GKR in LogUp-GKR, we need to build multi-linear extensions of the relevant main trace columns. These are originally over the base field but the multi-linear extension as it is currently constructed stores them as extension field elements.
We should investigate if it is possible to avoid this as it could provide a nice reduction in the memory footprint.

PR #326, opened 2024-09-20
Parallelize s-column generation

Al-Kindi-0:
This brings about 60% improvement in s-column generation times.
Fixes also a bug in the construction of the MLEs.

### DIFF
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -305,7 +305,6 @@ impl<B: StarkField, P> AirContext<B, P> {
         let trace_length = self.trace_len();
         let transition_divisior_degree = trace_length - self.num_transition_exemptions();
 
-        // we use the identity: ceil(a/b) = (a + b - 1)/b
         let num_constraint_col =
             (highest_constraint_degree - transition_divisior_degree).div_ceil(trace_length);
 
--- a/air/src/air/context.rs
+++ b/air/src/air/context.rs
@@ -307,8 +307,7 @@ impl<B: StarkField, P> AirContext<B, P> {
 
         // we use the identity: ceil(a/b) = (a + b - 1)/b
         let num_constraint_col =
-            (highest_constraint_degree - transition_divisior_degree + trace_length - 1)
-                / trace_length;
+            (highest_constraint_degree - transition_divisior_degree).div_ceil(trace_length);
 
         cmp::max(num_constraint_col, 1)
     }
--- a/prover/src/logup_gkr/mod.rs
+++ b/prover/src/logup_gkr/mod.rs
@@ -401,7 +401,7 @@ pub fn build_s_column<E: FieldElement>(
     let mean = c / E::from(E::BaseField::from(num_rows as u32));
 
     let mut deltas = unsafe { uninit_vector(main_segment.num_rows()) };
-    deltas[0] = E::ZERO;
+    
     batch_iter_mut!(&mut deltas[1..], 1024, |batch: &mut [E], batch_offset: usize| {
         let mut query = vec![E::BaseField::ZERO; num_oracles];
         let mut main_frame = EvaluationFrame::<E::BaseField>::new(num_cols);
@@ -417,8 +417,10 @@ pub fn build_s_column<E: FieldElement>(
 
     // note that `deltas[0]` is set `0` and thus `deltas` satisfies the conditions for invoking
     // the function
-    prefix_sum(&mut deltas);
-    deltas
+    let mut cumulative_sum = deltas;
+    cumulative_sum[0] = E::ZERO;
+    prefix_sum(&mut cumulative_sum);
+    cumulative_sum
 }
 
 /// Builds the Lagrange kernel column at a given point.
--- a/prover/src/logup_gkr/mod.rs
+++ b/prover/src/logup_gkr/mod.rs
@@ -15,7 +15,10 @@ use crate::Trace;
 mod prover;
 pub use prover::prove_gkr;
 #[cfg(feature = "concurrent")]
-pub use utils::rayon::{current_num_threads as rayon_num_threads, prelude::*};
+pub use utils::{
+    rayon::{current_num_threads as rayon_num_threads, prelude::*},
+    {chunks, chunks_mut, iter, iter_mut},
+};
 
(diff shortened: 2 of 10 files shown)


prover/src/logup_gkr/mod.rs
```diff
@@ -366,33 +374,40 @@ where
...
+            evaluator.build_query(&main_frame, &mut query);
+            *v = gkr_data.compute_batched_query(&query) * lagrange_kernel_col[i + batch_offset]
+                - mean;
+        }
+    });
+
+    // note that `deltas[0]` is set `0` and thus `deltas` satisfies the conditions for invoking
+    // the function
+    let mut cumulative_sum = deltas;
+    cumulative_sum[0] = E::ZERO;
```
plafer, line 408: nit: can we set `deltas[0] - E::ZERO` on line 391 so that it's clear before we get to the `cumulative_sum` part that all elements on `deltas` were properly initialized?

prover/src/logup_gkr/mod.rs
```diff
@@ -366,33 +374,40 @@ where
...
-        result.push(cur_value);
-        last_value = cur_value;
-    }
-
-    result
+            evaluator.build_query(&main_frame, &mut query);
+            *v = gkr_data.compute_batched_query(&query) * lagrange_kernel_col[i + batch_offset]
+                - mean;
+        }
+    });
```
plafer, line 403: Did you benchmark the serial version of this approach vs the old? In the case where this version turns out to be slower, we could keep the old version for the serial case, and this one for the parallel case.

plafer (commented):
LGTM!

prover/src/logup_gkr/prover.rs
```diff
@@ -147,7 +150,7 @@ fn build_mls_from_main_trace_segment<E: FieldElement>(
         match oracle {
             LogUpGkrOracle::CurrentRow(index) => {
                 let col = main_trace.get_column(*index);
-                let values: Vec<E> = col.iter().map(|value| E::from(*value)).collect();
+                let values: Vec<E> = iter!(col).map(|value| E::from(*value)).collect();
                 let ml = MultiLinearPoly::from_evaluations(values);
                 mls.push(ml)
             },
```
irakliyk, line 156: I'm curious if iterating over oracles would work a bit better. For example:
```rust
for oracle in iter!(oracles) {
    ...
}
```

Also, a general question (not for this PR): we seem to be proactively converting base elements to extension field elements. Maybe there is a way to avoid it and only do this when we actually need to?

prover/src/logup_gkr/mod.rs
```diff
@@ -407,3 +422,66 @@ pub enum GkrProverError {
...
+#[cfg(feature = "concurrent")]
+fn prefix_sum_parallel<E: FieldElement>(vector: &mut [E], batch_size: usize) {
+    use utils::{chunks, chunks_mut, iter, iter_mut};
+
+    let num_partitions = (vector.len() + batch_size - 1) / batch_size;
+    let mut sum_per_partition = vec![E::ZERO; num_partitions];
+
+    chunks!(vector, batch_size)
+        .zip(iter_mut!(sum_per_partition))
+        .for_each(|(chunk, entry)| *entry = chunk.iter().fold(E::ZERO, |acc, term| acc + *term));
```
irakliyk, line 460: This will split things up into relatively small batches (length 1024 each), right? Not sure if it will be materially different, but I would probably split this into the number of partitions which is based on the number of threads.

irakliyk (approved):
Looks good! Thank you! I left a couple of comments inline - but we can also address them in a follow-up PR.

There also seem to be some merge conflicts.

prover/src/logup_gkr/prover.rs
Al-Kindi-0: I have answered in the other PR related to building MLEs.
For the second point, we have an issue to address that in #303

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Makes sense, updated

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Good point! Updated to include the old implementation in the serial case as it a bit faster than the other one and also since I updated the implementation for the parallel case to take the number of threads into account

prover/src/logup_gkr/prover.rs
Al-Kindi-0: For the first point, it is addressed in the other PR for building MLEs.
For the second point, we already have an issue for it in #303

prover/src/logup_gkr/mod.rs
Al-Kindi-0: Good point! Updated to take the number of threads into account

sumcheck/src/prover/high_degree.rs
```diff
@@ -458,7 +458,7 @@ fn sumcheck_round<E: FieldElement>(
                     &mut numerators,
                     &mut denominators,
                 );
-                poly_evals[0] = evaluate_composition_poly(
+                poly_evals[0] += evaluate_composition_poly(
```
irakliyk, line 461: Was this a bug previously?

Al-Kindi-0: This is part of #331, which explains why it was not caught

air/src/air/context.rs
```diff
@@ -307,8 +307,7 @@ impl<B: StarkField, P> AirContext<B, P> {
 
         // we use the identity: ceil(a/b) = (a + b - 1)/b
         let num_constraint_col =
-            (highest_constraint_degree - transition_divisior_degree + trace_length - 1)
-                / trace_length;
+            (highest_constraint_degree - transition_divisior_degree).div_ceil(trace_length);
```
irakliyk, line 310: I think comment on line 308 is no longer relevant.

irakliyk (approved):
Looks good! Thank you! I left one small comment inline.

air/src/air/context.rs
Al-Kindi-0: Indeed! Fixed now

PR #327, opened 2024-09-20
Fixes multilinear built from a `NextRow` oracle

plafer:


Al-Kindi-0 (commented):
Looks great, thank you!
As you propose, the shift has to be circular otherwise the s-column constraints will fail.

irakliyk (approved):
Looks good! Thank you!

PR #328, opened 2024-09-20
Reduce degree of sum-check round polynomials

Al-Kindi-0:
Implements the optimization in Section 3.3 [here](https://eprint.iacr.org/2024/108).
Benchmarks show about 20% improvement in sum-check proving times.


plafer (commented):
LGTM!

I tried it with Miden VM and did observe an improvement. IMO we can do any cleanup later.

irakliyk (approved):
Looks good! Thank you!

PR #329, opened 2024-09-20
Improve construction of MLEs from main trace segment

Al-Kindi-0:


prover/src/logup_gkr/prover.rs
```diff
@@ -141,20 +144,24 @@ fn build_mls_from_main_trace_segment<E: FieldElement>(
     oracles: &[LogUpGkrOracle],
     main_trace: &ColMatrix<<E as FieldElement>::BaseField>,
 ) -> Result<Vec<MultiLinearPoly<E>>, GkrProverError> {
-    let mut mls = vec![];
+    let mut mls = Vec::with_capacity(oracles.len());
 
     for oracle in oracles {
         match oracle {
             LogUpGkrOracle::CurrentRow(index) => {
```
irakliyk, line 151: A better option may be to parallelize across oracles (rather than within) - i.e., have a separate thread constructing each oracle. Not 100% sure - so, we should benchmark.

Al-Kindi-0: The trace being in column major form, not sure if horizontal is better, but running a benchmark should clarify that

irakliyk: But each oracle is based on a single column, right?

Al-Kindi-0: Indeed, that's why I think within oracle makes more sense, but maybe I am missing a fine detail about your point

Al-Kindi-0: Oh, now I see your point. Let me fix it!

irakliyk: Let's change this to:
```rust
iter!(oracles).for_each(|oracle| {
    match oracle {
            let col = main_trace.get_column(*index);
            let values: Vec<E> = col.iter().map(|value| E::from(*value)).collect();
            let ml = MultiLinearPoly::from_evaluations(values);
            mls.push(ml)
        },
        LogUpGkrOracle::NextRow(index) => {
            let col = main_trace.get_column(*index);

            let mut values: Vec<E> = unsafe { uninit_vector(col.len()) };
            values[col.len() - 1] = E::from(col[0]);
            (&mut values[..col.len() - 1])
                .iter_mut()
                .enumerate()
                .for_each(|(i, value)| *value = E::from(col[i + 1]));
            let ml = MultiLinearPoly::from_evaluations(values);
            mls.push(ml)
        },
    };
});
```

Al-Kindi-0: Yes, so we go with the first option i.e., across.

Al-Kindi-0: Went with the other options as I couldn't find a way to overcome the problem of having to share `mls` between threads in the parallel case.

irakliyk (approved):
Looks good! Thank you!

PR #331, opened 2024-09-25
Fix bug parallel execution of input layer proving

Al-Kindi-0:
Fixes a bug that caused the input layer proving to not run in parallel.

Also fixes a couple bugs in the implementation of the high degree sum-check and adds benchmarks that are specific to LogUp-GKR proving.


irakliyk (approved):
Looks good! Thank you!

PR #332, opened 2024-09-25
feat(utils): Implement `get_size_hint` for default impls

PhilippGackstatter:
Implements `get_size_hint` for the types that are implemented by default.

Currently, crates that call `u8.get_size_hint()` get the default value of 0 which makes it harder to implement `get_size_hint` for custom types, since they cannot simply call `get_size_hint` on the contained types recursively. This PR fixes that.

This change is non-breaking.

See also https://github.com/0xPolygonMiden/miden-base/pull/895#pullrequestreview-2326556758 for more context.

### DIFF
--- a/utils/core/src/serde/byte_writer.rs
+++ b/utils/core/src/serde/byte_writer.rs
@@ -75,9 +75,9 @@ pub trait ByteWriter: Sized {
     /// # Panics
     /// Panics if the value could not be written into `self`.
     fn write_usize(&mut self, value: usize) {
-        let length = usize_encoded_len(value);
         // convert the value into a u64 so that we always get 8 bytes from to_le_bytes() call
         let value = value as u64;
+        let length = usize_encoded_len(value);
 
         // 9-byte special case
         if length == 9 {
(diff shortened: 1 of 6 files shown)


utils/core/src/serde/byte_writer.rs
```diff
@@ -141,9 +141,9 @@ impl ByteWriter for alloc::vec::Vec<u8> {
 // HELPER FUNCTIONS
 // ================================================================================================
 
-/// Returns the length of the value in vint64 encoding.
-fn encoded_len(value: u64) -> usize {
-    let zeros = value.leading_zeros() as usize;
+/// Returns the length of the usize value in vint64 encoding.
+pub(super) fn usize_encoded_len(value: usize) -> usize {
```
irakliyk, line 145: Passing value as `usize` into here would cause issues on 32-bit systems (e.g., WASM). Let's revert this change.

irakliyk (commented):
Thank you! I left one comment inline - but other than that, it looks great.

utils/core/src/serde/byte_writer.rs
PhilippGackstatter: I see, good point!
Updated to cast the usize in the `get_size_hint` function to a u64 instead.

irakliyk (approved):
All looks good! Thank you!

PR #334, opened 2024-10-03
feat: Add `maybe_async_trait`

igamigo:
Adds `maybe_async_trait` which adds a way to use `async_trait` based on whether the `async` feature is turned on. The attribute needs to be set for the trait impl and definition blocks. The macro takes functions annotated with `#[maybe_async]`, makes them async, and then sets `#[async_trait]`.
One small consequence of this is that when you have code like this:

```rust
use winter_maybe_async::{maybe_async_trait, maybe_async}

#[maybe_async_trait]
pub trait Example {
    #[maybe_async]
    pub fn foo() { }
}
```

If you turn on the `async` feature, then `maybe_async` becomes an unused import (because `maybe_async_trait` essentially removes it from `foo`) which might affect clippy lints. One way to deal with this would be to import macros like this: `use winter_maybe_async::*`.

### DIFF
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -229,6 +229,8 @@ impl Deserializable for OodFrame {
 // OOD FRAME TRACE STATES
 // ================================================================================================
 
+/// Represents the state of a frame trace.
+///
 /// Stores the trace evaluations at `z` and `gz`, where `z` is a random Field element in
 /// `current_row` and `next_row`, respectively. If the Air contains a Lagrange kernel auxiliary
 /// column, then that column interpolated polynomial will be evaluated at `z`, `gz`, `g^2 z`, ...
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -229,7 +229,7 @@ impl Deserializable for OodFrame {
 // OOD FRAME TRACE STATES
 // ================================================================================================
 
-/// Represents the state of a frame trace.
+/// Trace evaluation frame at the out-of-domain point.
 ///
 /// Stores the trace evaluations at `z` and `gz`, where `z` is a random Field element in
 /// `current_row` and `next_row`, respectively. If the Air contains a Lagrange kernel auxiliary
--- a/crypto/src/hash/mds/mds_f64_8x8.rs
+++ b/crypto/src/hash/mds/mds_f64_8x8.rs
@@ -25,7 +25,6 @@ use math::{
 /// Hamish Ivey-Law and Jacqueline Nabaglo of Polygon Zero is based on Nabaglo's implementation
 /// in [Plonky2](https://github.com/mir-protocol/plonky2).
 /// The circulant matrix is identified by its first row: [23, 8, 13, 10, 7, 6, 21, 8].
-
 // MDS matrix in frequency domain.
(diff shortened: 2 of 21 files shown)


igamigo:
Had to add various newlines to make the first paragraph shorter in doc comments because of failing clippy checks. For the same reason I also had to add a usage of `div_ceil()` instead of manually dividing.

crypto/src/hash/mds/mds_f64_8x8.rs
```diff
@@ -25,7 +25,6 @@ use math::{
 /// Hamish Ivey-Law and Jacqueline Nabaglo of Polygon Zero is based on Nabaglo's implementation
 /// in [Plonky2](https://github.com/mir-protocol/plonky2).
 /// The circulant matrix is identified by its first row: [23, 8, 13, 10, 7, 6, 21, 8].
-
```
irakliyk, line 28: The doc comments are for the module - so, we should probably convert them into module level comments rather then attach them to the constant below.

air/src/proof/ood_frame.rs
```diff
@@ -229,6 +229,8 @@ impl Deserializable for OodFrame {
 // OOD FRAME TRACE STATES
 // ================================================================================================
 
+/// Represents the state of a frame trace.
```
irakliyk, line 232: I would write this as: "Trace evaluation frame at the out-of-domain point".

crypto/src/merkle/concurrent.rs
```diff
@@ -18,6 +18,8 @@ pub const MIN_CONCURRENT_LEAVES: usize = 1024;
 // PUBLIC FUNCTIONS
 // ================================================================================================
 
+/// Builds merkle nodes.
```
irakliyk, line 21: I would write this as: "Returns internal nodes of a Merkle tree constructed from the provided leaves."

utils/maybe_async/src/lib.rs
```diff
@@ -67,6 +67,131 @@ pub fn maybe_async(_attr: TokenStream, input: TokenStream) -> TokenStream {
...
+/// ```
+///
+/// When the `async` feature is disabled, the code remains unchanged, and neither the `async`
+/// keyword nor the `#[async_trait::async_trait(?Send)]` attribute is applied.
+#[proc_macro_attribute]
+pub fn maybe_async_trait(_attr: TokenStream, input: TokenStream) -> TokenStream {
+    // Try parsing the input as a trait definition
+    if let Ok(trait_item) = syn::parse::<ItemTrait>(input.clone()) {
+        let output = if cfg!(feature = "async") {
+            let mut async_trait = trait_item;
```
irakliyk, line 131: Question: why do we need to re-define the variable here but we don't need to do it for the impl block on lines 161 - 162 below?

irakliyk (approved):
Looks great! Thank you! I left a few comments inline - but these are for files not related to this PR. So, we can merge as is and then make updates separately.

utils/maybe_async/src/lib.rs
igamigo: I refactored this a bit a couple of times (initially I was outputting both versions but guarded with `#[cfg()]`), and this was a remaining artifact. Just changed it to be similar to the second block.

irakliyk (approved):
All looks good! Thank you!

Issue #335, opened 2024-10-16
Partitioned hashing support on multiple GPU's for trace and constraint commitments.


TheMenko:
To support trace commitments with multiple GPUs, we need partitioned hashing for both prover and the verifier, so that output from GPU code is the same as the one from the CPU, when computing on multiple devices.

## Partition size
The GPU code uses "partition size" to determine the number of columns to be processed per GPU. 
Since trace commitment has a minimum column number of 64. 16 as partition size is usually good. 
But we can come up with some formula for calculating it automatically. 
For e.g. `16 - (num_columns % 16) / (num_columns / 16)` would in most cases work quite well for any number of GPU's.  
## Per device partitioning
Number of partitions is calculated with `num_partitions = (num_columns + partition_size - 1) / partition_size`, and then we calculate how many devices will be used for computing with `num_devices = min(num_devices, num_partitions)`. This means that in case when we have an 8 column matrix, and a partition size = 8 (e.g. in constraint commitments), only a single GPU will be used for computing purposes, since `num_partitions` would be `1`.
## Hashing process
We need a buffer for storing partitioned linear hash output. The required buffer size is: `num_partitions * num_rows * blowup * CAPACITY` - enough space to store all the partitioned outputs.
Each GPU takes `partition_size` number of columns, and it hashes 8 columns at a time up to the `partition_size`. So if we have for e.g. 14 columns, 1 GPU will hash 8 columns and then 6 columns, and write that to the correct position in the buffer. If we set our partition size to 16 on a 64 column trace, `num_partitons` would be 4, so we allocate space for `4 * lde_domain_size * CAPACITY`, and when hashing we would write linear hash of 16 columns 4 times.
If we set partition_size to 8 however, we would need double the allocated space, and we would have 8 linear hash outputs written to the buffer.
After the linear hash step, each linear hash partition is taken, producing `CAPACITY` number of columns, and then the final linear hash is applied from which later on we build a Merkle tree.

irakliyk:
I think what we need from Winterfell side to support this is adding `num_partitions` parameter to `ProofOptions`. The default value for `num_partitions` would be 1, and in this case, everything would work as it does now.

But, if the user would set `num_partitions` to, say, 4, we'd compute row commitments as follows:

1. Split the row into 4 equal segments and hash each segment using liner hash.
2. Compute linear hash of the 4 values we got in the previous step.

[...]

PR #336, opened 2024-10-18
Option for partitioned trace committment

Al-Kindi-0:
Addresses #335.
This still in draft as two points need to be clarified:

1. Once we have the resulting `num_partitions` parts we need to hash all of them in order to get a final digest per row of the trace we are committing to. The issue is that directly hashing digests is not possible, or at least I couldn't see how to do it. One pass through the bytes representation but I am not sure this was the desired implementation. For now, the code just uses a naive `merge` in order to fold all the digests into one digest.
2. The separation between `num_partitions` equal to `1` vs greater than `1` is not very elegant but I couldn't see a way to merge handling of the two cases.

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -260,15 +260,12 @@ impl ProofOptions {
         FriOptions::new(self.blowup_factor(), folding_factor, remainder_max_degree)
     }
 
-    /// Returns the size of each partition used when committing to the main and auxiliary traces as
-    /// well as the constraint evaluation trace.
-    pub fn partition_size<E: FieldElement>(&self, num_columns: usize) -> usize {
-        let base_elements_per_partition = cmp::max(
-            (num_columns * E::EXTENSION_DEGREE).div_ceil(self.num_partitions as usize),
-            self.min_partition_size as usize,
-        );
-
-        base_elements_per_partition.div(E::EXTENSION_DEGREE)
+    /// Returns the `[PartitionOption]` used in this instance of proof options.
+    pub fn get_partition_option(&self) -> PartitionOption {
+        PartitionOption {
+            num_partitions: self.num_partitions,
+            min_partition_size: self.min_partition_size,
+        }
     }
 }
 
@@ -339,9 +336,6 @@ impl FieldExtension {
     }
 }
 
-// SERIALIZATION
-// ================================================================================================
-
 impl Serializable for FieldExtension {
     /// Serializes `self` and writes the resulting bytes into the `target`.
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
@@ -365,6 +359,42 @@ impl Deserializable for FieldExtension {
                 "value {value} cannot be deserialized as FieldExtension enum"
             ))),
         }
+    }
+}
+
+// PARTITION OPTION IMPLEMENTATION
+// ================================================================================================
+
+/// Defines the parameters used when committing to the traces generated during the protocol.
+pub struct PartitionOption {
+    num_partitions: u8,
+    min_partition_size: u8,
+}
+
+impl PartitionOption {
+    /// Returns a new instance of `[PartitionOption]`.
+    pub fn new(num_partitions: usize, min_partition_size: usize) -> Self {
+        Self {
+            num_partitions: num_partitions as u8,
+            min_partition_size: min_partition_size as u8,
+        }
+    }
+
+    /// Returns the size of each partition used when committing to the main and auxiliary traces as
+    /// well as the constraint evaluation trace.
+    pub fn partition_size<E: FieldElement>(&self, num_columns: usize) -> usize {
+        let base_elements_per_partition = cmp::max(
+            (num_columns * E::EXTENSION_DEGREE).div_ceil(self.num_partitions as usize),
+            self.min_partition_size as usize,
+        );
+
+        base_elements_per_partition.div(E::EXTENSION_DEGREE)
+    }
+}
+
+impl Default for PartitionOption {
+    fn default() -> Self {
+        Self { num_partitions: 1, min_partition_size: 1 }
     }
 }
 
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -94,8 +94,7 @@ pub struct ProofOptions {
     field_extension: FieldExtension,
     fri_folding_factor: u8,
     fri_remainder_max_degree: u8,
-    num_partitions: u8,
-    min_partition_size: u8,
+    partition_options: PartitionOptions,
 }
 
 // PROOF OPTIONS IMPLEMENTATION
@@ -169,8 +168,7 @@ impl ProofOptions {
             field_extension,
             fri_folding_factor: fri_folding_factor as u8,
             fri_remainder_max_degree: fri_remainder_max_degree as u8,
-            num_partitions: 1,
-            min_partition_size: 1,
+            partition_options: PartitionOptions::new(1, 1),
         }
     }
 
@@ -185,20 +183,7 @@ impl ProofOptions {
         num_partitions: usize,
         min_partition_size: usize,
     ) -> ProofOptions {
-        assert!(num_partitions >= 1, "number of partitions must be greater than or eqaul to 1");
-        assert!(num_partitions <= 16, "number of partitions must be smaller than or equal to 16");
-
-        assert!(
-            min_partition_size >= 1,
-            "smallest partition size must be greater than or equal to 1"
-        );
-        assert!(
-            min_partition_size <= 256,
-            "smallest partition size must be smaller than or equal to 256"
-        );
-
-        self.num_partitions = num_partitions as u8;
-        self.min_partition_size = min_partition_size as u8;
+        self.partition_options = PartitionOptions::new(num_partitions, min_partition_size);
 
         self
     }
@@ -260,12 +245,9 @@ impl ProofOptions {
         FriOptions::new(self.blowup_factor(), folding_factor, remainder_max_degree)
     }
 
-    /// Returns the `[PartitionOption]` used in this instance of proof options.
-    pub fn get_partition_option(&self) -> PartitionOption {
-        PartitionOption {
-            num_partitions: self.num_partitions,
-            min_partition_size: self.min_partition_size,
-        }
+    /// Returns the `[PartitionOptions]` used in this instance of proof options.
+    pub fn partition_options(&self) -> PartitionOptions {
+        self.partition_options
     }
 }
 
@@ -294,8 +276,8 @@ impl Serializable for ProofOptions {
         target.write(self.field_extension);
         target.write_u8(self.fri_folding_factor);
         target.write_u8(self.fri_remainder_max_degree);
-        target.write_u8(self.num_partitions);
-        target.write_u8(self.min_partition_size);
+        target.write_u8(self.partition_options.num_partitions);
+        target.write_u8(self.partition_options.min_partition_size);
     }
 }
 
@@ -366,14 +348,27 @@ impl Deserializable for FieldExtension {
 // ================================================================================================
 
 /// Defines the parameters used when committing to the traces generated during the protocol.
-pub struct PartitionOption {
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct PartitionOptions {
     num_partitions: u8,
     min_partition_size: u8,
 }
 
-impl PartitionOption {
-    /// Returns a new instance of `[PartitionOption]`.
-    pub fn new(num_partitions: usize, min_partition_size: usize) -> Self {
+impl PartitionOptions {
+    /// Returns a new instance of `[PartitionOptions]`.
+    pub const fn new(num_partitions: usize, min_partition_size: usize) -> Self {
+        assert!(num_partitions >= 1, "number of partitions must be greater than or eqaul to 1");
+        assert!(num_partitions <= 16, "number of partitions must be smaller than or equal to 16");
+
+        assert!(
+            min_partition_size >= 1,
+            "smallest partition size must be greater than or equal to 1"
+        );
+        assert!(
+            min_partition_size <= 256,
+            "smallest partition size must be smaller than or equal to 256"
+        );
+
         Self {
             num_partitions: num_partitions as u8,
             min_partition_size: min_partition_size as u8,
@@ -392,7 +387,7 @@ impl PartitionOption {
     }
 }
 
-impl Default for PartitionOption {
+impl Default for PartitionOptions {
     fn default() -> Self {
         Self { num_partitions: 1, min_partition_size: 1 }
     }
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -77,12 +77,13 @@ pub enum FieldExtension {
 /// with 128-bit collision resistance is used, soundness of a STARK proof cannot exceed 128 bits.
 ///
 /// In addition to the above, the parameter `num_partitions` is used in order to specify the number
-/// of partitions each of the traces committed to during proof generation is split into.
+/// of partitions each of the traces committed to during proof generation is split into, and
+/// the parameter `min_partition_size` gives a lower bound on the size of each such partition.
 /// More precisely, and taking the main segment trace as an example, the prover will split the main
-/// segment trace into `num_partitions` parts. The prover will then proceed to hash each part row-wise
-/// resulting in `num_partitions` digests per row of the trace. The prover finally combines
-/// the `num_partitions` digest (per row) into one digest (per row) and at this point the vector
-/// commitment scheme can be called.
+/// segment trace into `num_partitions` parts each of size at least `min_partition_size`. The prover
+/// will then proceed to hash each part row-wise resulting in `num_partitions` digests per row of
+/// the trace. The prover finally combines the `num_partitions` digest (per row) into one digest
+/// (per row) and at this point the vector commitment scheme can be called.
 /// In the case when `num_partitions` is equal to `1` the prover will just hash each row in one go
 /// producing one digest per row of the trace.
 #[derive(Debug, Clone, Eq, PartialEq)]
@@ -129,16 +130,48 @@ impl ProofOptions {
         fri_folding_factor: usize,
         fri_remainder_max_degree: usize,
     ) -> ProofOptions {
-        Self::with_num_partitions(
-            num_queries,
-            blowup_factor,
-            grinding_factor,
+        // TODO: return errors instead of panicking
+        assert!(num_queries > 0, "number of queries must be greater than 0");
+        assert!(num_queries <= MAX_NUM_QUERIES, "number of queries cannot be greater than 255");
+
+        assert!(blowup_factor.is_power_of_two(), "blowup factor must be a power of 2");
+        assert!(blowup_factor >= MIN_BLOWUP_FACTOR, "blowup factor cannot be smaller than 2");
+        assert!(blowup_factor <= MAX_BLOWUP_FACTOR, "blowup factor cannot be greater than 128");
+
+        assert!(
+            grinding_factor <= MAX_GRINDING_FACTOR,
+            "grinding factor cannot be greater than 32"
+        );
+
+        assert!(fri_folding_factor.is_power_of_two(), "FRI folding factor must be a power of 2");
+        assert!(
+            fri_folding_factor >= FRI_MIN_FOLDING_FACTOR,
+            "FRI folding factor cannot be smaller than 2"
+        );
+        assert!(
+            fri_folding_factor <= FRI_MAX_FOLDING_FACTOR,
+            "FRI folding factor cannot be greater than 16"
+        );
+
+        assert!(
+            (fri_remainder_max_degree + 1).is_power_of_two(),
+            "FRI polynomial remainder degree must be one less than a power of two"
+        );
+        assert!(
+            fri_remainder_max_degree <= FRI_MAX_REMAINDER_DEGREE,
+            "FRI polynomial remainder degree cannot be greater than 255"
+        );
+
+        Self {
+            num_queries: num_queries as u8,
+            blowup_factor: blowup_factor as u8,
+            grinding_factor: grinding_factor as u8,
             field_extension,
-            fri_folding_factor,
-            fri_remainder_max_degree,
-            1,
-            1,
-        )
+            fri_folding_factor: fri_folding_factor as u8,
+            fri_remainder_max_degree: fri_remainder_max_degree as u8,
+            num_partitions: 1,
+            min_partition_size: 1,
+        }
     }
 
     /// Returns a new instance of [ProofOptions] struct constructed from the specified parameters.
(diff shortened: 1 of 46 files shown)


irakliyk:
> Once we have the resulting `num_partitions` parts we need to hash all of them in order to get a final digest per row of the trace we are committing to. The issue is that directly hashing digests is not possible, or at least I couldn't see how to do it. One pass through the bytes representation but I am not sure this was the desired implementation. For now, the code just uses a naive `merge` in order to fold all the digests into one digest.

I wonder if we should define a new method on the `Hasher` trait - something like:
```rust
pub trait Hasher {
    fn merge_many(values: &[Self::Digest]) -> Self::Digest;
}
```

Al-Kindi-0:
> I wonder if we should define a new method on the `Hasher` trait - something like:
> 
> ```rust
> pub trait Hasher {
>     fn merge_many(values: &[Self::Digest]) -> Self::Digest;
> }
> ```

I don't see a better solution. I will go with this unless there are any other remarks.

crypto/src/hash/rescue/rp64_256/mod.rs
```diff
@@ -191,6 +191,45 @@ impl Hasher for Rp64_256 {
...
+        // the Rescue permutation.
+        // we don't need to apply any extra padding because we injected total number of elements
+        // in the input list into the capacity portion of the state during initialization.
+        if i > 0 {
+            Self::apply_permutation(&mut state);
+        }
+
+        // return the first 4 elements of the state as hash result
+        ElementDigest::new(state[DIGEST_RANGE].try_into().unwrap())
+    }
```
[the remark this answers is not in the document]
Al-Kindi-0: That is indeed simpler :)
Changed it now!

prover/src/matrix/row_matrix.rs
```diff
@@ -180,24 +180,45 @@ impl<E: FieldElement> RowMatrix<E> {
     /// * A vector commitment is computed for the resulting vector using the specified vector
     ///   commitment scheme.
     /// * The resulting vector commitment is returned as the commitment to the entire matrix.
-    pub fn commit_to_rows<H, V>(&self) -> V
+    pub fn commit_to_rows<H, V>(&self, num_partitions: usize) -> V
```
irakliyk, line 183: One other alternative (and I'm not sure if it is better) is to have two separate methods: the original one (which would assume partition size 1) and a new one which could look like so:
```rust
pub fn commit_to_rows_partitioned<H, V>(&self, num_partitions: usize) -> V {
    ...
}
```
Then the caller could decide which one to call.

prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -63,10 +63,11 @@ where
         trace_info: &TraceInfo,
         main_trace: &ColMatrix<E::BaseField>,
         domain: &StarkDomain<E::BaseField>,
+        num_partitions: usize,
     ) -> (Self, TracePolyTable<E>) {
         // extend the main execution trace and build a commitment to the extended trace
         let (main_segment_lde, main_segment_vector_com, main_segment_polys) =
-            build_trace_commitment::<E, E::BaseField, H, V>(main_trace, domain);
+            build_trace_commitment::<E, E::BaseField, H, V>(main_trace, domain, num_partitions);
```
irakliyk, line 70: I would probably make `num_partitions` a filed in `DefaultTraceLde` and then we won't need to pass it again to `set_aux_trace()` method.

prover/src/lib.rs
```diff
@@ -568,14 +581,19 @@ pub trait Prover {
         &self,
         trace: &Self::Trace,
         domain: &StarkDomain<Self::BaseField>,
+        num_partitions: usize,
         channel: &mut ProverChannel<'_, Self::Air, E, Self::HashFn, Self::RandomCoin, Self::VC>,
     ) -> (Self::TraceLde<E>, TracePolyTable<E>)
```
irakliyk, line 586: Why do we need to pass `num_partitions` into this function? Shouldn't we already have access to it via `self.options()`?

prover/src/trace/trace_lde/mod.rs
```diff
@@ -49,6 +49,7 @@ pub trait TraceLde<E: FieldElement>: Sync {
         &mut self,
         aux_trace: &ColMatrix<E>,
         domain: &StarkDomain<E::BaseField>,
+        num_partitions: usize,
     ) -> (ColMatrix<E>, <Self::HashFn as Hasher>::Digest);
```
irakliyk, line 53: As mentioned in one of the previous comments, I think we can get rid of this parameter as we can assume that `TraceLde` keeps track of `num_partitions` since construction.

verifier/src/channel.rs
```diff
@@ -191,9 +194,26 @@ where
...
+                .main_states
+                .rows()
+                .map(|row| {
+                    row.chunks(num_elements_per_partition)
+                        .zip(buffer.iter_mut())
+                        .for_each(|(chunk, buf)| *buf = H::hash_elements(chunk));
+                    H::merge_many(&buffer)
+                })
+                .collect()
+        };
```
irakliyk, line 215: I would probably create a helper function to handle this and code on line 206 - 243. It could look something like:
```rust
fn hash_row<H, E>(row: &[E], num_partitions: usize) -> H::Digest
```
It would be slightly less efficient than this implementation - but I think the impact on the verifier would be negligible.

irakliyk (commented):
Thank you! Looks good! I left some comments inline.

Another thing I'm wondering is whether it is better to use `partition_size` rather than `num_partitions` after all. Or maybe we should use a combination of both - e.g., something like `num_partitions` and `min_partition_size`.

The motivation is that we actually compute commitments 3 times: main trace, auxiliary trace, and constraint evaluations. All of these may have different widths. Let's for example assume that we have

- main race: 72 columns
- aux trace: 7 columns (degree 2 extension)
- constraint evaluations: 8 columns (degree 2 extension)

With `num_partitions = 4`, we'd split things into:

- main trace: 4 partitions with 18 columns each.
- aux trace: 3 partitions with 4 columns each + 1 partition with 2 columns
- constraint evaluations: 4 partitions with 4 columns each.

But if we have `num_partitions = 4` and `min_partition_size = 8`, we'd get the following:

- main trace: 4 partitions with 18 columns each.
- aux trace: 1 partition with 8 columns and 1 partition with 6 columns.
- constraint evaluations: 2 partitions with 8 columns each.

This seems like a slight better arrangement - but also not sure if the complexity is worth it.

We can also specify `num_partitions`/`partition_size` separately for each of the 3 components - but not sure if that's worth it either.

cc @TheMenko

TheMenko:
> With `num_partitions = 4`, we'd split things into:
> 
>     * main trace: 4 partitions with 18 columns each.
>     * aux trace: 3 partitions with 4 columns each + 1 partition with 2 columns
>     * constraint evaluations: 4 partitions with 4 columns each.

This is how GPU code currently operates.

> But if we have `num_partitions = 4` and `min_partition_size = 8`, we'd get the following:
> 
>     * main trace: 4 partitions with 18 columns each.
>     * aux trace: 1 partition with 8 columns and 1 partition with 6 columns.
>     * constraint evaluations: 2 partitions with 8 columns each.

This wouldn't be a lot of work on the GPU side. The only thing is that we should measure the performance for these two cases.
My current assumption is that for relatively small number of columns, doing `min_partition_size = 8` would have a positive impact.

air/src/options.rs
```diff
@@ -108,6 +119,7 @@ impl ProofOptions {
     /// - `grinding_factor` is greater than 32.
     /// - `fri_folding_factor` is not 2, 4, 8, or 16.
     /// - `fri_remainder_max_degree` is greater than 255 or is not a power of two minus 1.
+    /// - `num_partitions` is zero or greater than 64.
     #[rustfmt::skip]
     pub const fn new(
         num_queries: usize,
```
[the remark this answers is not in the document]
Al-Kindi-0: Fixed

prover/src/matrix/row_matrix.rs
Al-Kindi-0: I have kept with only one implementation but I don't feel strongly about it, let me know if you would still prefer adding another method

prover/src/trace/trace_lde/default/mod.rs
Al-Kindi-0: Done, but then had to re-add it back when implementing the solution with the minimal partition size

prover/src/trace/trace_lde/mod.rs
Al-Kindi-0: Done

verifier/src/channel.rs
Al-Kindi-0: Added

Al-Kindi-0:
@irakliyk , I added a proposal to implement your idea of minimal partition size and it is included in the penultimate commit

air/src/options.rs
```diff
@@ -74,6 +75,16 @@ pub enum FieldExtension {
...
+///
+/// In addition to the above, the parameter `num_partitions` is used in order to specify the number
+/// of partitions each of the traces committed to during proof generation is split into.
+/// More precisely, and taking the main segment trace as an example, the prover will split the main
+/// segment trace into `num_partitions` parts. The prover will then proceed to hash each part row-wise
+/// resulting in `num_partitions` digests per row of the trace. The prover finally combines
+/// the `num_partitions` digest (per row) into one digest (per row) and at this point the vector
+/// commitment scheme can be called.
+/// In the case when `num_partitions` is equal to `1` the prover will just hash each row in one go
+/// producing one digest per row of the trace.
```
irakliyk, line 87: We should probably mention `min_partition_size` here as well.

air/src/options.rs
```diff
@@ -108,14 +121,47 @@ impl ProofOptions {
...
+    pub const fn with_num_partitions(
+        num_queries: usize,
+        blowup_factor: usize,
+        grinding_factor: u32,
+        field_extension: FieldExtension,
+        fri_folding_factor: usize,
+        fri_remainder_max_degree: usize,
+        num_partitions: usize,
+        min_partition_size: usize,
     ) -> ProofOptions {
```
irakliyk, line 165: I would change it to a "builder" pattern constructor. That is, it takes `Self` as a parameter and modifies it. The signature would look something like this:
```rust
pub const fn with_partitions(
    mut self,
    num_partitions: usize,
    min_partition_size: usize
) -> Self {
    ...
}
```
Then, to create options with partitions, users would be able to do something like this:
```rust
let options = ProofOptions::new(...).with_partitions(...);
```

air/src/options.rs
```diff
@@ -140,13 +186,18 @@ impl ProofOptions {
             "FRI polynomial remainder degree cannot be greater than 255"
         );
 
+        assert!(num_partitions > 0, "number of partitions must be greater than 0");
+        assert!(num_partitions <= 16, "number of partitions must be less than 16");
```
irakliyk, line 190: Should we also check `min_partition_size` to be between 1 and 256?

verifier/src/channel.rs
```diff
@@ -35,6 +35,9 @@ pub struct VerifierChannel<
     // constraint queries
     constraint_commitment: H::Digest,
     constraint_queries: Option<ConstraintQueries<E, H, V>>,
+    partition_size_main: usize,
+    partition_size_aux: usize,
+    partition_size_constraint: usize,
```
irakliyk, line 40: Let's add a comment line above these as we have for other variables.

An alternative could be to create a helper structure to hold these.

prover/src/trace/trace_lde/default/tests.rs
```diff
@@ -31,6 +31,7 @@ fn extend_trace_table() {
         trace.info(),
         trace.main_segment(),
         &domain,
+        1,
```
irakliyk, line 34: Should this not be `trace.info().main_trace_width()`?

prover/src/lib.rs
```diff
@@ -182,6 +182,7 @@ pub trait Prover {
         trace_info: &TraceInfo,
         main_trace: &ColMatrix<Self::BaseField>,
         domain: &StarkDomain<Self::BaseField>,
+        num_partitions: usize,
     ) -> (Self::TraceLde<E>, TracePolyTable<E>)
     where
         E: FieldElement<BaseField = Self::BaseField>;
```
irakliyk, line 188: I don't think we need to pass `num_partitions` here as we can get this info from `self.options()`.

prover/src/trace/trace_lde/default/mod.rs
```diff
@@ -63,10 +63,11 @@ where
         trace_info: &TraceInfo,
         main_trace: &ColMatrix<E::BaseField>,
         domain: &StarkDomain<E::BaseField>,
+        partition_size: usize,
     ) -> (Self, TracePolyTable<E>) {
```
irakliyk, line 67: Something doesn't feel right about passing `partition_size` here and also separately in `set_aux_trace()`. Maybe we should have a separate struct - e.g., `PartitionOptions` which could look like so:
```rust
pub struct PartitionOptions {
    num_partitions: usize,
    min_partition_size: usize,
}
```
Then we could put this struct into a filed and use it both here and in `set_aux_trace()`.

We could put the `parition_size()` method onto this struct.

irakliyk (commented):
Thank you! Looks good! I've added a few more comments. The main one is about potentially introducing `PartitionOptions` struct.

prover/src/trace/trace_lde/default/mod.rs
Al-Kindi-0: Just to emphasize that the `partition_size` passed in here an in `set_aux_trace()` need not be the same

prover/src/lib.rs
Al-Kindi-0: Yes, indeed, but it just feels more congruent with how it is done for the auxiliary trace and the constraint trace.

air/src/options.rs
Al-Kindi-0: Added

air/src/options.rs
Al-Kindi-0: Updated to your proposal

verifier/src/channel.rs
Al-Kindi-0: Add a comment

prover/src/trace/trace_lde/default/tests.rs
Al-Kindi-0: Indeed, thank you

prover/src/lib.rs
irakliyk: In `set_aux_trace()` we use `partition_size` no? It seems to me that here (and also in the examples), the variable is named `num_partitions` while the value passed in is `partition_size`.

I think having `PartitionOptions` structure would help avoid such confusions and so would be a bit cleaner.

prover/src/trace/trace_lde/default/mod.rs
irakliyk: Yes, understood - but I still think having `PartitionOptions` would be ab it cleaner:

1. We'd be able to get it from `ProofOptions` as a single unit.
2. It would contain the logic for computing partition size.
3. We won't need other pass it to `TraceLde::set_aux_trace()` (because we'd put it as a single new filed into the default trace LDE) or `Prover::new_trace_lde()` (because we can get it from options).

Al-Kindi-0: Makes sense, I have now introduced `PartitionOption` struct.

Al-Kindi-0:
I have also benchmarked the code to get an idea on the delta after this PR. The results are logical:

**Next**

```
Generated 128 private-public key pairs in 4294 ms
Signed 128 messages in 1 ms
Verified 128 signature in 2866 ms
Generating proof for verifying 128 Lamport+ signatures
INFO     generate_proof [ 4.73s | 6.95% / 100.00% ]
INFO     ┝━ generate_execution_trace [ 956ms | 20.19% ] num_cols: 22
INFO     ┝━ build_domain [ 3.48ms | 0.07% ] trace_length: 131072 | lde_domain_size: 1048576
INFO     ┝━ commit_to_main_trace_segment [ 1.02s | 0.00% / 21.60% ]
INFO     │  ┝━ extend_execution_trace [ 924ms | 19.51% ] num_cols: 22 | blowup: 8
INFO     │  ┕━ compute_execution_trace_commitment [ 98.6ms | 2.08% ] commitment_domain_size: 1048576
INFO     ┝━ evaluate_constraints [ 1.39s | 29.46% ] ce_domain_size: 1048576
INFO     ┝━ commit_to_constraint_evaluations [ 784ms | 0.00% / 16.55% ]
INFO     │  ┝━ build_composition_poly_columns [ 112ms | 2.37% ] num_columns: 6
INFO     │  ┝━ evaluate_composition_poly_columns [ 570ms | 12.04% ]
INFO     │  ┕━ compute_constraint_evaluation_commitment [ 101ms | 2.14% ] log_domain_size: 20
INFO     ┝━ build_deep_composition_poly [ 103ms | 2.19% ]
INFO     ┝━ evaluate_deep_composition_poly [ 71.7ms | 1.52% ]
INFO     ┝━ compute_fri_layers [ 53.2ms | 1.12% ] num_layers: 4
INFO     ┝━ determine_query_positions [ 343µs | 0.01% ] grinding_factor: 16 | num_positions: 28
INFO     ┕━ build_proof_object [ 16.0ms | 0.34% ]
---------------------
Proof generated in 4733 ms
Proof size: 94.4 KB
Proof security: 99 bits (55 proven)
Proof hash: e74c7a32f4c8f137a161d531915117df744146d401614778caf8b6ad00cadfb1
---------------------
Proof verified in 2.3 ms
```

**num_partitions=1 & min_partition_size=1**

```
Generated 128 private-public key pairs in 4341 ms
Signed 128 messages in 1 ms
Verified 128 signature in 2906 ms
Generating proof for verifying 128 Lamport+ signatures
INFO     generate_proof [ 4.78s | 6.99% / 100.00% ]
INFO     ┝━ generate_execution_trace [ 951ms | 19.87% ] num_cols: 22
INFO     ┝━ build_domain [ 3.70ms | 0.08% ] trace_length: 131072 | lde_domain_size: 1048576
INFO     ┝━ commit_to_main_trace_segment [ 1.04s | 0.00% / 21.78% ]
INFO     │  ┝━ extend_execution_trace [ 938ms | 19.62% ] num_cols: 22 | blowup: 8
INFO     │  ┕━ compute_execution_trace_commitment [ 104ms | 2.17% ] commitment_domain_size: 1048576
INFO     ┝━ evaluate_constraints [ 1.40s | 29.36% ] ce_domain_size: 1048576
INFO     ┝━ commit_to_constraint_evaluations [ 799ms | 0.00% / 16.70% ]
INFO     │  ┝━ build_composition_poly_columns [ 109ms | 2.28% ] num_columns: 6
INFO     │  ┝━ evaluate_composition_poly_columns [ 584ms | 12.20% ]
INFO     │  ┕━ compute_constraint_evaluation_commitment [ 106ms | 2.22% ] log_domain_size: 20
INFO     ┝━ build_deep_composition_poly [ 105ms | 2.19% ]
INFO     ┝━ evaluate_deep_composition_poly [ 73.2ms | 1.53% ]
INFO     ┝━ compute_fri_layers [ 54.9ms | 1.15% ] num_layers: 4
INFO     ┝━ determine_query_positions [ 230µs | 0.00% ] grinding_factor: 16 | num_positions: 28
INFO     ┕━ build_proof_object [ 16.0ms | 0.33% ]
---------------------
Proof generated in 4783 ms
Proof size: 95.0 KB
Proof security: 99 bits (55 proven)
Proof hash: c57076047ef71a4eab840140bca38f7b4ab61470dfb15905974b1ec333fcdb8b
---------------------
Proof verified in 2.4 ms
```

**num_partitions=2 & min_partition_size=4**

```
Generated 128 private-public key pairs in 4512 ms
Signed 128 messages in 1 ms
Verified 128 signature in 3047 ms
Generating proof for verifying 128 Lamport+ signatures
INFO     generate_proof [ 4.91s | 6.51% / 100.00% ]
INFO     ┝━ generate_execution_trace [ 941ms | 19.17% ] num_cols: 22
INFO     ┝━ build_domain [ 3.54ms | 0.07% ] trace_length: 131072 | lde_domain_size: 1048576
INFO     ┝━ commit_to_main_trace_segment [ 1.13s | 0.00% / 23.02% ]
INFO     │  ┝━ extend_execution_trace [ 946ms | 19.29% ] num_cols: 22 | blowup: 8
INFO     │  ┕━ compute_execution_trace_commitment [ 183ms | 3.74% ] commitment_domain_size: 1048576
INFO     ┝━ evaluate_constraints [ 1.45s | 29.45% ] ce_domain_size: 1048576
INFO     ┝━ commit_to_constraint_evaluations [ 818ms | 0.00% / 16.67% ]
INFO     │  ┝━ build_composition_poly_columns [ 129ms | 2.64% ] num_columns: 6
INFO     │  ┝━ evaluate_composition_poly_columns [ 581ms | 11.84% ]
INFO     │  ┕━ compute_constraint_evaluation_commitment [ 108ms | 2.20% ] log_domain_size: 20
INFO     ┝━ build_deep_composition_poly [ 106ms | 2.16% ]
INFO     ┝━ evaluate_deep_composition_poly [ 72.1ms | 1.47% ]
INFO     ┝━ compute_fri_layers [ 54.7ms | 1.11% ] num_layers: 4
INFO     ┝━ determine_query_positions [ 1.19ms | 0.02% ] grinding_factor: 16 | num_positions: 28
INFO     ┕━ build_proof_object [ 16.5ms | 0.34% ]
---------------------
Proof generated in 4907 ms
Proof size: 96.0 KB
Proof security: 99 bits (55 proven)
Proof hash: f8fc4c8d1fab445366f6556ada92c04770981cf188ffd56dd0e7cabc274c920e
---------------------
Proof verified in 2.2 ms
```

air/src/options.rs
```diff
@@ -301,6 +362,42 @@ impl Deserializable for FieldExtension {
     }
 }
 
+// PARTITION OPTION IMPLEMENTATION
+// ================================================================================================
+
+/// Defines the parameters used when committing to the traces generated during the protocol.
+pub struct PartitionOption {
```
irakliyk, line 369: Let's add `#[derive(Debug, Clone, Eq, PartialEq)]`.

air/src/options.rs
```diff
@@ -301,6 +362,42 @@ impl Deserializable for FieldExtension {
...
+}
+
+impl PartitionOption {
+    /// Returns a new instance of `[PartitionOption]`.
+    pub fn new(num_partitions: usize, min_partition_size: usize) -> Self {
+        Self {
+            num_partitions: num_partitions as u8,
+            min_partition_size: min_partition_size as u8,
+        }
+    }
```
irakliyk, line 381: Since the constructor is public, I'd add the bound checks for the inputs (we can probably just move them here).

air/src/options.rs
```diff
@@ -82,6 +94,8 @@ pub struct ProofOptions {
     field_extension: FieldExtension,
     fri_folding_factor: u8,
     fri_remainder_max_degree: u8,
+    num_partitions: u8,
+    min_partition_size: u8,
```
irakliyk, line 98: is there a reason not to store these as a single `partition: PartitionOptions` field? Then, these can be exposed from options as a single field.

air/src/options.rs
```diff
@@ -206,6 +259,14 @@ impl ProofOptions {
         let remainder_max_degree = self.fri_remainder_max_degree as usize;
         FriOptions::new(self.blowup_factor(), folding_factor, remainder_max_degree)
     }
+
+    /// Returns the `[PartitionOption]` used in this instance of proof options.
+    pub fn get_partition_option(&self) -> PartitionOption {
```
irakliyk, line 264: If we store the options as a single field, this can be renamed to `partition_options()`.

prover/src/lib.rs
```diff
@@ -522,6 +523,7 @@ pub trait Prover {
         composition_poly_trace: CompositionPolyTrace<E>,
         num_constraint_composition_columns: usize,
         domain: &StarkDomain<Self::BaseField>,
+        partition_size: usize,
```
irakliyk, line 526: Why do we need to pass `partition_size` into this function? Shouldn't we be able to compute it from the info in `self`?

verifier/src/channel.rs
```diff
@@ -121,6 +125,19 @@ where
...
+                .get_partition_option()
+                .partition_size::<E::BaseField>(air.context().trace_info().main_trace_width()),
+            partition_size_aux: air
+                .options()
+                .get_partition_option()
+                .partition_size::<E>(air.context().trace_info().aux_segment_width()),
+            partition_size_constraint: air
+                .options()
+                .get_partition_option()
+                .partition_size::<E>(air.context().num_constraint_composition_columns()),
```
irakliyk, line 140: nit: I would probably define a variable `let partition_options = air.options().get_partition_option();` and then use it in this block of code. Would improve readability a bit.

irakliyk (commented):
Looks good! Thank you! I left a few more small comments inline.

air/src/options.rs
Al-Kindi-0: Added

air/src/options.rs
Al-Kindi-0: Makes sense, changed

prover/src/lib.rs
Al-Kindi-0: just a leftover, changed it now

verifier/src/channel.rs
Al-Kindi-0: agreed, changed now

irakliyk (approved):
All looks good! Thank you!

PR #338, opened 2024-10-30
Fix partition hashing

Al-Kindi-0:
This is a small fix for the condition under which partitioned hashing when committing is enabled.

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -377,6 +377,7 @@ impl PartitionOptions {
 
     /// Returns the size of each partition used when committing to the main and auxiliary traces as
     /// well as the constraint evaluation trace.
+    /// The returned size is given in terms of number of columns in the field `E`.
     pub fn partition_size<E: FieldElement>(&self, num_columns: usize) -> usize {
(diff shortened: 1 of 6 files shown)


air/src/options.rs
```diff
@@ -378,6 +378,9 @@ impl PartitionOptions {
     /// Returns the size of each partition used when committing to the main and auxiliary traces as
     /// well as the constraint evaluation trace.
     pub fn partition_size<E: FieldElement>(&self, num_columns: usize) -> usize {
```
irakliyk, line 380: Would be good to clarify in the comment that the returned number of columns is the number in the field `E` (rather than the base field of `E`).

irakliyk (approved):
Looks good! Thank you! I left one small comment inline. Also, could we rebase this on the `main` branch?

air/src/options.rs
Al-Kindi-0: Agreed, added

Al-Kindi-0:
I have now changed the base to `main`

irakliyk (approved):
All looks good! Thank you!

PR #340, opened 2024-11-14
Multi GPU partition fixes

gswirski:
This simplifies partition calculations, fixes the wrong buffer size for merging partitions, and provides accessor methods for GPU to determine correct partition sizes.

`min_partition_size` is replaced by `hash_rate` - how many base elements are consumed by one hash iteration. This allows us to calculate at which point it makes sense to start a new partition. `min_partition_size = hash_rate / E::EXTENSION_DEGREE`.

Supersedes https://github.com/facebook/winterfell/pull/339

### DIFF
--- a/air/src/options.rs
+++ b/air/src/options.rs
@@ -4,7 +4,7 @@
 // LICENSE file in the root directory of this source tree.
 
 use alloc::vec::Vec;
-use core::{cmp, ops::Div};
+use core::cmp;
 
 use fri::FriOptions;
 use math::{FieldElement, StarkField, ToElements};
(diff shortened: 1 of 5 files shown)


air/src/options.rs
```diff
@@ -181,9 +181,9 @@ impl ProofOptions {
     pub const fn with_partitions(
         mut self,
         num_partitions: usize,
-        min_partition_size: usize,
+        hash_rate: usize,
     ) -> ProofOptions {
```
irakliyk, line 185: Let's also update doc comments on line 180 above.

air/src/options.rs
```diff
@@ -351,49 +351,56 @@ impl Deserializable for FieldExtension {
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub struct PartitionOptions {
     num_partitions: u8,
-    min_partition_size: u8,
+    hash_rate: u8,
 }
```
irakliyk, line 355: It may be good to add a brief explanation about how `num_partitions` and `hash_rate` are used to the doc comments for this struct.

air/src/options.rs
```diff
@@ -351,49 +351,56 @@ impl Deserializable for FieldExtension {
...
 
     /// Returns the size of each partition used when committing to the main and auxiliary traces as
     /// well as the constraint evaluation trace.
     /// The returned size is given in terms of number of columns in the field `E`.
     pub fn partition_size<E: FieldElement>(&self, num_columns: usize) -> usize {
-        if self.num_partitions == 1 && self.min_partition_size == 1 {
+        if self.num_partitions == 1 {
             return num_columns;
         }
+        let min_partition_size = self.hash_rate as usize / E::EXTENSION_DEGREE;
```
irakliyk, line 385: Seems like this can under-estimate `min_partition_size` (assuming it is in base elements). For example, for `hash_rate=8` and `EXTENSION_DEGREE = 3`, we'd get 2 here instead of 6.

air/src/options.rs
```diff
@@ -351,49 +351,56 @@ impl Deserializable for FieldExtension {
...
         let base_elements_per_partition = cmp::max(
             (num_columns * E::EXTENSION_DEGREE).div_ceil(self.num_partitions as usize),
-            self.min_partition_size as usize,
+            min_partition_size,
         );
 
         base_elements_per_partition.div(E::EXTENSION_DEGREE)
```
irakliyk, line 391: Would this logic work correctly in all edge cases? For example, if I did my math correctly, for:

- `EXTENSION_DEGREE = 3`
- `num_columns = 7`
- `hash_rate = 8`
- `num_partitions = 2`.

We'd get 3 as the partition size - which wouldn't be enough to cover all 7 columns with 2 partitions.

irakliyk (commented):
Looks good! Thank you! I left a few questions/comments inline.

gswirski:
All comments have been addressed. Partition sizes are expressed in `E` elements, not base elements, I've added docs to clarify. Also, since it's taking so many attempts to get right, I've also added unit tests with the edge cases you've mentioned.

air/src/options.rs
gswirski: Great catch, thank you! The old logic was overly complicated and wrong. Hopefully it's fixed now.

irakliyk (approved):
All looks good! Thank you!

PR #341, opened 2024-11-18
Implement `core::error::Error` for all error types

PhilippGackstatter:
Implements `core::error::Error` for all error types in a non-breaking way, which is why this targets `main`.

The motivation is to be able to wrap these errors in `Box<dyn Error>` and use them as `source` errors which requires the trait to be implemented. 

Also fixes a clippy lint introduced in a recent rust nightly version replacing `string.as_bytes().len()` with `string.len()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.10.2 (TBD)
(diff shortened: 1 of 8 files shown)


irakliyk (approved):
Looks good! Thank you!

PR #356, opened 2024-12-27
Add security estimate in unique decoding regime

Al-Kindi-0:
As the title says.
It also simplifies the estimate in LDR and moreover gives the round-by-round (knowledge) soundness error now.


air/src/proof/mod.rs
```diff
@@ -212,13 +212,30 @@ impl Deserializable for Proof {
...
 fn get_conjectured_security(
     options: &ProofOptions,
     base_field_bits: u32,
     trace_domain_size: usize,
     collision_resistance: u32,
-) -> u32 {
+) -> ConjecturedSecurityBits {
     // compute max security we can get for a given field size
     let field_size = base_field_bits * options.field_extension().degree();
     let field_security = field_size - (trace_domain_size * options.blowup_factor()).ilog2();
```
WizardOfMenlo, line 241: As far as I can tell, Conjecture 1 in ethSTARK does not include the trace domain size? Might be good to add a reference to where that comes from.

air/src/proof/mod.rs
```diff
@@ -241,7 +261,7 @@ fn get_proven_security(
     base_field_bits: u32,
     trace_domain_size: usize,
     collision_resistance: u32,
-) -> u32 {
+) -> ProvenSecurityBits {
     let m_min: usize = 3;
```
WizardOfMenlo, line 265: Maybe add a comment re the strategy used to compute m?

air/src/proof/mod.rs
```diff
@@ -258,15 +278,22 @@ fn get_proven_security(
...
             options,
             base_field_bits,
             trace_domain_size,
             m_optimal as usize,
         ),
         collision_resistance as u64,
-    ) as u32
+    ) as u32;
+
+    let unique_decoding = cmp::min(
```
WizardOfMenlo, line 291: Move to before list_decoding as it does not depend on m?

air/src/proof/mod.rs
```diff
@@ -282,78 +309,116 @@ fn proven_security_protocol_for_m(
...
+    epsilons_bits_neg.into_iter().fold(f64::INFINITY, |a, b| a.min(b)) as u64
 }
 
 // HELPER FUNCTIONS
 // ================================================================================================
 
-/// Computes the largest proximity parameter m needed for Theorem 8
-/// in <https://eprint.iacr.org/2022/1216.pdf> to work.
+/// Computes the largest proximity parameter m such that eta is greater than 0 in Theorem 1 in
+/// https://eprint.iacr.org/2021/582.
```
WizardOfMenlo, line 413: Thm 1 in ethSTARK is the Johnson bound, which does not mention m at all. Maybe we should refer to the definition of eta here.

WizardOfMenlo (commented):
Did a pass, apart from some small minor docs improvements looks good to me.
One comment, it might be interesting to have an intermediate conjecture setting, such as Conjecture 8.4 in proximity gaps. This sits between the Toy problem and the provable in terms of strenght.

air/src/proof/mod.rs
```diff
@@ -99,28 +99,28 @@ impl Proof {
...
-            )
-        } else {
-            get_proven_security(
-                self.context.options(),
-                self.context.num_modulus_bits(),
-                self.trace_info().length(),
-                H::COLLISION_RESISTANCE,
-            )
-        }
+    /// This is Conjecture 1 in https://eprint.iacr.org/2021/582.
```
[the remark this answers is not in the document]
Al-Kindi-0: Added

air/src/proof/mod.rs
Al-Kindi-0: Indeed, this is a sort of hybrid between the conjecture on the toy problem (or more succinctly toy conjecture) and  conjecture 8.4 in the proximity gaps paper.
As you suggest, it might make more sense to remove the domain size from the above so as to match the toy conjecture exactly and then add an additional "security mode" to capture Conjecture 8.4 with the exponents therein equal to 1.

air/src/proof/mod.rs
Al-Kindi-0: Added

air/src/proof/mod.rs
Al-Kindi-0: Done

air/src/proof/mod.rs
Al-Kindi-0: Indeed, it is hidden in the proof. Expanded the comment in order to make it easier to get the point

Al-Kindi-0:
@irakliyk can you do a pass on this PR? This way I can build the next PR on the changes in this one.

air/src/proof/mod.rs
```diff
@@ -212,13 +213,30 @@ impl Deserializable for Proof {
 // HELPER FUNCTIONS
 // ================================================================================================
 
+pub struct ConjecturedSecurityBits(pub u32);
+
+pub struct ProvenSecurityBits {
+    unique_decoding: u32,
+    list_decoding: u32,
+}
```
irakliyk, line 221: Given that proof security estimation code is now the bulk of this module, I'm thinking we should move this code to its own module - e.g., `src/proof/security.rs`.

I would also probably just call these structs `ConjecturedSecurity` and `ProvenSecurity`, and I would probably attach the functions that compute the security to these structs. This could look like so:
```rust

pub struct ConjecturedSecurity(u32);

impl ConjecturedSecurity {

    pub fn compute(
        options: &ProofOptions,
        base_field_bits: u32,
        trace_domain_size: usize,
        collision_resistance: u32,
    ) -> Self {
        ...
    }

    pub fn bits(&self) -> u32 {
        self.0
    }

    pub fn is_at_least(bits: u32) -> bool {
        ...
    }
}
```
We could do the same for the `ProvenSecurity` struct and the interface should remain consistent between the two.

air/src/proof/mod.rs
```diff
@@ -99,28 +99,29 @@ impl Proof {
...
+            self.context.num_modulus_bits(),
+            self.trace_info().length(),
+            H::COLLISION_RESISTANCE,
+        )
+    }
+    /// Returns security level of this proof (in bits) using proven security.
+    ///
+    /// Usually, the number of queries needed for provable security is 2x - 3x higher than
+    /// the number of queries needed for conjectured security at the same security level.
+    pub fn security_level_proven<H: Hasher>(&self) -> ProvenSecurityBits {
```
irakliyk, line 118: nit: I would rename these into `proven_security()` and `conjectured_security()`. Internally, the could just call `ConjecturedSecurity::compute()` and `ProvenSecurity::compute()`.

irakliyk (commented):
Looks good! Thank you! (and sorry for such a delay with the review). I left a couple of comments inline - they are mostly about getting the API to be more ergonomic.

Also, did these changes affect security estimates for any of our common parameter sets?

air/src/proof/mod.rs
Al-Kindi-0: Makes sense, changed to the proposed structure.

air/src/proof/mod.rs
Al-Kindi-0: Changed

Al-Kindi-0:
> Also, did these changes affect security estimates for any of our common parameter sets?

Nothing changed, but due to not using multiple calls to `cmp::min` in LDR, we have regained 2 bits of security.

air/src/proof/mod.rs
```diff
@@ -99,28 +95,29 @@ impl Proof {
...
-            get_proven_security(
-                self.context.options(),
-                self.context.num_modulus_bits(),
-                self.trace_info().length(),
-                H::COLLISION_RESISTANCE,
-            )
-        }
+    /// This is the conjecture on the security of the Toy problem (Conjecture 1)
+    /// in https://eprint.iacr.org/2021/582.
+    pub fn security_level_conjectured<H: Hasher>(&self) -> ConjecturedSecurity {
```
irakliyk, line 102: nit: I would rename this to just `conjectured_security()`.

air/src/proof/mod.rs
```diff
@@ -99,28 +95,29 @@ impl Proof {
...
+            self.context.num_modulus_bits(),
+            self.trace_info().length(),
+            H::COLLISION_RESISTANCE,
+        )
+    }
+    /// Returns security level of this proof (in bits) using proven security.
+    ///
+    /// Usually, the number of queries needed for provable security is 2x - 3x higher than
+    /// the number of queries needed for conjectured security at the same security level.
+    pub fn security_level_proven<H: Hasher>(&self) -> ProvenSecurity {
```
irakliyk, line 114: nit: I would rename this to just `proven_security()`.

examples/src/main.rs
```diff
@@ -92,15 +92,21 @@ fn main() {
 
     let proof_bytes = proof.to_bytes();
     println!("Proof size: {:.1} KB", proof_bytes.len() as f64 / 1024f64);
-    let conjectured_security_level = options.get_proof_security_level(&proof, true);
+    let conjectured_security_level = options.get_proof_security_level_conjectured(&proof);
```
irakliyk, line 95: Not from this PR, but I think this line should move into code block which starts on line 99 and ends can also get rid of lines 112 and 113, right?

air/src/proof/security.rs
```diff
@@ -0,0 +1,728 @@
...
+//! Contains helper structs and methods to estimate the security of STARK proofs.
+
+use core::cmp;
+
+use crate::ProofOptions;
+
+const GRINDING_CONTRIBUTION_FLOOR: u32 = 80;
+const MAX_PROXIMITY_PARAMETER: u64 = 1000;
+
+pub struct ConjecturedSecurity(u32);
```
irakliyk, line 15: nit: let's add a section separator above this line with title `CONJECTURED SECURITY`.

Also, let's add doc comments to this struct.

air/src/proof/security.rs
```diff
@@ -0,0 +1,728 @@
...
+
+impl ConjecturedSecurity {
+    pub fn compute(
```
irakliyk, line 18: Since this is a part of the public interface, let's add doc comments to all public methods of this struct.

air/src/proof/security.rs
```diff
@@ -0,0 +1,728 @@
...
+
+    pub fn is_at_least(&self, bits: u32) -> bool {
+        self.0 >= bits
+    }
+}
+
+pub struct ProvenSecurity {
+    unique_decoding: u32,
+    list_decoding: u32,
+}
```
irakliyk, line 52: nit: let's add a section separator above this line with title `PROVEN SECURITY`.

Also, let's add doc comments to this struct.

air/src/proof/security.rs
```diff
@@ -0,0 +1,728 @@
...
+
+impl ProvenSecurity {
+    pub fn compute(
```
irakliyk, line 55: Since this is a part of the public interface, let's add doc comments to all public methods of this struct.

air/src/proof/security.rs
```diff
@@ -0,0 +1,728 @@
...
+    value.ceil()
+}
+
+#[cfg(not(feature = "std"))]
+pub fn ceil(value: f64) -> f64 {
+    libm::ceil(value)
+}
+
+#[cfg(test)]
+mod prove_security_tests {
```
irakliyk, line 282: nit: let's add a section separator above this line with title `TESTS`.

Also, we could probably rename the module to just `tests`.

irakliyk (approved):
Looks good! Thank you! I left some more comments inline - mostly related to doc comments.

One question: how long does computing security (both conjectured and proven) now take?

air/src/proof/mod.rs
Al-Kindi-0: changed

air/src/proof/mod.rs
Al-Kindi-0: changed

examples/src/main.rs
Al-Kindi-0: I think we want to print only conjectured security estimates in `std` and both estimates in `no-std`, so I think the code is correct

air/src/proof/security.rs
Al-Kindi-0: added

air/src/proof/security.rs
Al-Kindi-0: added

air/src/proof/security.rs
Al-Kindi-0: added

air/src/proof/security.rs
Al-Kindi-0: added

air/src/proof/security.rs
Al-Kindi-0: changed

irakliyk (approved):
Looks good! Thank you!

PR #357, opened 2025-01-22
Add option for algebraic batching to build DEEP polynomial

Al-Kindi-0:
We probably need to sync the next branch so that we can change the base branch for this PR.
Once #356 is merged, I can update the security estimator in this PR.


air/src/options.rs
```diff
@@ -406,6 +419,16 @@ impl Default for PartitionOptions {
...
 
+// BATCHING OPTION
+// ================================================================================================
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(u8)]
+pub enum BatchingType {
+    Linear = 0,
+    Algebraic = 1,
+}
```
irakliyk, line 430: Question: is `Linear` the term usually used for the non-algebraic approach? I thought it would be more like "independent" or something like that.

Also, let's add doc comments (both for the structure and for the variants) explaining what these do.

Lastly, I would probably implement `Serializable` and `Deserializable` on this structure so that we don't have to leak its internals higher up the chain.

air/src/options.rs
```diff
@@ -291,6 +300,10 @@ impl Deserializable for ProofOptions {
             FieldExtension::read_from(source)?,
             source.read_u8()? as usize,
             source.read_u8()? as usize,
+            match source.read_u8()? {
+                0 => BatchingType::Linear,
+                _ => BatchingType::Algebraic,
```
irakliyk, line 305: I would do exact matching - e.g., if it's not `0` or `1`, let's return an error.

air/src/options.rs
```diff
@@ -246,6 +249,11 @@ impl ProofOptions {
     pub fn partition_options(&self) -> PartitionOptions {
         self.partition_options
     }
+
+    /// Returns the `[BatchingType]` used in this instance of proof options for the DEEP polynomial.
+    pub fn batching_type_deep(&self) -> BatchingType {
+        self.batching_deep
+    }
```
irakliyk, line 256: nit: I would probably name this `batching_deep()` or `deep_batching()`.

Also, would the following description be a bit more precise:
```rust
/// Returns the `[BatchingType]` which defines how linear combinations are to be computed during
/// the composition of the DEEP polynomial.
///
/// Linear batching implies that independently drawn random values will be used for random linear
/// combination computation, while Algebraic batching implies that powers of a single random
/// value will be used.
///
/// Depending on other parameters, Algebraic batching may lead to a small reduction in the security
/// level of the generated proofs, but avoids extra calls to the random oracle (i.e., hash function).
```

air/src/options.rs
```diff
@@ -92,6 +92,7 @@ pub struct ProofOptions {
     fri_folding_factor: u8,
     fri_remainder_max_degree: u8,
     partition_options: PartitionOptions,
+    batching_deep: BatchingType,
```
irakliyk, line 95: nit: I would move this above `partition_options`.

Also, let's add a paragraph to the doc comments above to explain the meaning of this parameter (maybe re-use the text I suggested in the above comment).

air/src/air/mod.rs
```diff
@@ -579,25 +582,59 @@ pub trait Air: Send + Sync {
...
+                let mut c_coefficients = Vec::new();
+                for _ in 0..self.context().num_constraint_composition_columns() {
+                    c_coefficients.push(public_coin.draw()?);
+                }
+
+                let lagrange_cc = if self.context().has_lagrange_kernel_aux_column() {
+                    Some(public_coin.draw()?)
+                } else {
+                    None
+                };
```
irakliyk, line 600: nit: I would maybe move this logic into methods on `DeepCompositionCoefficients`. For example, something like:

```rust
impl<E: FieldElement> DeepCompositionCoefficients<E> {
    pub draw_linear(
        public_coin: &mut impl RandomCoin<BaseField = E::BaseField>,
        num_values: usize,
    ) -> Self {
        ...
    }

    pub draw_algebraic(
        public_coin: &mut impl RandomCoin<BaseField = E::BaseField>,
        num_values: usize,
    ) -> Self {
        ...
    }
}
```

And then the code here becomes:
```rust
match self.context().options.deep_batching() {
    BatchingType::Linear => DeepCompositionCoefficients::draw_linear(),
    BatchingType::Algebraic => DeepCompositionCoefficients::draw_algebraic(),
}
```

air/src/air/mod.rs
```diff
@@ -579,25 +582,59 @@ pub trait Air: Send + Sync {
...
+            BatchingType::Algebraic => {
+                let mut t_coefficients = Vec::new();
+                let alpha: E = public_coin.draw()?;
+                t_coefficients
+                    .extend_from_slice(&get_power_series(alpha, self.trace_info().width()));
+
+                let mut c_coefficients = Vec::new();
+                c_coefficients.extend_from_slice(&get_power_series_with_offset(
+                    alpha,
+                    alpha.exp(((self.trace_info().width()) as u32).into()),
```
irakliyk, line 617: nit: I would define a variable for this to make it clear what we are computing with `alpha.exp()`.

air/src/air/mod.rs
```diff
@@ -579,25 +582,59 @@ pub trait Air: Send + Sync {
...
+                    trace: t_coefficients,
+                    constraints: c_coefficients,
+                    lagrange: lagrange_cc,
+                })
+            },
+            BatchingType::Algebraic => {
+                let mut t_coefficients = Vec::new();
+                let alpha: E = public_coin.draw()?;
+                t_coefficients
+                    .extend_from_slice(&get_power_series(alpha, self.trace_info().width()));
```
irakliyk, line 612: `get_power_series()` would return `1` as the first entry in the series - is that an issue?

prover/benches/lagrange_kernel.rs
irakliyk, line 1: Not for this PR, but would be great to remove Lagrange kernel-related code from the `main` branch.

irakliyk (commented):
Looks good! Thank you! I left some comments inline.

Also, shouldn't proof security estimates be affected by the type of batching we do? Or is this too difficult to incorporate in the current structure?

irakliyk:
Also, let's resolve the merge conflict against the current `main`.

air/src/air/mod.rs
Al-Kindi-0: Yeah, that is the intended behavior. Namely if we have two codewords we want to batch then the batched word will be $\alpha^0 \cdot f_0 + \alpha^1 \cdot f_1$ i.e. $f_0 + \alpha\cdot f_1$

air/src/options.rs
Al-Kindi-0: I think this is the most common terminology used. Linear signifies the fact that multivariate polynomial with indeterminates the random value symbols has (total) degree one in each variable. Algebraic signifies that the randomness' are derived using an algebraic relation. Algebraic is also called sometimes parametric or curved.
I added a small comment to explain a bit more.

Added the comments and better (de)serialization

air/src/options.rs
Al-Kindi-0: Edited a bit and added

Al-Kindi-0: Used `batching_used_for_deep_poly` instead, but can change as it is not a strong preference

air/src/options.rs
Al-Kindi-0: Done

air/src/air/mod.rs
Al-Kindi-0: The issue is that `num_values` is not enough, we need the dimensions of the committed traces

air/src/air/mod.rs
Al-Kindi-0: done

prover/benches/lagrange_kernel.rs
Al-Kindi-0: Agreed, I can do that before we make the next release

Al-Kindi-0:
> Also, shouldn't proof security estimates be affected by the type of batching we do? Or is this too difficult to incorporate in the current structure?

Exactly! The only thing we need is the the widths of each of the committed traces or the total number of committed columns. If this number is say `N`, then there will be a (potential) loss of `log2(N-1)`.

air/src/options.rs
```diff
@@ -246,6 +252,19 @@ impl ProofOptions {
...
+    /// Returns the `[BatchingType]` defining the method used for batching the multi-point quotients
+    /// defining the DEEP polynomial.
+    ///
+    /// Linear batching implies that independently drawn random values per multi-point quotient
+    /// will be used to do the batching, while Algebraic batching implies that powers of a single
+    /// random value are used.
+    ///
+    /// Depending on other parameters, Algebraic batching may lead to a small reduction in the security
+    /// level of the generated proofs, but avoids extra calls to the random oracle (i.e., hash function).
+    pub fn batching_used_for_deep_poly(&self) -> BatchingType {
```
irakliyk, line 265: nit: let's rename this to `deep_poly_batching_method()` and also, let's rename `BatchingType` to `BatchingMethod`.

air/src/options.rs
```diff
@@ -275,6 +294,7 @@ impl Serializable for ProofOptions {
         target.write_u8(self.fri_remainder_max_degree);
         target.write_u8(self.partition_options.num_partitions);
         target.write_u8(self.partition_options.hash_rate);
+        target.write_u8(self.batching_deep as u8);
```
irakliyk, line 297: We should be able to write this as `target.write(self.batching_deep)`.

air/src/options.rs
```diff
@@ -291,6 +311,10 @@ impl Deserializable for ProofOptions {
             FieldExtension::read_from(source)?,
             source.read_u8()? as usize,
             source.read_u8()? as usize,
+            match source.read_u8()? {
+                0 => BatchingType::Linear,
+                _ => BatchingType::Algebraic,
```
irakliyk, line 316: We shouldn't need to use the `match` statement here - i.e., the following should work: `BatchingType::read_from(source)?`.

irakliyk (commented):
Looks good! Thank you! I left a couple more comments inline.

air/src/options.rs
Al-Kindi-0: changed

air/src/options.rs
Al-Kindi-0: forgot to change this after adding (de)-serialization to the relevant struct.
Changed now

air/src/options.rs
Al-Kindi-0: same as above

air/src/air/mod.rs
irakliyk: Ah - I see! I guess we could pass in the individual dimensions - e.g., like:
```rust
impl<E: FieldElement> DeepCompositionCoefficients<E> {
    pub draw_linear(
        public_coin: &mut impl RandomCoin<BaseField = E::BaseField>,
        trace_width: usize,
        num_constraint_composition_columns: usize,
    ) -> Self {
        ...
    }

    pub draw_algebraic(
        public_coin: &mut impl RandomCoin<BaseField = E::BaseField>,
        trace_width: usize,
        num_constraint_composition_columns: usize,
    ) -> Self {
        ...
    }
}
```
I'm skipping the Lagrange column info as we'll be removing it.

irakliyk:
> Exactly! The only thing we need is the the widths of each of the committed traces or the total number of committed columns. If this number is say `N`, then there will be a (potential) loss of `log2(N-1)`.

Would we do it in this PR? Or should we leave it for a follow-up?

Al-Kindi-0:
Some clippy warnings related to `div_ceil` are appearing, think we should just supress them as we are not really doing `div_ceil`

Al-Kindi-0:
> > Exactly! The only thing we need is the the widths of each of the committed traces or the total number of committed columns. If this number is say `N`, then there will be a (potential) loss of `log2(N-1)`.
> 
> Would we do it in this PR? Or should we leave it for a follow-up?

I propose we do it in a subsequent PR

air/src/air/mod.rs
Al-Kindi-0: Yeah, that would work. I can change to this

irakliyk:
> Some clippy warnings related to `div_ceil` are appearing, think we should just supress them as we are not really doing `div_ceil`

I guess these are functionally equivalent to `div_ceil` but, yeah - if that's not what we are doing, I'd suppress them.

winterfell/src/lib.rs
```diff
@@ -584,6 +584,7 @@
 //!     FieldExtension::None,
 //!     8,  // FRI folding factor
 //!     31, // FRI max remainder polynomial degree
+//!     BatchingMethod::Linear, // Type of batching used in computing DEEP polynomial
```
irakliyk, line 587: nit: I would change the comment to "// method of batching used in computing DEEP polynomial".

air/src/options.rs
```diff
@@ -246,6 +252,19 @@ impl ProofOptions {
     pub fn partition_options(&self) -> PartitionOptions {
         self.partition_options
     }
+
+    /// Returns the `[BatchingType]` defining the method used for batching the multi-point quotients
+    /// defining the DEEP polynomial.
```
irakliyk, line 257: This should be "[`BatchingMethod`]".

air/src/options.rs
```diff
@@ -406,6 +427,55 @@ impl Default for PartitionOptions {
     }
 }
 
+// BATCHING OPTION
```
irakliyk, line 430: nit: I would change this to "BATCHING METHOD".

air/src/options.rs
```diff
@@ -406,6 +427,55 @@ impl Default for PartitionOptions {
...
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        target.write_u8(*self as u8);
+    }
+}
+
+impl Deserializable for BatchingMethod {
+    /// Reads [BatchingType] from the specified `source` and returns the result.
+    ///
+    /// # Errors
+    /// Returns an error if the value does not correspond to a valid [BatchingType].
```
irakliyk, line 467: We should use "[`BatchingMethod`]" in these comments.

air/src/options.rs
```diff
@@ -406,6 +427,55 @@ impl Default for PartitionOptions {
...
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        match source.read_u8()? {
+            0 => Ok(BatchingMethod::Linear),
+            1 => Ok(BatchingMethod::Algebraic),
+            n => Err(DeserializationError::InvalidValue(format!(
+                "value {n} cannot be deserialized as a BatchingType enum"
```
irakliyk, line 473: Should be "BatchingMethod enum".

air/src/air/coefficients.rs
```diff
@@ -110,3 +111,56 @@ pub struct DeepCompositionCoefficients<E: FieldElement> {
...
+        t_coefficients.extend_from_slice(&get_power_series(alpha, trace_width));
+
+        let mut c_coefficients = Vec::new();
+
+        let alpha_pow_trace_width = alpha.exp((trace_width as u32).into());
+        c_coefficients.extend_from_slice(&get_power_series_with_offset(
+            alpha,
+            alpha_pow_trace_width,
+            num_constraint_composition_columns,
+        ));
```
irakliyk, line 158: I think this would be correct, but maybe worth double checking: the last value in `t_coefficients` would be different from the first value in `c_coefficients`, right? That is:
```
t_coefficients[trace_width - 1] * alpha == c_coefficients[0]
```

irakliyk (approved):
Looks good! Thank you! I left some comments inline (mostly about rename `BatchType` into `BatchMethod`). After these are addressed, we can merge.

air/src/air/coefficients.rs
Al-Kindi-0: Would a `debug_assert_eq` be enough?

irakliyk: yes - that should be sufficient.

air/src/options.rs
Al-Kindi-0: fixed

air/src/options.rs
Al-Kindi-0: fixed

air/src/options.rs
Al-Kindi-0: fixed

PR #358, opened 2025-01-23
Update serialization logic of OOD frame

Al-Kindi-0:
Moves away from serializing the current and next raw in a zig-zag manner and serializes them instead as two vectors one after the other.

### DIFF
--- a/air/src/proof/ood_frame.rs
+++ b/air/src/proof/ood_frame.rs
@@ -43,8 +43,8 @@ impl OodFrame {
     /// Updates the trace state portion of this out-of-domain frame, and returns the hash of the
     /// trace states.
     ///
-    /// The out-of-domain frame is stored as one vector of interleaved values, one from the current
-    /// row and the other from the next row. Given the input frame
+    /// The out-of-domain frame is stored as one vector built from the concatenation of values of two
+    /// vectors, the current row vector and the next row vector. Given the input frame
     ///
     ///    +-------+-------+-------+-------+-------+-------+-------+-------+
     ///    |   a1  |   a2  |  ...  |  an   |  c1   |  c2   |  ...  |  cm   |
@@ -54,7 +54,7 @@ impl OodFrame {
     ///
     /// with n being the main trace width and m the auxiliary trace width, the values are stored as
     ///
-    /// [a1, b1, a2, b2, ..., an, bn, c1, d1, c2, d2, ..., cm, dm]
+    /// [a1, ..., an, c1, ..., cm, b1, ..., bn, d1, ..., dm]
     ///
     /// into `Self::trace_states` (as byte values).
     ///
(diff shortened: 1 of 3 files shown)


air/src/proof/ood_frame.rs
```diff
@@ -67,7 +67,7 @@ impl OodFrame {
     {
         assert!(self.trace_states.is_empty(), "trace sates have already been set");
 
-        // save the evaluations with the current and next evaluations interleaved for each polynomial
+        // save the evaluations of the current and then next evaluations for each polynomial
         let (main_and_aux_trace_states, lagrange_trace_states) = trace_ood_frame.to_trace_states();
```
irakliyk, line 71: When we build `elements_to_hash` below, should we try to arrange things so that we always end up with 4 base field elements per extension field element (for 64-bit fields)? Basically, the arrangement would look something like so:

For sequence of degree 2 extension field elements: `(a_0, a_1), (b_0, b_1), (c_0, c_1), ...`, `elements_to_hash` would look like:
```
[a_0, a_1, 0, 0, b_0, b_1, 0, 0, c_0, c_1, 0, 0, ...]
```
And then, if we were in the degree 3 extension field, it'd look like:
```
[a_0, a_1, a_2, 0, b_0, b_1, b_2, 0, c_0, c_1, c_2, 0, ...]
```

One open question here is how to handle this generically given that we could also be in a 128-bit field.

Al-Kindi-0: I would keep it as it is for now as the benefits of the layout you proposed are still not clear to me, especially given its "unnaturalness".

Al-Kindi-0: But maybe you are thinking of it in light of the following proposal https://github.com/0xPolygonMiden/miden-vm/issues/1610#issuecomment-2571587390 ?
If yes, then we can include it in this PR to build on it the solution in the previous issue.

Al-Kindi-0: > One open question here is how to handle this generically given that we could also be in a 128-bit field.

I would probably just go for the simplest solution of conditioning on the case of $64$ bit primes. Once we commit to this change downstream, we can come back and modify the implementation to something more generic.

irakliyk: > But maybe you are thinking of it in light of the following proposal [0xPolygonMiden/miden-vm#1610 (comment)](https://github.com/0xPolygonMiden/miden-vm/issues/1610#issuecomment-2571587390) ?
> If yes, then we can include it in this PR to build on it the solution in the previously mentioned issue.

Yes, that's what I had in mind. Basically, "unlashing" becomes more "uniform" across different extension fields and also potentially easier to process in that context (though, it depends on whether this methodology will be used in the VM)..

Al-Kindi-0: So the question is do we do it now with partial genericity or do we leave it till we need it?

irakliyk: Let's keep as is for now, but create an issue to look into it.

irakliyk (approved):
Looks good! Thank you! Let's first merge #357, then rebase and merge this PR.

PR #359, opened 2025-01-30
Remove GKR related code

Al-Kindi-0:
As the integration of LogUp-GKR is on pause for the time being, it makes sense to remove the code related to it from the codebase.
If things change and we go ahead with the integration then there is at least one branch containing the full version of LogUp-GKR namely, the `logup-gkr` branch.
Putting in draft until some open PRs are merged.


prover/src/lib.rs
```diff
@@ -343,25 +321,11 @@ pub trait Prover {
...
-
-                (Some(gkr_proof), aux_rand_elements)
-            } else {
-                let rand_elements = air
-                    .get_aux_rand_elements(channel.public_coin())
-                    .expect("failed to draw random elements for the auxiliary trace segment");
-
-                (None, AuxRandElements::new(rand_elements))
-            };
+            let aux_rand_elements = AuxRandElements::new(rand_elements);
```
irakliyk, line 328: We can probably make `Air::get_aux_rand_elements()` return `AuxRandElements`, and then we won't need this line.

air/src/proof/ood_frame.rs
```diff
@@ -291,37 +233,26 @@ impl<E: FieldElement> TraceOodFrame<E> {
...
     fn has_aux_frame(&self) -> bool {
         self.current_row.len() > self.main_trace_width
     }
 
-    /// Returns the main/aux frame and Lagrange kernel frame as element vectors as described in
-    /// [`OodFrame::set_trace_states`].
-    fn to_trace_states(&self) -> (Vec<E>, Vec<E>) {
+    /// Returns the main/aux frames as element vectors. Specifically, the main and auxiliary frames
+    /// are interleaved, as described in [`OodFrame::set_trace_states`].
+    fn to_trace_states(&self) -> Vec<E> {
```
irakliyk, line 251: I don't think we need to change the comment here.

air/src/proof/ood_frame.rs
```diff
@@ -67,26 +64,15 @@ impl OodFrame {
     {
         assert!(self.trace_states.is_empty(), "trace sates have already been set");
 
-        // save the evaluations of the current and then next evaluations for each polynomial
-        let (main_and_aux_trace_states, lagrange_trace_states) = trace_ood_frame.to_trace_states();
+        // save the evaluations with the current and next evaluations interleaved for each polynomial
+        let main_and_aux_trace_states = trace_ood_frame.to_trace_states();
```
irakliyk, line 68: I don't think the comment here should mentioned "interleaved".

irakliyk (approved):
Looks good! Thank you! I left a few small comments inline. Once these are addressed, we can merge.

verifier/src/lib.rs
```diff
@@ -176,36 +176,13 @@ where
...
-
-            public_coin.reseed(trace_commitments[AUX_TRACE_IDX]);
-
-            Some(AuxRandElements::new(rand_elements))
-        }
+        let rand_elements = air
+            .get_aux_rand_elements(&mut public_coin)
+            .expect("failed to generate the random elements needed to build the auxiliary trace");
+
+        public_coin.reseed(trace_commitments[AUX_TRACE_IDX]);
```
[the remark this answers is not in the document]
Al-Kindi-0: Created an issue for it

prover/src/lib.rs
Al-Kindi-0: Agreed, changed it now

air/src/proof/ood_frame.rs
Al-Kindi-0: Indeed, probably happened during the rebase

air/src/proof/ood_frame.rs
Al-Kindi-0: Indeed, fixed now

irakliyk (approved):
Looks good! Thank you!

PR #361, opened 2025-02-01
Update security estimator to take into account method of batching

Al-Kindi-0:
Updates the estimator to take into account the method of batching used in computing the DEEP polynomial.
Note also:

- I have updated the error term $\epsilon_3$ as this was missed in the previous PR which updated the error of the protocol to be the round-by-round error.
- I have updated the conjectured security formula to match exactly the one given in the ethSTARK paper. Moreover, and as far as I can see, the formula is independent of the type of batching used and hence the security estimator for conjectured security is unaffected.

### DIFF
--- a/air/src/proof/security.rs
+++ b/air/src/proof/security.rs
@@ -921,4 +921,178 @@ mod tests {
 
         assert!(security_1 < security_2);
     }
+
+    #[test]
+    fn batching_method_udr() {
+        let field_extension = FieldExtension::Quadratic;
+        let base_field_bits = BaseElement::MODULUS_BITS;
+        let fri_folding_factor = 8;
+        let fri_remainder_max_degree = 255;
+        let grinding_factor = 20;
+        let blowup_factor = 8;
+        let num_queries = 120;
+        let collision_resistance = 128;
+        let trace_length = 2_usize.pow(16);
+        let total_num_of_polys = 1 << 1;
+
+        let mut options = ProofOptions::new(
+            num_queries,
+            blowup_factor,
+            grinding_factor,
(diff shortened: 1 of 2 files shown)


air/src/proof/security.rs
```diff
@@ -24,19 +24,16 @@ const MAX_PROXIMITY_PARAMETER: u64 = 1000;
...
     pub fn compute(
         options: &ProofOptions,
         base_field_bits: u32,
-        trace_domain_size: usize,
         collision_resistance: u32,
     ) -> Self {
         // compute max security we can get for a given field size
-        let field_size = base_field_bits * options.field_extension().degree();
-        let field_security = field_size - (trace_domain_size * options.blowup_factor()).ilog2();
+        let field_security = base_field_bits * options.field_extension().degree();
```
irakliyk, line 36: This makes conjectured security estimate a bit less conservative, right?

irakliyk (commented):
Looks good! Thank you! I left one small question inline.

Also, could we add a few tests to sanity check security estimate results for algebraic batching?

air/src/proof/security.rs
Al-Kindi-0: Indeed, the conjecture implies that the soundness of the protocol is not affected by the domain size or, as far as I can see, the batching used in either the computation of the quotient polynomials or the DEEP polynomial.

Al-Kindi-0:
> Also, could we add a few tests to sanity check security estimate results for algebraic batching?

Added a couple of tests to check for that.

irakliyk (approved):
All looks good! Thank you!

PR #363, opened 2025-02-06
Add option for algebraic/curve batching for constraint composition polynomial

### LINKED ISSUE #362: Use algebraic/curve batching for constraints
I think that we should probably be thinking of transition to an algebraic/curve method of batching for computing the constraint composition polynomial.
The reason for this is that the linear-in-the-number-of-constraints degradation in the soundness due to the use of algebraic/curve batching is not critical to the overall RbR soundness of the protocol. This is because, whenever the domain size is bigger than the number of constraints, we are almost always guaranteed that the soundness error of the constraint batching step will never be the largest error and hence will not affect the RbR soundness error of the protocol.

Al-Kindi-0:
Closes #362

### DIFF
--- a/air/src/air/coefficients.rs
+++ b/air/src/air/coefficients.rs
@@ -14,8 +14,11 @@ use math::{get_power_series, get_power_series_with_offset, FieldElement};
 ///
 /// These coefficients are created by the
 /// [Air::get_constraint_composition_coefficients()](crate::Air::get_constraint_composition_coefficients)
-/// function. In the interactive version of the protocol, the verifier draws these coefficients
-/// uniformly at random from the extension field of the protocol.
+/// function. In the interactive version of the protocol, the verifier either draws these coefficients
+/// uniformly at random from the extension field of the protocol or draws a single random extension
+/// field element $\alpha$ and defines the coefficients as $\alpha_i = \alpha^i$. We call the former
+/// way way of generating the alpha-s, and hence of batching the constraints, linear/affine batching
+/// while we call the latter algebraic/curve batching.
 ///
 /// There is one coefficient for each constraint so that we can compute a random linear
 /// combination of constraints as:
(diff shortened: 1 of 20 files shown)


air/src/air/coefficients.rs
```diff
@@ -28,12 +31,71 @@ use math::{get_power_series, get_power_series_with_offset, FieldElement};
...
     pub boundary: Vec<E>,
 }
 
+impl<E: FieldElement> ConstraintCompositionCoefficients<E> {
+    /// Generates the random values used in the construction of the constraint composition
+    /// polynomial when linear batching is used.
+    pub fn draw_linear(
+        public_coin: &mut impl RandomCoin<BaseField = E::BaseField>,
+        num_of_transition_constraints: usize,
+        num_of_boundary_constraints: usize,
```
irakliyk, line 51: nit: we usually simply say `num_transition_constraints` (without the `of` part).

air/src/air/coefficients.rs
```diff
@@ -28,12 +31,71 @@ use math::{get_power_series, get_power_series_with_offset, FieldElement};
...
+            boundary: b_coefficients,
+        })
+    }
+
+    /// Generates the random values used in the construction of the constraint composition
+    /// polynomial when algebraic batching is used.
+    pub fn draw_algebraic(
+        public_coin: &mut impl RandomCoin<BaseField = E::BaseField>,
+        num_of_transition_constraints: usize,
+        num_of_boundary_constraints: usize,
```
irakliyk, line 74: Same nit as above.

air/src/options.rs
```diff
@@ -122,13 +127,15 @@ impl ProofOptions {
...
     pub const fn new(
         num_queries: usize,
         blowup_factor: usize,
         grinding_factor: u32,
         field_extension: FieldExtension,
         fri_folding_factor: usize,
         fri_remainder_max_degree: usize,
+        batching_constraints: BatchingMethod,
         batching_deep: BatchingMethod,
     ) -> ProofOptions {
```
irakliyk, line 140: Not for this PR, but we should refactor how we construct `ProofOptions` - maybe it should be more like a builder pattern.

air/src/proof/context.rs
```diff
@@ -18,18 +18,24 @@ pub struct Context {
...
-    /// blowup factor is greater then [u32::MAX].
-    pub fn new<B: StarkField>(trace_info: TraceInfo, options: ProofOptions) -> Self {
+    /// - If either trace length or the LDE domain size implied by the trace length and the
+    ///   blowup factor is greater then [u32::MAX].
+    /// - If the number of constraints is greater than [u32::MAX].
+    pub fn new<B: StarkField>(
+        trace_info: TraceInfo,
+        options: ProofOptions,
+        total_num_of_constraints: usize,
+    ) -> Self {
```
irakliyk, line 38: nit: I would rename into just `num_constraints` here and in other places.

Also, we should probably make sure that the number of constraints is greater than 0.

air/src/proof/context.rs
```diff
@@ -148,7 +165,15 @@ impl Deserializable for Context {
         // read options
         let options = ProofOptions::read_from(source)?;
 
-        Ok(Context { trace_info, field_modulus_bytes, options })
+        // read total number of constraints
+        let total_num_of_constraints = source.read_usize()?;
```
irakliyk, line 169: Would this not make deserialization tricky in the context of a recursive verifier? Or we don't need to deserialize the context there?

Al-Kindi-0: Why would that be the case?

air/src/options.rs
Al-Kindi-0: Created an issue for it

air/src/air/coefficients.rs
Al-Kindi-0: changed

air/src/proof/context.rs
Al-Kindi-0: I see what you mean, I think. 
Currently `Context::to_elements` discards the number of constraints and I think this is fine. We could add also add it at the cost one extra field element that has to be absorbed when initializing the Fiat-Shamir transcript.

irakliyk: Adding 1 more element would make `Context::to_elements()` return 9 elements, right? Maybe we can change how `ProofOptions::to_elements()` works so that it takes 1 or 2 elements less, and then adding 1 more element here should be fine.

For `ProofOptions`, we could probably put blowup factor, grinding factor, and the number of queries into a single element.

Al-Kindi-0: Indeed, it will increase it to 9.
I went ahead with the solution of adding the blowup factor to the field element already encoding field extension, fri folding factor and remainder poly max degree.
The reason for this choice is that it would be nice to have number of queries and grinding factor as field elements so that it is easy to change them, for example, in testing.

Al-Kindi-0:
Added a round trip test for the serialization of ProofOptions

air/src/proof/context.rs
Al-Kindi-0: updated

irakliyk (approved):
All looks good! Thank you!

PR #365, opened 2025-02-07
Update options.rs

HoverCatz:
Fixed a bug in the serializer. It was writing the batching_deep variable in the wrong place, compared to the deserializer. Tested and seems to work.


irakliyk (approved):
Looks good! Thank you for catching this!

PR #366, opened 2025-02-10
fix: Proof de/serialization

SantiagoPittella:
Serialization of the `Proof` struct was failing due to a missing field in the `Context` serialization and because it was not writing the number of trace queries.

### DIFF
--- a/air/src/proof/context.rs
+++ b/air/src/proof/context.rs
@@ -147,6 +147,7 @@ impl Serializable for Context {
         target.write_u8(self.field_modulus_bytes.len() as u8);
         target.write_bytes(&self.field_modulus_bytes);
         self.options.write_into(target);
+        target.write_usize(self.num_constraints);
     }
 }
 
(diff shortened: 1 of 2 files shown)


irakliyk:
Thank you for finding this issue. There are actually two separate problems here which need a different fix.

- First, the dummy proof wasn't being constructed correctly (i.e., it was internally inconsistent). This caused an issue for deserializing dummy proofs specifically (when correct proofs were working).
- Second, in v0.12.0 release, we broke `Context` struct serialization. Thus, serialization of valid proofs was failing as well.

I fixed both issues in b067ed5efb963a4c709b3950b6a03491c4e94a35 and published the `winter-air` crate v0.12.1 with this fix.

Also, separately, I fixed dummy proof construction and released `winter-air` v0.11.1 crate.

SantiagoPittella:
Thank for the releases! I just checked it and solved my issue.

I'm closing this PR since it is completed. 🚀

PR #368, opened 2025-02-15
Fix FRI folded layers domain

plafer:
Fixes the (implied) FRI layer domains.

For example, let 
- `domain_size = 16`,
- `folding_factor = 2`,
- $\omega$ be the generator of group size 16,
- $h$ be the domain offset,

then the FRI domains are

- $[h, h\omega, h \omega^2, h\omega^3, \dots, h\omega^{15}]$,
- $[h^2, h^2\omega^2, h^2 \omega^4, h^2\omega^6, h^2\omega^8, h^2\omega^{10}, h^2\omega^{12}, h^2\omega^{14}]$,
- $[h^4, h^4\omega^4, h^4 \omega^8, h^4\omega^{12}]$,
- $[h^8, h^8\omega^8]$.

Noticeably, the domain offset is also squared - which the previous implementation did not do. The Miden VM's recursive verifier will probably need to be updated accordingly as well.

cc @Al-Kindi-0

### DIFF
--- a/fri/src/prover/mod.rs
+++ b/fri/src/prover/mod.rs
@@ -184,14 +184,17 @@ where
 
         // reduce the degree by folding_factor at each iteration until the remaining polynomial
         // has small enough degree
+        let mut domain_offset = self.options.domain_offset();
         for _ in 0..self.options.num_fri_layers(evaluations.len()) {
             match self.folding_factor() {
-                2 => self.build_layer::<2>(channel, &mut evaluations),
-                4 => self.build_layer::<4>(channel, &mut evaluations),
-                8 => self.build_layer::<8>(channel, &mut evaluations),
-                16 => self.build_layer::<16>(channel, &mut evaluations),
+                2 => self.build_layer::<2>(channel, &mut evaluations, domain_offset),
+                4 => self.build_layer::<4>(channel, &mut evaluations, domain_offset),
+                8 => self.build_layer::<8>(channel, &mut evaluations, domain_offset),
+                16 => self.build_layer::<16>(channel, &mut evaluations, domain_offset),
                 _ => unimplemented!("folding factor {} is not supported", self.folding_factor()),
             }
+
+            domain_offset = domain_offset.exp_vartime((self.folding_factor() as u32).into());
         }
 
         self.set_remainder(channel, &mut evaluations);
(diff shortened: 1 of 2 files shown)


Al-Kindi-0:
Unless there is a strong reason for this, I would keep it as is. I agree that it is probably more natural to define the folded domains using a "pure" square map in the case of folding factor 2, but the current implementation is equally valid and can be expressed using the same framework.

plafer:
> but the current implementation is equally valid and can be expressed using the same framework.

I can't find the right way to explain what's going on though. For example, for a folding factor of 2, we have

$$
f^{(d-1)}(x) = \frac{1}{2} (f^{(d)}(x^{1/2}) + f^{(d)}(-x^{1/2})) + \frac{\alpha}{2 x^{1/2}} (f^{(d)}(x^{1/2}) - f^{(d)}(-x^{1/2}))
$$

If the domain of $f^{(d-1)}$ is $[h, h\omega^{2}, h\omega^{4}, \dots, h\omega^{14}]$ (as currently implemented), then *e.g.* $f^{(d)}(x^{1/2})$ over this domain is $[f^{(d)}(h^{1/2}), f^{(d)}(h^{1/2} \omega), f^{(d)}(h^{1/2} \omega^2), \dots]$, which are not evaluations committed to in the first layer.

What am I missing?

Al-Kindi-0:
I think you can just define the projection map to be, instead of a pure square, $\pi_d\left(\\{h\cdot \omega^i, h\cdot\omega^{i + \frac{N}{2}}\\} \right)  = h \cdot \omega^{2i}$.
We can now define the evaluations of the folded polynomial at the points of the next domain to be
$$P^{d + 1}\left( h \cdot \omega^{2i} \right) := \frac{1}{2} \cdot \left(P^d\left(h\cdot\omega^i\right) + P^d\left(-h\cdot\omega^i\right)\right) + \alpha \cdot \frac{P^d\left(h\cdot\omega^i\right) - P^d\left(-h\cdot\omega^i\right)}{h\cdot\omega^i - (-h\cdot\omega^i)}$$
where we used $\omega^{i + \frac{N}{2}} = -\omega^i$.

plafer:
Awesome, thank you for the detailed explanation, that clears things up!

PR #370, opened 2025-02-23
fix: Rename FRI_MAX_REMAINDER_DEGREE constant to FRI_MAX_REMAINDER_MAX_DEGREE

crStiv:
Renamed constant to match its usage in ProofOptions struct and make the naming more consistent with its purpose as a maximum value for the remainder max degree parameter.


irakliyk:
Closing as this is not a desired change.

PR #373, opened 2025-03-18
Reverse remainder commitment

Al-Kindi-0:
This PR changes the way we commit to the FRI remainder polynomial i.e., we commit to its coefficients in reverse order.

### DIFF
--- a/fri/src/prover/mod.rs
+++ b/fri/src/prover/mod.rs
@@ -223,6 +223,10 @@ where
 
     /// Creates remainder polynomial in coefficient form from a vector of `evaluations` over a
     /// domain.
+    ///
+    /// We commit to the coefficients of the remainder polynomial in reverse order so that
+    /// evaluating the remainder polynomial on the verifier end, using Horner's evaluation
+    /// method, becomes easier.
     fn set_remainder(&mut self, channel: &mut C, evaluations: &mut [E]) {
         let inv_twiddles = fft::get_inv_twiddles(evaluations.len());
         fft::interpolate_poly_with_offset(evaluations, &inv_twiddles, self.options.domain_offset());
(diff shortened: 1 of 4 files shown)


fri/src/prover/mod.rs
```diff
@@ -227,7 +227,8 @@ where
         let inv_twiddles = fft::get_inv_twiddles(evaluations.len());
         fft::interpolate_poly_with_offset(evaluations, &inv_twiddles, self.options.domain_offset());
         let remainder_poly_size = evaluations.len() / self.options.blowup_factor();
-        let remainder_poly = evaluations[..remainder_poly_size].to_vec();
+        let remainder_poly: Vec<_> =
+            evaluations[..remainder_poly_size].iter().copied().rev().collect();
```
irakliyk, line 231: Let's document this - probably in the doc comments for this function, but also we should describe this in publicly accessible comments as well.

fri/src/verifier/mod.rs
```diff
@@ -313,7 +313,7 @@ where
             return Err(VerifierError::RemainderDegreeMismatch(max_degree_plus_1 - 1));
         }
         let offset: E::BaseField = self.options().domain_offset();
-
+        let remainder_poly: Vec<_> = remainder_poly.iter().copied().rev().collect();
```
irakliyk, line 316: Let's add a comment here about why reversal is needed.

irakliyk (approved):
Looks good! Thank you! I left a couple of doc-related comments. Also, let's add an entry to the changelog for this.

fri/src/verifier/mod.rs
Al-Kindi-0: I think the best approach would be to have say a `eval_horner_rev` function which would be just like `eval_horner` but without the `rev()`. This way we can remove the above line and it would be consistent with the goal of this PR, namely simplify/reduce the work on the verifier side.
What do you think?

irakliyk: Sounds good!

irakliyk (approved):
All looks good! Thank you!

PR #382, opened 2025-06-02
Absorb OOD evaluations only once into the transcript

Al-Kindi-0:
Merges the OOD evaluations of the trace and constraints composition polynomials into one vector which is absorbed into the transcript in one-go.


adr1anh (commented):
Looks good!

PR #383, opened 2025-06-06
fix(utils): incorrect buffer length check in slice reader impl

bitwalker:
When looking into a deserialization issue in another project, I discovered that we were incorrectly checking the length of the unread buffer when attempting to read `N` bytes as a slice. As a result, deserialization would fail any time `read_slice` was called when the current total buffer length as `>= N`, but the unread portion was `< N`.

Additionally, when running some of the various CI checks locally, I ran into some new nightly warnings with clippy that I've fixed in a separate commit, anticipating that CI would fail unless I addressed those as well.


irakliyk (approved):
Looks good! Thank you! Should we also add a test to prevent potential future regressions?

> @irakliyk I think it might be problematic to run `clippy`'s nightly version when the MSRV of these crates is much lower, at least as a _required_ check. At least in this case, it causes CI to fail even though I haven't made any changes to the affected code.

No worries - I can merge as is and then fix the lints in a follow-up commit.

PR #394, opened 2025-07-19
fix: prevent panic in rand_vector when `n == 0`

huitseeker:
When called with `n == 0` (aka asked to produce a vector of random numbers of size 0)
`rand_vector` falls through to its final panic.

This special-cases n == 0 to return an empy vector instead.

This fixes https://github.com/0xMiden/miden-vm/issues/1314


irakliyk (approved):
Looks good! Thank you for fixing it!

PR #395, opened 2025-07-25
feat: Implement Serializable/Deserializable for hashbrown structures

c410-f3r:
`hashbrown` allows the usage of hash-based structures in `no_std` scenarios.

It is not known why `winter-utils` doesn't implement Serializable/Deserializable for the hash-based structures provided by the standard library (out-of-order insertion?). `serde` does and it is valid use-case.

### DIFF
--- a/utils/core/Cargo.toml
+++ b/utils/core/Cargo.toml
@@ -21,6 +21,7 @@ default = ["std"]
 std = []
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


c410-f3r:
@irakliyk Let me know if I am missing anything

c410-f3r:
Ping @irakliyk 

Sorry bothering you :)

PR #408, opened 2026-03-07
fix(air): fix inconsistent assertion

YichiZhang0613:
In winterfell/air/src/proof/table.rs, the comment and assertion's message requires num_rows and num_cols <= 255(MAX_ROWS and MAX_COLS = 255), while the code check is num_rows < MAX_ROWS and  num_cols < MAX_ROWS, which should be fixed into num_rows <= MAX_ROWS and num_rows <= MAX_COLS.
```rust
const MAX_ROWS: usize = 255;
const MAX_COLS: usize = 255;
...
    /// Panics if:
    /// * Specified number of rows is 0 or greater than 255.
    /// * Specified number of columns is 0 or greater than 255.
    /// * Provided bytes do not encode valid field elements required to fill the table.
    pub fn from_bytes(
        bytes: &[u8],
        num_rows: usize,
        num_cols: usize,
    ) -> Result<Self, DeserializationError> {
        assert!(num_rows > 0, "number of rows must be greater than 0");
        assert!(
            num_rows < MAX_ROWS,
            "number of rows cannot exceed {MAX_ROWS}, but was {num_rows}"
        );
        assert!(num_cols > 0, "number of columns must be greater than 0");
        assert!(
            num_cols < MAX_ROWS,
            "number of columns cannot exceed {MAX_COLS}, but was {num_cols}"
        );

        let mut reader = SliceReader::new(bytes);
        let num_elements = num_rows * num_cols;
        Ok(Self {
            data: reader.read_many(num_elements)?,
            row_width: num_cols,
        })
    }
```


YichiZhang0613:
@irakliyk Hi, sorry for the interruption. I see you've been active reviewer recently. Could you please review this PR when you have time? It's been open for 2 months and I'd appreciate any feedback. Thank you very much.

